PR #36, opened 2024-05-20
TableMetadataBuilder

rwwwx:
Draft


crates/iceberg-ext/src/spec/spec_binder.rs
```diff
@@ -0,0 +1,476 @@
...
+    }
+
+    pub fn bind(mut self, spec: &UnboundPartitionSpec) -> Result<PartitionSpec> {
+        Ok(PartitionSpec {
+            spec_id: self.spec_id,
+            fields: spec
+                .fields
+                .iter()
+                .map(|unbounded_field| self.bind_field(unbounded_field))
+                .collect::<Result<Vec<PartitionField>>>()?,
```
c-thiel, line 39: There is a check missing here I believe: We need to make sure, that each PartitionField only occurs once in a Partition Spec. Java has the `CheckForRedundantPartitions` function (https://github.com/apache/iceberg/blob/2e291c2b67a643ffe93e139483df55f3639cc39d/api/src/main/java/org/apache/iceberg/PartitionSpec.java#L393) which is then called by each of the transforms:
https://github.com/apache/iceberg/blob/2e291c2b67a643ffe93e139483df55f3639cc39d/api/src/main/java/org/apache/iceberg/PartitionSpec.java#L423

We should also add a test for this then, to make sure that we don't have the same column twice in a single partition spec.

crates/iceberg-ext/src/spec/spec_binder.rs
```diff
@@ -0,0 +1,476 @@
...
+        let schema_field = self.get_schema_field_by_name(&spec_field.name)?;
+
+        self.check_transform_compatibility(&spec_field.transform, &schema_field.field_type)?;
+        Self::check_schema_field_eq_source_id(schema_field, spec_field.source_id)?;
+
+        self.update_names(&spec_field.name);
+
+        let partition_id = spec_field
+            .partition_id
+            .unwrap_or_else(|| self.increment_and_get_next_partition_id());
```
c-thiel, line 55: I was really confused by the naming of the `partition_id` attribute - which is a really bad name in the core rust crate.
I would prefer to rename `increment_and_get_next_partition_id` to `increment_and_get_next_field_id` (since its modifying `self.last_assigned_*field*_id`) and also name the assigned variable let field_id instead of partition_id.
Otherwise line 64 is also quite confusing where we take the max between a field and a partition id.
We could then add a small comment in line 52 stating that the partition_id attribute is not actually the id of this partition, but instead the identifier of the field in the partition. What do you think?

c-thiel (changes requested):
Nicely written PartitionBinder - thanks @rwwwx!
Just a few minor comments from my side and I believe one check is missing.

crates/iceberg-ext/src/spec/partition_binder.rs
```diff
@@ -0,0 +1,510 @@
...
+    pub(crate) fn bind(mut self, spec: UnboundPartitionSpec) -> Result<PartitionSpec> {
+        Ok(PartitionSpec {
+            spec_id: self.spec_id,
+            fields: spec
+                .fields
+                .into_iter()
+                .map(|unbound_field| self.bind_field(unbound_field))
+                .collect::<Result<Vec<PartitionField>>>()?, // .into_iter()
+                                                            // .map(|bound_field| self.check_for_redundant_partitions(bound_field))
+                                                            // .collect::<Result<Vec<PartitionField>>>()?,
```
c-thiel, line 44: The implementation in bind_field is perfect, no need to add it here as well

crates/iceberg-ext/src/spec/partition_binder.rs
```diff
@@ -0,0 +1,510 @@
...
+    fn check_schema_field_eq_source_id(
+        schema_field: &NestedFieldRef,
+        source_id: i32,
+    ) -> Result<()> {
+        if schema_field.id.eq(&source_id) {
+            return Ok(());
+        }
+
+        Err(Self::err(
+            "Cannot create identity partition sourced from different field in schema.",
```
c-thiel, line 121: Could we add a bit more information in the Error about the field that failed?

crates/iceberg-ext/src/spec/partition_binder.rs
```diff
@@ -0,0 +1,510 @@
...
+        transform: Transform,
+    ) -> Result<()> {
+        if !self
+            .dedup_fields
+            .contains(&(source_id, transform.dedup_name()))
+        {
+            return Ok(());
+        }
+
+        Err(Self::err("Cannot add redundant partition!"))
```
c-thiel, line 159: Could we add a bit more information in the Error about the field that failed?

crates/iceberg-ext/src/spec/table_metadata.rs
```diff
@@ -29,36 +48,45 @@ pub struct TableMetadataBuilder {
...
-            current_snapshot_id: None,
-            snapshots: HashMap::new(),
-            snapshot_log: vec![],
-            sort_orders: HashMap::new(),
-            metadata_log: vec![],
-            default_sort_order_id: -1,
-            refs: HashMap::default(),
-        };
-
         Self {
```
c-thiel, line 61: Are there any changes to log on new()? Maybe set table uuid?
Lets double check the java implementation.

If we don't add the change here, a reproduced table from a series of changes will end up with a different ID.

crates/iceberg-ext/src/spec/table_metadata.rs
```diff
@@ -186,233 +303,292 @@ impl TableMetadataBuilder {
     /// # Errors
     /// - Current schema already set.
     pub fn set_current_schema(&mut self, schema_id: i32) -> Result<&mut Self> {
-        if self.current_schema_id.is_some() {
+        if schema_id == -1 {
```
c-thiel, line 306: Lets use the const

c-thiel (commented):
I only got half-through today. I'll check the rest tomorrow.
It would be great if you could check the pipeline errors.
Once they are solved, I'll add some integration tests for spark & pyiceberg to actually use all of these fancy new features :)

crates/iceberg-ext/src/spec/table_metadata.rs
rwwwx: According to Java implementation if we create `Builder` from scratch the `changes` list will be empty. However, if we create `Builder` from a `TableMetadata` the  `changes` will be copied from this metadata. But the Rust implementation doesn't have this `changes` field inside of the `TableMetadata` struct so we cannot copy it. 😖

c-thiel:
@rwwwx  a few things to do:

- [ ] For current-schema-id, default-spec-id & last-partition-id updates the datatypes also aren't right (search for i64::from). Lets PR fixes over to iceberg-rust as well (i.e.  https://github.com/rwwwx/iceberg-rest-server/blob/f869f34ce4730bcad5eef91eeb63b2ac97aae238/crates/iceberg-ext/src/catalog/rest/table.rs#L192)
- [ ] I fixed some problems with reuse_or_create_...id functions, but I really don't like the way how I implemented partial equality (without id). https://github.com/rwwwx/iceberg-rest-server/blob/f869f34ce4730bcad5eef91eeb63b2ac97aae238/crates/iceberg-ext/src/spec/table_metadata.rs#L249 . Maybe we can add a .eq_without_id method to the structs in the main crate?
- [ ] We desperately need unit tests
- [ ] I changed the add_schema / sort_order / spec_id functions together with their reuse or create id functions. Please review the changes and try to improve them. A few bugs in the previous implementation: If an explicit id was specified (lets say 2), it could still happen that the old reuse function assigned a new one. It should only assign a new one if -1 was specified. Also, comparisons where made which included the .id fields, so that the .eq comparison would never hit. We need tests please.
- [ ] Calculated field partitioning is not working: https://github.com/rwwwx/iceberg-rest-server/blob/f869f34ce4730bcad5eef91eeb63b2ac97aae238/tests/python/tests/test_spark.py#L175C1-L184C27

PR #50, opened 2024-06-03
added `warehouse_status` column

rwwwx:


c-thiel (commented):
This goes in a good direction. Let's implement the remaining method.

We also don't want  people to use inactive warehouses, so we should add filter to queries in https://github.com/hansetag/iceberg-rest-server/blob/main/crates/iceberg-rest-server/src/implementations/postgres/table.rs and https://github.com/hansetag/iceberg-rest-server/blob/main/crates/iceberg-rest-server/src/implementations/postgres/namespace.rs
(maybe other files as well?) to make sure all queries only work for active Warehouses.

crates/iceberg-rest-server/src/api/v1/warehouse.rs
```diff
@@ -75,6 +105,113 @@ pub trait WarehouseService<C: Catalog, A: AuthZHandler, S: SecretStore> {
...
         })
     }
+
+    async fn update_warehouse(
+        request: UpdateWarehouseRequest,
+        context: ApiContext<State<A, C, S>>,
+        _headers: HeaderMap,
+    ) -> Result<UpdateWarehouseResponse> {
+        let warehouse_id = &WarehouseIdent::from(request.warehouse_id);
+        let mut transaction = C::Transaction::begin_write(context.v1_state.catalog.clone()).await?;
```
c-thiel, line 115: Its more efficient to start the transaction after we have retrieved the old storage profile and secret.

crates/iceberg-rest-server/src/api/v1/warehouse.rs
```diff
@@ -75,6 +105,113 @@ pub trait WarehouseService<C: Catalog, A: AuthZHandler, S: SecretStore> {
...
+        if let Some(new_name) = request.new_name {
+            if new_name != old_name {
+                C::update_warehouse_name(warehouse_id, &new_name, transaction.transaction())
+                    .await?;
+            }
+        }
+
+        match (request.new_storage_credential, request.new_storage_profile) {
+            (None, Some(_)) => {
+                let err: IcebergErrorResponse = ErrorModel::builder()
```
c-thiel, line 135: There are some Storage Profiles that don't need credentials. Sorry, I forgot that case earlier.
In this case of the match statement we should validate the Storage Profile just like you did above, if it works, delete the old secret and store the new profile with storage_credential_id = None

crates/iceberg-rest-server/src/api/v1/warehouse.rs
```diff
@@ -75,6 +105,113 @@ pub trait WarehouseService<C: Catalog, A: AuthZHandler, S: SecretStore> {
...
+                    C::update_warehouse_storage_profile(
+                        warehouse_id,
+                        new_storage_profile,
+                        transaction.transaction(),
+                    )
+                    .await?;
+                }
+
+                if let Some(old_cred) = old_storage_credential_id {
+                    S::delete_secret(&old_cred, context.v1_state.secrets.clone()).await?;
```
c-thiel, line 167: This is an atomic transaction. We should delete the secret at the very end of the function - even after the Postgres Transaction has been committed. Otherwise we might end up with a corrupt state (i.e. old secret already deleted, warehouse commit failed, so that old warehouse commits now points to nonexistent secret).

Also use .ok() here, as the transaction can still succeed if deleting the secret fails.

crates/iceberg-rest-server/src/api/v1/warehouse.rs
```diff
@@ -75,6 +105,113 @@ pub trait WarehouseService<C: Catalog, A: AuthZHandler, S: SecretStore> {
...
+                }
+
+                let new_secret_id =
+                    Some(S::create_secret(new_storage_credential, context.v1_state.secrets).await?);
```
c-thiel, line 171: In contrast, this is exactly at the right place and can stay here.

crates/iceberg-rest-server/src/api/v1/warehouse.rs
```diff
@@ -75,6 +105,113 @@ pub trait WarehouseService<C: Catalog, A: AuthZHandler, S: SecretStore> {
...
+                Ok(C::update_warehouse_storage_secret_id(
+                    warehouse_id,
+                    new_secret_id,
+                    transaction.transaction(),
+                )
+                .await?)
+            }
+            (Some(new_storage_credential), None) => {
+                if let Some(old_cred) = old_storage_credential_id {
+                    S::delete_secret(&old_cred, context.v1_state.secrets.clone()).await?;
```
c-thiel, line 182: This also needs to be at the very end after the transaction succeeded.
Also use .ok() here and just log an error. If this fails, the update still works. We need an external reconciliation process later.

crates/iceberg-rest-server/src/api/v1/warehouse.rs
```diff
@@ -75,6 +105,113 @@ pub trait WarehouseService<C: Catalog, A: AuthZHandler, S: SecretStore> {
...
+            }
+        }?;
+
+        if let Some(new_status) = request.new_warehouse_status {
+            match new_status {
+                WarehouseStatus::Active => {
+                    C::activate_warehouse(warehouse_id, context.v1_state.catalog)
+                }
+                WarehouseStatus::Inactive => {
+                    C::deactivate_warehouse(warehouse_id, context.v1_state.catalog)
```
c-thiel, line 203: I think for C:: we should just have one function that takes the Enum state as input, otherwise we will end up with a lot of functions in the future if we add more states.

c-thiel: Lets also compare new to old state and only take action if they differ.

crates/iceberg-rest-server/src/api/v1/warehouse.rs
```diff
@@ -75,6 +105,113 @@ pub trait WarehouseService<C: Catalog, A: AuthZHandler, S: SecretStore> {
...
+                }
+            }
+            .await?
+        }
+
+        transaction.commit().await?;
+
+        Ok(UpdateWarehouseResponse {
```
c-thiel, line 211: We should return unchanged (304) status code if there was no need to change anything.
If credentials where specified, we always need to return 200 even if the old creds are identical to the new creds as we would otherwise leak information about the credentials.

crates/iceberg-rest-server/src/implementations/postgres/warehouse.rs
```diff
@@ -140,6 +154,237 @@ fn validate_warehouse_name(warehouse_name: &str) -> Result<()> {
...
+
+fn warehouse_not_found(warehouse_id: &WarehouseIdent) -> ErrorModel {
+    ErrorModel::builder()
+        .code(StatusCode::NOT_FOUND.into())
+        .message(format!("Warehouse not found: '{warehouse_id:?}'"))
+        .r#type("WarehouseNotFound")
+        .build()
+}
+
+fn serialize_storage_profile(storage_profile: StorageProfile) -> Result<Value, ErrorModel> {
```
c-thiel, line 366: How do we do it in the "create_warehouse" function? Could we also use this function? Or make it part of the implementation block of StorageProfile?

crates/iceberg-rest-server/src/api/v1/warehouse.rs
```diff
@@ -75,6 +105,113 @@ pub trait WarehouseService<C: Catalog, A: AuthZHandler, S: SecretStore> {
             warehouse_id: warehouse_id.into_uuid(),
         })
     }
+
+    async fn update_warehouse(
+        request: UpdateWarehouseRequest,
+        context: ApiContext<State<A, C, S>>,
+        _headers: HeaderMap,
+    ) -> Result<UpdateWarehouseResponse> {
```
c-thiel, line 113: Here, as well as in the create_warehouse request there is no check for AuthZ yet (using the A generic).
Could you add those as well?

c-thiel:
Consolidated in https://github.com/hansetag/iceberg-catalog/pull/99

Issue #20, opened 2024-05-14
Emit Change Events


c-thiel:
We want to emit events on Metadata changes.
For this, similar to current `SecretStore` or `Catalog` implementations, the goal is to support different event queues via different implementations.

Some guidelines:
- Use [CloudEvents](https://cloudevents.io), otherwise stick as closely as possible to Iceberg Spec Messages (such as `TableUpdate`)
- First implementation should be a Nirvana Handler -> Do not emit events and secondly support for https://nats.io
- Besides just publishing events, we need request/response mechanism for "commit_table", "delete_table" and "commit_transaction" endpoints, which can be used by external systems to allow or disallow changes. This can be used. to prohibit changes that violate Data Contracts in the future.

PR #55, opened 2024-06-04
add Publisher trait + dummy impls

twuebi:
#20

### DIFF
--- a/crates/iceberg-rest-server/src/catalog/tables.rs
+++ b/crates/iceberg-rest-server/src/catalog/tables.rs
@@ -17,6 +17,7 @@ use super::{
     namespace::{uppercase_first_letter, validate_namespace_ident},
     require_warehouse_id, CatalogServer,
 };
+use crate::service::event_publisher::EventPublisher;
 use crate::service::storage::StorageCredential;
 use crate::service::{
     auth::AuthZHandler, secrets::SecretStore, Catalog, CreateTableResult,
@@ -25,14 +26,14 @@ use crate::service::{
 use crate::service::{GetStorageConfigResult, TableIdentUuid};
 
 #[async_trait::async_trait]
-impl<C: Catalog, A: AuthZHandler, S: SecretStore>
-    iceberg_rest_service::v1::tables::Service<State<A, C, S>> for CatalogServer<C, A, S>
+impl<C: Catalog, A: AuthZHandler, S: SecretStore, P: EventPublisher>
+    iceberg_rest_service::v1::tables::Service<State<A, C, S, P>> for CatalogServer<C, A, S, P>
 {
     /// List all table identifiers underneath a given namespace
     async fn list_tables(
         parameters: NamespaceParameters,
         _query: PaginationQuery,
-        state: ApiContext<State<A, C, S>>,
+        state: ApiContext<State<A, C, S, P>>,
         headers: HeaderMap,
     ) -> Result<ListTablesResponse> {
         // ------------------- VALIDATIONS -------------------
@@ -65,7 +66,7 @@ impl<C: Catalog, A: AuthZHandler, S: SecretStore>
         // mut because we need to change location
         mut request: CreateTableRequest,
         data_access: DataAccess,
-        state: ApiContext<State<A, C, S>>,
+        state: ApiContext<State<A, C, S, P>>,
         headers: HeaderMap,
     ) -> Result<LoadTableResult> {
         // ------------------- VALIDATIONS -------------------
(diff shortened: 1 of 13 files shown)


crates/iceberg-rest-server/src/catalog/tables.rs
```diff
@@ -779,3 +827,33 @@ fn validate_table_ident(table: &TableIdent) -> Result<()> {
...
     Ok(())
 }
+
+// This function does not return a result but serde_json::Value::Null if serialization of tab.table
+// fails. This follows the rationale that we'll likely end up ignoring the error in the API handler
+// anyway since we already effected the change and only the event emission about the change failed.
+// Given that we are serializing stuff that we've received as path parameters and successfully
+// processed, it's unlikely to cause issues.
+fn maybe_table_parameters_to_json(tab: &TableParameters) -> serde_json::Value {
+    let table_json = if let Ok(table) = serde_json::to_value(&tab.table) {
```
c-thiel, line 837: Two notes here:
1. I think we can make it not fail like this:
```rust
let table_json = serde_json::json!({"namespace": tab.table.namespace.clone().inner(),
                                        "name": tab.table.name.as_str().to_string()});
```

2., we will at least need to include the namespace (which is part of the table object) as well.
Currently Namespaces are uniquely identified by their path segments, i.e. the Vec<Str> that is the inner of `NamespaceIdent`. Tables are however not best identified by their names, hence `TableIdentifier` is a bad name that we import from iceberg-rust. Instead, we should identify tables via their uuids which are must more persistent. Using the uuids everywhere internally, and also for the table path, makes it much easier to rename a table - it's just a metadata operation without any file movement. For Tables we went the extra mile to use IDs everywhere - for Namespaces we did not. In the Namespace ident (Vec<String>) there is very important information about the hierarchy that we would otherwise have to fetch each time when interacting with a namespace. However, eventually also switching to IDs for Namespaces would be nice.

I think the events would be most usable if we had:
- table-id
- warehouse-id (instead of prefix, we extract warehouse-id from the prefix anyway in each function)
- namespace: Vec<String>
- Maybe to prepare for the future also namespace-id, but for now I am happy with the namespace itself.

crates/iceberg-rest-server/src/catalog/tables.rs
```diff
@@ -725,11 +752,32 @@ impl<C: Catalog, A: AuthZHandler, S: SecretStore>
...
+    parameters: serde_json::Value,
+    body: serde_json::Value,
+    operation_id: &str,
+    publisher: impl EventPublisher,
+) {
+    publisher
+        .publish(
+            Uuid::now_v7(),
+            operation_id,
+            &serde_json::json!({"body": body, "path_parameters": parameters}),
```
c-thiel, line 776: I think we should separate the "body" (as data) from the "path_parameters" (as context attributes) in Cloudevents.
The reason is that context of Cloudevents are great for routing events or building listeners on certain events (i.e. for a specific Warehouse) without the need to parse the full body.
You can see this in this [Example](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md#example) of a json representation of a Cloudevent: Extensions are on the outer level, while data is hidden away in its own field.

Quoting the docs:
"These attributes, while descriptive of the event, are designed such that they can be serialized independent of the event data. This allows for them to be inspected at the destination without having to deserialize the event data."

crates/iceberg-rest-server/src/catalog/tables.rs
```diff
@@ -779,3 +827,33 @@ fn validate_table_ident(table: &TableIdent) -> Result<()> {
...
+    serde_json::json!({"prefix": tab.prefix.as_ref().map(|pre| pre.as_str().to_string()),
+                           "table": table_json})
+}
+
+// This function does not return a result but serde_json::Value::Null if serialization of tab.table
+// fails. This follows the rationale that we'll likely end up ignoring the error in the API handler
+// anyway since we already effected the change and only the event emission about the change failed.
+// Given that we are serializing stuff we've received as a json body and also successfully
+// processed, it's unlikely to cause issues.
+fn maybe_body_to_json(request: impl Serialize) -> serde_json::Value {
```
c-thiel, line 852: This function might even become obsolete if we follow my suggestion from above to split the body from metadata attributes, as error handling if serialization fails is already part of the publisher

c-thiel (changes requested):

crates/iceberg-rest-server/src/catalog/tables.rs
twuebi: My motivation for putting the path params inside the data section was that - based on my still superficial understanding - the path parameters are necessary information to do anything useful with the request data, without these ids it's only about some tables which made me hesitate to put them in the extensions. 

> "This allows for them to be inspected at the destination without having to deserialize the event data."
Convinces me, I'll change it, it's probably very useful to have the ability to filter down events for certain tables without deserializing the data section.

crates/iceberg-rest-server/src/catalog/tables.rs
twuebi: That'd mean we need to make all requests (and their fields) clone and we will end up doing Request -> cloned Request -> Json, I agree that we get clearer encapsulation

c-thiel (approved):

PR #56, opened 2024-06-04
nats publisher

twuebi:


crates/iceberg-rest-server/src/service/event_publisher.rs
```diff
@@ -40,10 +40,45 @@ impl EventPublisher for TracingPublisher {
...
+    pub topic: String,
+}
+
+#[async_trait::async_trait]
+impl EventPublisher for NatsPublisher {
+    async fn publish(&self, id: Uuid, typ: &str, data: serde_json::Value, metadata: Vec<(String, String)>,
+    ) -> anyhow::Result<()> {
+        use cloudevents::{EventBuilder, EventBuilderV10};
+
+        let event_builder = EventBuilderV10::new()
```
c-thiel, line 63: I think we should move this logic to the EventPublisher or introduce a separate function or layer.
We want the exact same cloud event to be sent to Kafka and REST APIs later.

PR #65, opened 2024-06-07
nats publisher

twuebi:
the last mr was accidentally merged into a feature branch instead of main


twuebi:
tests are passing locally, maybe flaky? @c-thiel

c-thiel (approved):
Maybe we could add also event publishing to create_table in a separate issue

PR #77, opened 2024-06-11
Fix flaky tests

twuebi:
service_started is not enough for nats, this adds a health check, seems to fix flaky CI


c-thiel:
Fixes https://github.com/hansetag/iceberg-catalog/issues/72

c-thiel (approved):

Issue #58, opened 2024-06-06
External schema change approval


c-thiel:
Implement a Trait with separate State that allows to approve schema changes of existing tables.

PR #80, opened 2024-06-11
Table change check

twuebi:
#58


crates/iceberg-catalog/src/catalog/tables.rs
```diff
@@ -819,6 +842,19 @@ impl<C: Catalog, A: AuthZHandler, S: SecretStore>
...
+        for ((update, (_, table_uuid)), response) in updates
+            .into_iter()
+            .zip(&event_table_ids)
+            .zip(&commit_response)
+        {
+            state
+                .v1_state
+                .contract_verifiers
+                .check(&update, *table_uuid, &response.previous_table_metadata)
+                .await?
```
c-thiel, line 854: There might be quite a few updates here - I would prefer concurrent execution and try_join

crates/iceberg-catalog/src/catalog/tables.rs
```diff
@@ -819,6 +842,19 @@ impl<C: Catalog, A: AuthZHandler, S: SecretStore>
             ));
         }
 
+        for ((update, (_, table_uuid)), response) in updates
```
c-thiel, line 845: I think we should push the whole block up to right after C::commit_table_transaction.
There is no need to load storage secret if the contract check fails.

crates/iceberg-catalog/src/catalog/tables.rs
c-thiel: It doesn't quite fit here, but I am also a bit hesitant about awaiting the emit_change_event block.
We currently implement a "send and forget" approach. But right now, if the event queue is offline, we are blocking the client response until the event queue gets a timeout.
What's your feeling about emitting events non-blocking?

crates/iceberg-catalog/src/catalog/tables.rs
```diff
@@ -929,3 +965,65 @@ fn maybe_body_to_json(request: impl Serialize) -> serde_json::Value {
         serde_json::Value::Null
     }
 }
+
+// TableUpdate is not clone but all containing fields are so we use this function to clone..
+fn copy_updates(updates: &[TableUpdate]) -> Vec<TableUpdate> {
```
c-thiel, line 970: Should we maybe rename to clone_updates?

https://github.com/apache/iceberg-rust/pull/402

crates/iceberg-catalog/src/service/contract_verification.rs
```diff
@@ -0,0 +1,265 @@
...
+#[async_trait]
+impl ContractVerification for ContractVerifiers {
+    fn name(&self) -> &'static str {
+        "ContractVerifiers"
+    }
+
+    async fn check(
+        &self,
+        table_updates: &[TableUpdate],
+        table_ident_uuid: TableIdentUuid,
```
c-thiel, line 190: Now that we send the ` current_metadata`, we might want to skip the table_ident_uuid.
The table-uuid is already part of `TableMetadata` and might change during updates.
If we  specify it here additionally, we should rename it to current_table_ident_uuid to make sure it is the UUID before updates.
I think removing it is fine though.

crates/iceberg-catalog/src/service/contract_verification.rs
```diff
@@ -0,0 +1,265 @@
...
+    ) -> Result<ContractVerificationOutcome, ErrorModel> {
+        for checker in &self.checkers {
+            match checker
+                .check(table_updates, table_ident_uuid, current_metadata)
+                .await
+            {
+                Ok(ContractVerificationOutcome::Clear {}) => {}
+                Ok(block_result @ ContractVerificationOutcome::Violation { error_model: _ }) => {
+                    tracing::info!(
+                        "Checker {} blocked change on table '{}'",
```
c-thiel, line 201: Lets reflect the trait in the log - maybe ContractValidator {} blocked change on table ..
Also for other logs here.

crates/iceberg-catalog/src/catalog/tables.rs
twuebi: Do we expect this endpoint to be called a lot? How many changes do we expect per call? 

We could separate it out via a channel and have some background task shoveling it into nats.

crates/iceberg-catalog/src/service/contract_verification.rs
twuebi: done

crates/iceberg-catalog/src/catalog/tables.rs
twuebi: done

crates/iceberg-catalog/src/service/contract_verification.rs
twuebi: thanks, missed it in table-metadata, removed!

crates/iceberg-catalog/src/catalog/tables.rs
twuebi: done

crates/iceberg-catalog/src/catalog/tables.rs
twuebi: > There might be quite a few updates here - I would prefer concurrent execution and try_join

made it futures + `try_join_all`

twuebi: > Do we expect this endpoint to be called a lot? How many changes do we expect per call?
> 
> We could separate it out via a channel and have some background task shoveling it into nats.

https://github.com/hansetag/iceberg-catalog/issues/86

c-thiel (approved):

PR #89, opened 2024-06-12
Secure routes via openid provider

twuebi:


crates/iceberg-catalog/src/config.rs
```diff
@@ -73,6 +73,12 @@ pub struct DynAppConfig {
     pub nats_address: Option<Url>,
     #[clap(env = "ICEBERG_REST__NATS_TOPIC")]
     pub nats_topic: Option<String>,
+
+    // ------------- AUTHORIZATION -------------
+    #[clap(env = "ICEBERG_REST__OAUTH_PROVIDER_URI")]
+    pub openid_provider_uri: Option<Url>,
```
c-thiel, line 79: Rename to oauth2?

c-thiel: +Add to docs

twuebi: added to docs, decided to stick to OPENID_PROVIDER_URI since we are actually relying on openid endpoints like `.well-known/openid-configuration`.

c-thiel (approved):

PR #99, opened 2024-06-14
Management API

c-thiel:


crates/iceberg-catalog/src/api/management/mod.rs
```diff
@@ -37,6 +35,109 @@ pub mod v1 {
...
+                    get(
+                        |AxumState(api_context): AxumState<ApiContext<State<A, C, S>>>,
+                         Extension(metadata): Extension<RequestMetadata>| {
+                            Self::list_projects(api_context, metadata)
+                        },
+                    ),
+                )
+                
+                .route(
+                    "/project/:project_id/warehouses",
```
twuebi, line 50: here we have warehouses, all other entities here are singular, e.g. `GET /project`, `GET /warehouse`, i'd use plural everywhere, we can also go for singular, but i'd try to keep it consistent

twuebi, line 50: move to top level with query param? e.g. `GET /warehouse?project_id={some_id}` that way we're consistent with the rest of the endpoints here

crates/iceberg-catalog/src/api/management/mod.rs
```diff
@@ -37,6 +35,109 @@ pub mod v1 {
...
+                        |Path(project_id): Path<uuid::Uuid>,
+                        Query(request): Query<warehouse::ListWarehouseRequest>,
+                         AxumState(api_context): AxumState<ApiContext<State<A, C, S>>>,
+                         Extension(metadata): Extension<RequestMetadata>| {
+                            Self::list_warehouses(project_id.into(), request, api_context, metadata)
+                        },
+                    )
+                )
+                // Delete warehouse - must be empty
+                .route("/warehouse/:warehouse_id",
```
twuebi, line 62: This get is more or less the same as `GET /warehouse?id={some_id}`. Although it offers some convenience, i.e. I just have to deal with a single object and not a list of objects, I don't have strong feelings regarding this, I'd leave it up to you.

crates/iceberg-catalog/src/api/management/mod.rs
```diff
@@ -37,6 +35,109 @@ pub mod v1 {
...
+                .route(
+                    "/warehouse/:warehouse_id/activate",
+                    post(
+                        |
+                        Path(warehouse_id): Path<uuid::Uuid>,
+                        AxumState(api_context): AxumState<ApiContext<State<A, C, S>>>,
+                         Extension(metadata): Extension<RequestMetadata>| {
+                            Self::activate_warehouse(warehouse_id.into(), api_context, metadata)
+                        },
+                    ),
```
twuebi, line 111: Is there something inherently different between deactivate & activate? Could make this a single `POST /warehouse/:id/status`, saves a few lines & makes for a smaller API surface

crates/iceberg-catalog/src/api/management/v1/warehouse.rs
```diff
@@ -26,29 +31,97 @@ pub struct CreateWarehouseResponse {
...
+    /// provided, we assume that this storage does not require credentials.
+    pub storage_credential: Option<StorageCredential>,
+}
+
+#[derive(Debug, Deserialize, ToSchema)]
+#[serde(rename_all = "kebab-case")]
+pub struct ListWarehouseRequest {
+    /// Optional filter to include inactive warehouses.
+    #[serde(default)]
+    pub include_inactive: Option<bool>,
```
twuebi, line 50: I this an Option to please the ToSchema gods to make it appear as not required in the openapi spec? If not then we could consider omitting the option, bool::default() should give false which I assume is what None also means.

crates/iceberg-catalog/src/api/management/mod.rs
c-thiel: I think I'll keep it for now

crates/iceberg-catalog/src/api/management/mod.rs
c-thiel: There is - deactivating is something a user might want to do. It happens when someone presses the "delete" button in the UI.
Reactivation is more for recovery. Also the processes might become quite different in the future - that's why I prefer to have two endpoints.

crates/iceberg-catalog/src/api/management/v1/warehouse.rs
```diff
@@ -26,29 +31,97 @@ pub struct CreateWarehouseResponse {
...
+    /// ID of the warehouse.
+    pub id: uuid::Uuid,
+    /// Name of the warehouse.
+    pub name: String,
+    /// Project ID in which the warehouse is created.
+    pub project_id: uuid::Uuid,
+    /// Storage profile used for the warehouse.
+    pub storage_profile: StorageProfile,
+    /// Whether the warehouse is active.
+    pub status: String,
```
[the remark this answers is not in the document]
c-thiel: I thought about this too, but then went for String to not introduce another type only for a return in the UI. I think I might switch back to an enum :)

crates/iceberg-catalog/src/api/management/v1/warehouse.rs
c-thiel: Exactly, it's meant for openapi. I will have to check how it shows up there :D

crates/iceberg-catalog/src/api/management/v1/warehouse.rs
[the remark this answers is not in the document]
twuebi: i'd go for enum since it's self descriptive, with string i as someone calling the api have to hope and pray that the return set is what i expect

crates/iceberg-catalog/src/api/management/mod.rs
```diff
@@ -1,41 +1,293 @@
...
+    /// Create a new warehouse in the given project. The project
+    /// of a warehouse cannot be changed after creation.
+    /// The storage configuration is validated by this method.
+    #[utoipa::path(
+        post,
+        tags = ["management"],
+        path = "management/v1/warehouse",
+        request_body = CreateWarehouseRequest,
+        responses(
+            (status = 200, description = "Warehouse created successfully", body = [CreateWarehouseResponse]),
```
twuebi, line 77: should we return a 201 here?

crates/iceberg-catalog/src/api/management/v1/warehouse.rs
```diff
@@ -26,29 +30,101 @@ pub struct CreateWarehouseResponse {
...
+pub struct RenameWarehouseRequest {
+    /// New name for the warehouse.
+    pub new_name: String,
+}
+
+#[derive(Debug, Clone, serde::Serialize, serde::Deserialize, ToSchema)]
+#[serde(rename_all = "kebab-case")]
+pub struct ListProjectsResponse {
+    /// List of project IDs.
+    pub project_ids: Vec<uuid::Uuid>,
```
twuebi, line 67: Let's make this a `Vec<Project>` with  `struct Project {id: Uuid}`, I'd imagine, we'd later want to extend `ListProjectsResponse` by a human readable project name / label or other fields, if we use a struct here we can make these changes non-breaking.

crates/iceberg-catalog/src/api/management/v1/warehouse.rs
```diff
@@ -75,6 +151,305 @@ pub trait WarehouseService<C: Catalog, A: AuthZHandler, S: SecretStore> {
...
+        if let Some(projects) = projects {
+            return Ok(ListProjectsResponse {
+                project_ids: projects.into_iter().map(|id| id.into_uuid()).collect(),
+            });
+        }
+
+        let projects = C::list_projects(context.v1_state.catalog).await?;
+        Ok(ListProjectsResponse {
+            project_ids: projects.into_iter().map(|id| id.into_uuid()).collect(),
+        })
```
twuebi, line 172: why does `A::check_list_projects` potentially return projects?

crates/iceberg-catalog/src/api/management/v1/warehouse.rs
```diff
@@ -75,6 +151,305 @@ pub trait WarehouseService<C: Catalog, A: AuthZHandler, S: SecretStore> {
...
+    }
+
+    async fn list_warehouses(
+        request: ListWarehouseRequest,
+        context: ApiContext<State<A, C, S>>,
+        request_metadata: RequestMetadata,
+    ) -> Result<ListWarehouseResponse> {
+        // ------------------- AuthZ -------------------
+        let project_id = ProjectIdent::from(
+            request.project_id.ok_or(
```
twuebi, line 182: Are we using `ListWarehouseRequest` elsewhere? We could make the field required on the object already

crates/iceberg-catalog/src/api/router.rs
```diff
@@ -40,11 +43,16 @@ pub fn new_full_router<
...
     >();
     let management_routes = Router::new().merge(ApiServer::new_v1_router());
+
     maybe_add_auth(
         token_verifier,
         Router::new()
             .nest("/catalog/v1", v1_routes)
-            .nest("/management/v1", management_routes),
+            .nest("/management/v1", management_routes)
+            .merge(utoipa_swagger_ui::SwaggerUi::new("/swagger-ui").url(
```
twuebi, line 52: i think this puts swagger-ui under auth, do we want that?

twuebi: adding it outside of `maybe_add_auth` should make swagger & openapi json available without auth

crates/iceberg-catalog/src/api/management/mod.rs
c-thiel: Done

crates/iceberg-catalog/src/api/management/v1/warehouse.rs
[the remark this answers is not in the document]
c-thiel: Done

crates/iceberg-catalog/src/api/management/mod.rs
c-thiel: Done

crates/iceberg-catalog/src/api/management/v1/warehouse.rs
c-thiel: Done

crates/iceberg-catalog/src/api/management/v1/warehouse.rs
c-thiel: One deployment might have multiple projects. A typical user should not be able to see other projects that might belong to other tenants. For simple AuthZ solutions, that implement an "all or nothing" approach on list, this feature wouldn't be used. When using a Zanzibar based approach, we would use something like [Expand](https://openfga.dev/api/service#/Relationship%20Queries/Expand) to implement this function as "get my projects".

crates/iceberg-catalog/src/api/management/v1/warehouse.rs
c-thiel: I did it this way only to improve the error a bit

crates/iceberg-catalog/src/api/router.rs
c-thiel: Done

crates/iceberg-catalog/src/api/management/v1/warehouse.rs
twuebi: I was thinking about the openapi rep, is it shown as optional there too?

PR #132, opened 2024-06-24
fix(examples): add pipeline check for examples & fix examples

### LINKED ISSUE #127: Integration test fails sometimes due to keycloak not being healthy
Likely just a timeout being too short

### LINKED ISSUE #130: examples need to be updated
Examples currently don't work since the health check of server relies on bash being available, we need to put the changes from the test docker compose yaml to the examples docker compose yaml.

Also it would be a good idea to execute the example docker compose in an action.

twuebi:
Fixes #130 
Fixes #127

### DIFF
--- a/.github/workflows/test-examples.yml
+++ b/.github/workflows/test-examples.yml
@@ -77,9 +77,12 @@ jobs:
       - name: Test Examples
         run: |
           cd examples &&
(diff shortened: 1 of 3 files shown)


v-kessler (approved):
tested

PR #141, opened 2024-07-01
feat(views): authz interface for views & view-ident resolve

twuebi:
- add authz interfaces for views
- add calls to respective authz interface to service layer
- add view-ident -> view-uuid resolve fn to enable authz functionality

### DIFF
--- a/crates/iceberg-catalog/src/catalog/views.rs
+++ b/crates/iceberg-catalog/src/catalog/views.rs
@@ -5,10 +5,11 @@ use crate::api::iceberg::v1::{
 };
 use crate::request_metadata::RequestMetadata;
 use http::StatusCode;
-use std::vec;
+use iceberg::NamespaceIdent;
+use iceberg_ext::catalog::rest::ViewUpdate;
 
 use super::tables::{
-    maybe_body_to_json, validate_lowercase_property, validate_table_or_view_ident,
+    require_no_location_specified, validate_table_or_view_ident, validate_table_properties,
 };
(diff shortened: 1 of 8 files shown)


crates/iceberg-catalog/src/catalog/views.rs
```diff
@@ -30,7 +31,7 @@ impl<C: Catalog, A: AuthZHandler, S: SecretStore>
         validate_namespace_ident(&namespace)?;
 
         // ------------------- AUTHZ -------------------
-        A::check_list_tables(
```
c-thiel, line 33: I am thinking about just having a "check_list_tabular" and not separate between tables and views.
Initially I thought this would suffice, as Databases also only have a grant "CREATE", ... on a Namespace, that allows to create / read / ... any type of Object (View, Materialized View, Tabular,...) that supports this capability.

However, I came to the conclusion that implementing it with only one "tabular" type might be more complicated, as possible grants will differ in the future between views and tables (i.e. insert). Thus, the AuthZ system will have a notion of the type of the tabular it's working with. 
So I would leave it as is, but would be interested in your thoughts.

c-thiel (approved):

crates/iceberg-catalog/src/catalog/views.rs
twuebi: Hm, no strong feelings, this interface may also not necessarily dictate how we store tuples, right? That'd depend on the impls, in general I also agree that tables and views are kinda different so treating them as such makes sense

PR #160, opened 2024-07-12
Expose more ways to connect to postgres

twuebi:
enable either database string or fine-grained env vars


c-thiel (commented):
Lets add also some docs, then we are good to merge.

c-thiel (approved):

PR #186, opened 2024-07-19
feat(pagination): add pagination for namespaces & tables & views

### LINKED ISSUE #38: Add Pagination for List Requests
Implement Pagination for list_namespaces & list_tables & list_views

twuebi:
First try at pagination for namespaces + tables + views, some decisions I took:

1. we paginate back to front, i.e. first page has the oldest entries, that somewhat efficient polling for new items by paginating through all items until empty page and then keeping to poll until a non empty page is returned, if we paginate the other way around, you'd have to keep track of the newest item client side to know when to stop iterating again. We could also expose the order as some custom header.
2. ~I've set a max page size of 100 items, I'm not attached to that number but feel that there should be some sane max~ according to `crates/iceberg-ext/src/catalog/rest/namespace.rs:115`, we're supposed to return all pages if we get no page token, MAX_PAGE_SIZE is now set to i64::MAX
3. due to reasons, page size comes in as a i32, i clamp it to (1, MAX_PAGE_SIZE)

closes #38


c-thiel:
> First try at pagination for namespaces + tables + views, some decisions I took:
> 
> 1. we paginate back to front, i.e. first page has the oldest entries, that somewhat efficient polling for new items by paginating through all items until empty page and then keeping to poll until a non empty page is returned, if we paginate the other way around, you'd have to keep track of the newest item client side to know when to stop iterating again. We could also expose the order as some custom header.
> 2. ~I've set a max page size of 100 items, I'm not attached to that number but feel that there should be some sane max~ according to `crates/iceberg-ext/src/catalog/rest/namespace.rs:115`, we're supposed to return all pages if we get no page token, MAX_PAGE_SIZE is now set to i64::MAX
> 3. due to reasons, page size comes in as a i32, i clamp it to (1, MAX_PAGE_SIZE)
> 
> closes #38

I agree with all your decisions

c-thiel (approved):

PR #192, opened 2024-07-23
feat(secrets): add support for kv2 secret storage

twuebi:


crates/iceberg-catalog/src/config.rs
```diff
@@ -99,6 +99,26 @@ pub struct DynAppConfig {
...
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
+pub enum SecretBackend {
+    Vault,
+    Postgres,
+}
+
+#[derive(Clone, Serialize, Deserialize, PartialEq, Redact)]
+pub struct VaultConfig {
```
c-thiel, line 116: We should add those to the Readme (also secret backend selection)

crates/iceberg-catalog/src/implementations/kv2.rs
```diff
@@ -0,0 +1,307 @@
...
+    format!("secret/{secret_id}", secret_id = secret_id.as_uuid())
+}
+
+impl std::fmt::Debug for SecretsState {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("SecretsState")
+            .field("vault_client", &"VaultClient")
+            .field("secret_mount", &self.secret_mount)
+            .field("vault_user", &self.vault_user)
+            .field("vault_password", &"REDACTED")
```
c-thiel, line 141: We are using veil in other places to redact - should we use it here as well?

twuebi: vault_client doesn't impl Debug, cursory read of veils docs makes me think we cannot use it here since it derives Debug for the struct?

crates/iceberg-catalog/src/config.rs
twuebi: updated readme

c-thiel (approved):

PR #217, opened 2024-07-27
fix(secrets): Add Encryption Secret for postgres SecretStore to README & warn on startup

c-thiel:


crates/iceberg-catalog/src/config.rs
```diff
@@ -30,6 +31,9 @@ lazy_static::lazy_static! {
         config.s3_signer_uri_for_warehouse(WarehouseIdent::from(uuid::Uuid::new_v4()));
         config.base_uri_catalog();
         config.base_uri_management();
+        if config.secret_backend == SecretBackend::Postgres && config.pg_encryption_key == DEFAULT_ENCRYPTION_KEY {
+            tracing::warn!("THIS IS UNSAFE! Using default encryption key for secrets in postgres, please set a proper key using ICEBERG_REST__PG_ENCRYPTION_KEY environment variable.");
```
c-thiel, line 35: @twuebi or should we rather make it neon blinking without using trace?

twuebi (approved):

PR #223, opened 2024-07-31
feat(storage): azdls

twuebi:
azure datalake storage support, currently relies on forks of iceberg-rust -> opendal -> reqsign

we're also pulling in a lot of azure sdk dependencies just for the sas token stuff + validating the storage profile, once FileIO is up to speed we may wanna reconsider this, reqsign also has some sas facilities.

### DIFF
--- a/crates/iceberg-catalog/src/catalog/io.rs
+++ b/crates/iceberg-catalog/src/catalog/io.rs
@@ -2,6 +2,7 @@ use crate::api::{ErrorModel, Result};
 use crate::service::storage::path_utils;
 use flate2::{write::GzEncoder, Compression};
 use iceberg::io::FileIO;
+use iceberg_ext::catalog::rest::IcebergErrorResponse;
 use serde::Serialize;
 use std::io::Write;
 
@@ -9,7 +10,7 @@ pub(crate) async fn write_metadata_file(
     metadata_location: &str,
     table_metadata: impl Serialize,
     file_io: &FileIO,
-) -> Result<()> {
+) -> Result<(), IoError> {
     tracing::debug!("Received location: {}", metadata_location);
     let metadata_location = if metadata_location.starts_with("abfs") {
         path_utils::reduce_scheme_string(metadata_location, false)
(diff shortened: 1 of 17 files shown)


crates/iceberg-catalog/src/service/storage/az.rs
```diff
@@ -0,0 +1,520 @@
...
+    #[must_use]
+    pub fn tabular_location(
+        &self,
+        namespace_id: NamespaceIdentUuid,
+        tabular_id: TabularIdentUuid,
+    ) -> String {
+        format!("{}{tabular_id}", self.namespace_location(namespace_id))
+    }
+
+    fn namespace_location(&self, namespace_id: NamespaceIdentUuid) -> String {
```
c-thiel, line 211: Same here as for `tabular_location`, lets rename to `initial_`, ...

crates/iceberg-catalog/src/service/storage/az.rs
```diff
@@ -0,0 +1,520 @@
...
+        format!("{}{namespace_id}/", self.location())
+    }
+
+    fn location(&self) -> String {
```
c-thiel, line 215: See above

crates/iceberg-catalog/src/service/storage/az.rs
```diff
@@ -0,0 +1,520 @@
...
+            client_id,
+            tenant_id,
+            client_secret,
+        } = creds;
+        let http_client = azure_core::new_http_client();
+        let token = azure_identity::ClientSecretCredential::new(
+            http_client,
+            self.authority_host
+                .clone()
+                .unwrap_or(Url::parse("https://login.microsoftonline.com").unwrap()),
```
c-thiel, line 263: Let's double check if this is constant also for private link and special azure clouds (i.e. government)

crates/iceberg-catalog/src/service/storage/az.rs
```diff
@@ -0,0 +1,520 @@
...
+                    Some(Box::new(e)),
+                )
+                .into());
+            }
+        }
+
+        Ok(())
+    }
+
+    async fn get_sas_token(&self, cred: StorageCredentials) -> Result<String> {
```
c-thiel, line 395: We need to add the Table / View Location here. The token is currently too wide.

crates/iceberg-catalog/src/catalog/io.rs
```diff
@@ -9,6 +10,14 @@ pub(crate) async fn write_metadata_file(
     table_metadata: impl Serialize,
     file_io: &FileIO,
 ) -> Result<()> {
+    tracing::debug!("Received location: {}", metadata_location);
+    let metadata_location = if metadata_location.starts_with("abfs") {
+        path_utils::sanitize_azdls_path(metadata_location)
```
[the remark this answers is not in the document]
twuebi: Good catch with the panic, I missed the unwrap on capture

crates/iceberg-catalog/src/service/storage/az.rs
```diff
@@ -0,0 +1,520 @@
...
+        return Err(ErrorModel::bad_request(
+            "Bucket name can consist only of lowercase letters, numbers, dots (.), and hyphens (-).",
+            "InvalidBucketName",
+            None,
+        ).into());
+    }
+
+    // Bucket names must begin and end with a letter or number.
+    // Unwrap will not fail as the length is already checked.
+    if !bucket.chars().next().unwrap().is_ascii_alphanumeric()
```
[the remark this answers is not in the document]
twuebi: the validate bucket fn is for now just a copy-paste from s3, it probably has different rules anyways

crates/iceberg-catalog/src/service/storage/az.rs
twuebi: Location is fixed, no? We don't allow updating if key_prefix changes, under which circumstances can the warehouse location change?

crates/iceberg-catalog/src/service/storage/az.rs
twuebi: gov & chinese clouds differ which is why we take it as a param but I feel this is an ok default value

crates/iceberg-catalog/src/service/storage/az.rs
```diff
@@ -0,0 +1,520 @@
...
+
+    Ok(())
+}
+
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize, utoipa::ToSchema)]
+#[allow(clippy::module_name_repetitions)]
+#[schema(rename_all = "kebab-case")]
+#[serde(rename_all = "kebab-case")]
+pub struct AzProfile {
+    /// Name of the S3 bucket
```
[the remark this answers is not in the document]
twuebi: fixed

crates/iceberg-catalog/src/service/storage/az.rs
twuebi: renamed

crates/iceberg-catalog/src/service/storage/az.rs
twuebi: changed to be more narrow, now we're taking location here

crates/iceberg-catalog/src/catalog/io.rs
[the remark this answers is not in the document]
c-thiel: Shouldn't we in this case always store abfss://filesystem@account_name/path as path and just strip abfss:// where needed? I just want to make have the data we store as well-defined as possbible, so we can more specific functions than arbitrary sanitazation (i.e. strip_filesystem_prefix) instead.

crates/iceberg-catalog/src/service/storage/mod.rs
```diff
@@ -59,29 +67,54 @@ impl StorageProfile {
...
 
     /// Create a new file IO instance for the storage profile.
     ///
     /// # Errors
     /// Fails if the underlying storage profile's file IO creation fails.
-    pub fn file_io(&self, secret: Option<&StorageCredential>) -> Result<iceberg::io::FileIO> {
+    pub fn file_io(
+        &self,
+        secret: Option<&StorageCredential>,
+        validating: bool,
```
c-thiel, line 97: I don't like to add this too much just to set the error code. Couldn't we just throw the more used error code by default and overwrite on the outer scope if needed? It doesn't feel worth adding an argument.

crates/iceberg-catalog/src/service/storage/az.rs
```diff
@@ -0,0 +1,536 @@
...
+use iceberg::io::azdls;
+use serde::{Deserialize, Serialize};
+use std::collections::HashMap;
+use std::sync::Arc;
+use std::time::SystemTime;
+use url::Url;
+use uuid::Uuid;
+use veil::Redact;
+
+fn is_valid_bucket_name(bucket: &str) -> Result<()> {
```
c-thiel, line 24: Let's call this "is_valid_container_name" maybe? Or "is_valid_filesystem"?

crates/iceberg-catalog/src/service/storage/az.rs
```diff
@@ -0,0 +1,536 @@
...
+    if bucket.len() < 3 || bucket.len() > 63 {
+        return Err(ErrorModel::bad_request(
+            "Bucket name must be between 3 and 63 characters long.",
+            "InvalidBucketName",
+            None,
+        )
+        .into());
+    }
+
+    // Bucket names can consist only of lowercase letters, numbers, dots (.), and hyphens (-).
```
c-thiel, line 35: Are we sure dots can be in there?

I am getting:
This name may only contain lowercase letters, numbers, and hyphens, and must begin with a letter or a number. Each hyphen must be preceded and followed by a non-hyphen character. The name must also be between 3 and 63 characters long.

crates/iceberg-catalog/src/service/storage/az.rs
```diff
@@ -0,0 +1,536 @@
...
+            )
+        } else {
+            format!(
+                "abfss:/{}@{}.{}/",
+                self.filesystem,
+                self.account_name,
+                self.endpoint_suffix
+                    .as_deref()
+                    .unwrap_or(DEFAULT_ENDPOINT_SUFFIX)
+                    .trim_end_matches('/'),
```
c-thiel, line 255: I am not sure if this could be used maliciously, but maybe we should make sure there are no "/" altogether in the account_name and endpoint_suffix?

crates/iceberg-catalog/src/service/storage/az.rs
```diff
@@ -0,0 +1,536 @@
...
+        );
+        let cred = azure_storage::StorageCredentials::token_credential(Arc::new(token));
+        let mut config = HashMap::new();
+
+        let sas = self.get_sas_token(table_location, cred).await?;
+        config.insert(self.iceberg_sas_property_key(), sas);
+        Ok(config)
+    }
+
+    /// Create a new `FileIO` instance for S3.
```
c-thiel, line 296: ADLS

c-thiel (commented):
We should also update the openapi-spec.

crates/iceberg-catalog/src/service/storage/mod.rs
twuebi: Ideally we'd not deal with error codes on this layer but just return an error which the respective handler then converts into the api response, relates to the revisit error handling issue, this module seems isolated enough to give it a try here.

crates/iceberg-catalog/src/service/storage/az.rs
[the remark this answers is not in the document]
twuebi: replaced by unwrap_or(false)

crates/iceberg-catalog/src/service/storage/error.rs
```diff
@@ -0,0 +1,173 @@
...
+        match value {
+            TableConfigError::Credentials(e) => e.into(),
+        }
+    }
+}
+
+#[derive(thiserror::Error, Debug)]
+pub enum UpdateError {
+    #[error("Field `{0}` cannot be updated to prevent dataloss.")]
+    Field(String),
```
c-thiel, line 104:
```suggestion
    FieldUnchangable(String),
```

crates/iceberg-catalog/src/service/storage/mod.rs
```diff
@@ -142,9 +162,23 @@ impl StorageProfile {
...
+                profile
+                    .generate_table_config(
+                        warehouse_id,
+                        table_id,
+                        namespace_id,
+                        data_access,
+                        table_location,
+                        secret
+                            .ok_or_else(|| {
+                                CredentialsError::MissingCredential(self.storage_type())
```
c-thiel, line 179: By doing this we rule out the possibility of a credential-free adls - completely public. I am not sure if this is relevant- would the storage driver support it?

c-thiel (approved):

crates/iceberg-catalog/src/service/storage/mod.rs
twuebi: Reqsign blows up if there's no credential, even worse, it'll take whatever it can find in env if nothing is passed to it which may cause really weird behavior if some az creds end up mounted into a catalog container, something like a profile being created for a warehouse based on the catalog provider's az account.

c-thiel (approved):

Issue #156, opened 2024-07-08
Add support for disabled metadata compression


twuebi:
Metadata properties can turn off compression via `write.metadata.compression-codec`, currently, we only allow gzip, we should also allow disabled compression.

PR #235, opened 2024-08-04
feat(compression-codec): Support setting and altering write.metadata.compression-codec

Grongrilla:


brgr-s:
There are a bunch of TODO comments in places where I wasn't sure if my implementation is good enough

crates/iceberg-catalog/src/catalog/io.rs
```diff
@@ -1,14 +1,16 @@
...
 use iceberg_ext::catalog::rest::IcebergErrorResponse;
 use serde::Serialize;
 use std::io::Write;
 
+use super::CommonMetadata;
+
 pub(crate) async fn write_metadata_file(
     metadata_location: &str,
-    table_metadata: impl Serialize,
+    metadata: impl CommonMetadata + Serialize,
```
twuebi, line 13: Let's pass the compression codec into here so we don't have to verify properties in this function. I'm thinking something along the lines of:


```rust

#[derive(thiserror::Error, Debug)]
#[error("Unsupported compression codec: {0}")]
pub struct UnsupportedCompression(pub String);

#[derive(Debug, Clone, Copy)]
pub enum CompressionCodecs {
    Gzip,
    None,
}

impl CompressionCodecs {
    pub fn compress(&self, payload: &[u8]) -> Result<Vec<u8>, IoError> {
        match self {
            CompressionCodecs::Gzip => {
                let mut compressed_metadata = GzEncoder::new(Vec::new(), Compression::default());
                compressed_metadata.write_all(payload).map_err(IoError::FileCompression)?;
                compressed_metadata.finish().map_err(IoError::FileCompression)
            }
            CompressionCodecs::None => {
                Ok(payload.to_vec())
            }
        }
    }

    pub fn as_file_extension(&self) -> &'static str {
        match self {
            CompressionCodecs::Gzip => "gz",
            CompressionCodecs::None => "",
        }
    }
}

impl CompressionCodecs {
    pub fn try_from_metadata(val: impl CommonMetadata) -> Result<Self, UnsupportedCompression> {
        val.properties()
            .get(METADATA_COMPRESSION)
            .map(String::as_str)
            .map_or(Ok(CompressionCodecs::Gzip), |codec| {
                match codec {
                    "gzip"  => Ok(CompressionCodecs::Gzip),
                    "none" => Ok(CompressionCodecs::None),
                    _ => Err(UnsupportedCompression(codec.to_string())),
                }
            })
    }
}

pub(crate) async fn write_metadata_file(
    metadata_location: &str,
    metadata: impl Serialize,
    codec: CompressionCodecs,
    file_io: &FileIO,
) -> Result<(), IoError> {
...
}
```

That way, we can avoid passing the untyped (`HashMap<String, String>`) properties into the location fns in `crates/iceberg-catalog/src/service/storage/` and remove the duplication of functionality, thanks to the 
`as_file_extension` fn we can easily figure out what ending to use without having to deny unsupported codecs again over there.

By introducing another function on the CommonMetadata trait:

```rust
pub(crate) trait CommonMetadata {
    fn properties(&self) -> &HashMap<String, String>;
    fn compression(&self) -> Result<io::CompressionCodecs, UnsupportedCompression> {
        Ok(io::CompressionCodecs::try_from_metadata(self)?)
    }
}
```

we could then just extract the compression codec in the respective handler fns and short-circuit there. E.g. in

crates/iceberg-catalog/src/service/storage/mod.rs
```diff
@@ -45,6 +47,29 @@ pub enum StorageType {
     Test,
 }
 
+// write_metadata expects Serialize + CommonMetadata, so for "test writes"
+// we use a TestMetadata struct with empty properties
+// TODO: define somewhere else?
```
twuebi, line 52: We could either give it a `#[cfg(test)]` attr or move it into the test module further down

crates/iceberg-catalog/src/service/storage/mod.rs
```diff
@@ -45,6 +47,29 @@ pub enum StorageType {
...
+// TODO: current default is to compress metadata, introducing a tiny overhead for test writes
```
twuebi, line 53: I think that's fine

twuebi (changes requested):
Thanks for the nice PR! 

I'd like to avoid duplicating the logic to extract the codec, I think we could make use of an enum and have all the lower level functions take that enum instead of having them take the whole untyped hashmap, this way we can push the fallible ops a layer up which makes it a bit more intuitive to me.  I added some codesnippet in one of the review comments.

crates/iceberg-catalog/src/service/storage/mod.rs
twuebi: oh, I just realized we're using this for writing, I think location is fine then, if you're worried about it getting mixed up with other stuff you could move it into a `pub(crate) mod validation_data {}` or something

crates/iceberg-catalog/src/catalog/io.rs
```diff
@@ -28,19 +30,37 @@ pub(crate) async fn write_metadata_file(
...
-        .write_all(&buf)
-        .map_err(IoError::FileCompression)?;
-
-    let compressed_metadata = compressed_metadata
-        .finish()
-        .map_err(IoError::FileCompression)?;
+    let buf = serde_json::to_vec(&metadata).map_err(IoError::Serialization)?;
+    let do_compression = match metadata.properties().get(METADATA_COMPRESSION) {
+        Some(value) if value == "gzip" => true,
+        Some(value) if value == "none" => false,
```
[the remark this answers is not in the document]
Grongrilla: see comment below

crates/iceberg-catalog/src/catalog/io.rs
```diff
@@ -28,19 +30,37 @@ pub(crate) async fn write_metadata_file(
...
+        None => true, // this is TIPs' default
+        // TODO: This is not a IoError...
+        // Or silently "coerce" to TIPs' default "Gzip Compresion"
+        Some(other_value) => {
+            tracing::error!(
+                "Unexpted value {} for property {}",
+                other_value,
+                METADATA_COMPRESSION
+            );
+            panic!("Unexpted value {other_value} for property {METADATA_COMPRESSION}");
```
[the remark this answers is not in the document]
Grongrilla: Sounds good to me

crates/iceberg-catalog/src/catalog/tables.rs
```diff
@@ -1033,7 +1037,12 @@ where
     I: IntoIterator<Item = &'a String>,
 {
     for prop in properties {
-        if prop.starts_with("write.metadata") || prop.starts_with("write.data.path") {
+        // TODO: #150 says properties like write.metadata.previous-version-max or write.data.path
+        // or write.metadata.path should be supportedm but PR #197 introduces this check and
+        // deniues these (and similar) properties...
```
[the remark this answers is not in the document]
Grongrilla: Ok, so since passing in compression codec is now supported, the excecption below is fine?

crates/iceberg-catalog/src/catalog/io.rs
Grongrilla: Thanks for that suggesion. I was hesitant to indroduce "even more" new structs or enums, but I like your suggestion much more than passing in a HashMap.

crates/iceberg-catalog/src/service/storage/mod.rs
Grongrilla: Yes, I was kind of worried that it is "too visible"

crates/iceberg-catalog/src/catalog/io.rs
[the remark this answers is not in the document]
twuebi: In general, handler threads of the server should never panic, so we always prefer an error over it

crates/iceberg-catalog/src/catalog/tables.rs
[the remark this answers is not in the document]
twuebi: yes

brgr-s:
I have implemented the changes that where suggested.

I extended UnsupportedCompressionCodec and changed the signature of write_metadata_file to only accept metadata and use those to create CompressionCodec

crates/iceberg-catalog/src/service/storage/mod.rs
Grongrilla: I ended up not moving it 😄

twuebi (changes requested):

crates/iceberg-catalog/src/catalog/io.rs
```diff
@@ -1,14 +1,110 @@
...
-use iceberg::io::FileIO;
+use iceberg::{io::FileIO, spec::view_properties::METADATA_COMPRESSION};
 use iceberg_ext::catalog::rest::IcebergErrorResponse;
 use serde::Serialize;
+use std::collections::HashMap;
 use std::io::Write;
 
+use super::CommonMetadata;
+
+#[derive(thiserror::Error, Debug, strum::IntoStaticStr)]
```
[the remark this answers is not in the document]
Grongrilla: Yes, its enough code to warrant its' own module

crates/iceberg-catalog/src/catalog/io.rs
```diff
@@ -1,14 +1,110 @@
...
+        let boxed = Box::new(value);
+        let message = boxed.to_string();
+
+        match value {
+            CompressionCodecError::UnsupportedCompressionCodec(_) => {
+                ErrorModel::bad_request(message, typ, Some(boxed)).into()
+            }
+        }
+    }
+}
```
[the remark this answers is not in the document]
Grongrilla: I don't expect other variants. I made it an enum because I wanted to mimic the existing code to create an `IcebergErrorResponse`, which was necessary because `CompressionCodec` is used in places where an `IcebergErrorResponse` is expected. But you are of course correct, a `struct` is more appropriate, and an `IcebergErrorResponse` can still be created.

crates/iceberg-catalog/src/catalog/io.rs
```diff
@@ -19,6 +115,9 @@ pub(crate) async fn write_metadata_file(
     };
     tracing::debug!("Going to write metadata file to {}", metadata_location);
 
+    let compression_codec =
+        CompressionCodec::try_from_metadata(&metadata).map_err(IoError::UnknownCompressionCodec)?;
+
```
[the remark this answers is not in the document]
Grongrilla: Currently, it would fail earlier, when the location for metadata is calculated using `fn initial_metadata_location`, which also needs a `CompressionCodec`. The compiler can't know this, nor can we be sure this will always be the case. I will change it to `compression_codec` beeing a parameter.

crates/iceberg-catalog/src/catalog/io.rs
```diff
@@ -1,14 +1,110 @@
...
+    ) -> Result<Self, CompressionCodecError> {
+        Self::try_from_properties(metadata.properties())
+    }
+}
+
+impl Default for CompressionCodec {
+    fn default() -> Self {
+        Self::Gzip
+    }
+}
```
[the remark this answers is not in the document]
Grongrilla: Thank you for that hint, I wasn't aware of that possibility. I did not get that suggestion from clippy and my IDE setup (nvim + rust-analyzer + rustaceanvim) should suggest something like this. Weird, because it absolutely makes sense to use that feature.

brgr-s:
I pushed some changes but there are two things tbd:

- I implemented `IoError::UnsupportedCompressionCodec`, but then realized, it is not needed -> should be removed, I think
- I changed the signature of `write_metadata_file` s. th. `metadata` is only `Serialize`, because a bound to `CommonMetadata` is no longer necessary...
  - which means that we do not need `TestMetadata` anymore. I left it in there because maybe it is usefull down the line? However for #156 it is not necessary anymore...
  - unless we want to keep the bound on `CommonMetadata` because it "feels weird" that `write_metadata_file` can potentially be used to write anything as a "metadata file"

twuebi:
> I pushed some changes but there are two things tbd:
> 
> * I implemented `IoError::UnsupportedCompressionCodec`, but then realized, it is not needed -> should be removed, I think
> * I changed the signature of `write_metadata_file` s. th. `metadata` is only `Serialize`, because a bound to `CommonMetadata` is no longer necessary...
>   
>   * which means that we do not need `TestMetadata` anymore. I left it in there because maybe it is usefull down the line? However for [Add support for disabled metadata compression #156](https://github.com/hansetag/iceberg-catalog/issues/156) it is not necessary anymore...

I think we can get rid of it then?

>   * unless we want to keep the bound on `CommonMetadata` because it "feels weird" that `write_metadata_file` can potentially be used to write anything as a "metadata file"

IMO, we should conside renaming that fn at some point, there's no reason to limit it to metadata writes, it's a compressible write

crates/iceberg-catalog/src/catalog/tables.rs
```diff
@@ -492,9 +504,12 @@ impl<C: Catalog, A: AuthZHandler, S: SecretStore>
             .storage_config
             .storage_profile
             .file_io(storage_secret.as_ref())?;
+        let compression_codec =
```
twuebi, line 507: let's move this to line 447, then we don't end up calling db and all other things if we're gonna reject the request due to unsupported compression anyways

crates/iceberg-catalog/src/catalog/views/commit.rs
```diff
@@ -253,21 +254,30 @@ pub(crate) async fn commit_view<C: Catalog, A: AuthZHandler, S: SecretStore>(
...
-        storage_profile.initial_metadata_location(&view_location, Uuid::now_v7());
+    let requested_update_metadata = m.build().map_err(|e| {
+        ErrorModel::builder()
+            .code(StatusCode::BAD_REQUEST.into())
+            .message(format!("Error building metadata: {e}"))
+            .r#type("BuildMetadataError".to_string())
+            .build()
+    })?;
+
+    // TODO: Check this: this only works iff compression cannot be set to none on
```
twuebi, line 265: does this TODO still apply?

Grongrilla: I think so. Unfortunately, I am not familiar enough with the codebase to be able to really tell.

This might even not be the only location in code that has this problem, and if my assumptions are correct, we might have a problem every time a Table or View is changed without changing metadata compression and when metadata compression is set to the non-default value.

Assuming a Table or View is created using "none" for metadata compression, or that value was at some point set to "none".

Then a Table or View is updated, but there is no change to the metadata compression property. Each codepath that creates `CompressionCodec` from properties supplied in the request will then assume that compression codec is "gzip", as this is the default if no value for that property is supplied in `try_from_properties`.

My assumptions might be wrong, and I did not have time yet to check this thouroughly... I think I would not be able to tell if I caught every possible path in the code, or if this is a non-issue because there is some other safeguard or codepath that prevents this from happening....

I think this kinda boils down to #156 maybe not beeing a "good first issue" :wink:

crates/iceberg-catalog/src/catalog/tables.rs
Grongrilla: I moved it to 484 because at that point the "correct" "one and only" result is available.

crates/iceberg-catalog/src/catalog/views/commit.rs
twuebi: Could it be that the name of `initial_metadata_location` is a bit confusing? I'm not sure I understand the issue, the function interpolates the new location of the metadata based on view location  and some unique identifier. The initial in its name was an attempt to signal that the metadata location of a table may be different from what this function tells you, hence on reads, we need to take the metadata location from the respective db row instead of simply calling initial_metadata_location and returning that to a caller. 

It probably makes sense to rename the function to new_metadata_location. We've just recently renamed it from metadata_location to initial_metadata_location but it seems it's still confusing.

crates/iceberg-catalog/src/catalog/tables.rs
twuebi: You're right, then let's not do that, the updates may , I also just saw that we are already extracting the codec in `implementations/postgres/tabular/table.rs:575` which kinda short circuits the function. To fully avoid doing unnecessary DB roundtrips, we'd have to check if `request.updates` has a `TableUpdate::SetProperties` where `updates` contains a unsupported codec.

Grongrilla: The properties of the request are put into `validate_table_properties`, however that function just checks if there is an unsupported property and if it is in lowercase. 

Might it be a good idea to extend that function to check for valid values for properties, where possible? 

... or should a `CommitTableRequest` already be valid at that point, at least as far as it is possible to check? It should be possible to implement that by `impl Deserialize for CreateTableRequest`, but I am not sure if it should be done like this.

I think implementing a solution for this might warrant its' own issue.

crates/iceberg-catalog/src/catalog/views/commit.rs
Grongrilla: I just re-read the code. The issue is plain and simple that I overlooked that in `commit_view`, the old metadata are loaded, the builder is then initialized from that old metadata, and the updated metadata are a result of feeding the updates from the request to the builder. I don't know how I overlooked that, but I did, and the comment is simply wrong. My bad, sorry about the confusion.

crates/iceberg-catalog/src/catalog/tables.rs
twuebi: sure, let's revisit at a later point

twuebi (dismissed):
Thanks @Grongrilla! 

There's a small conflict left, once that's fixed we're good to go!

twuebi (approved):

PR #255, opened 2024-08-12
chore: Cleanup S3 signing

c-thiel:


crates/iceberg-catalog/Cargo.toml
```diff
@@ -83,4 +83,3 @@ rand = "0.8.5"
 http-body-util = { workspace = true }
 tower = { workspace = true }
 needs_env_var = { workspace = true }
-tracing-subscriber = "0.3.18"
```
twuebi, line 86: i kept this in since it is nice to be able to activate tracing within tests without having to go back to Cargo.toml every time

crates/iceberg-catalog/Cargo.toml
```diff
@@ -67,6 +67,7 @@ tower-http = { workspace = true, optional = true, features = [
     "util",
 ] }
 tracing = { workspace = true }
+tracing-subscriber = { workspace = true }
```
twuebi, line 70: let's move this down to dev-dependencies, i don't think it's needed for the crate itself

crates/iceberg-catalog/src/catalog/s3_signer.rs
```diff
@@ -350,184 +348,109 @@ async fn validate_table_method<A: AuthZHandler>(
...
-            .message("Table location does not have a bucket".to_string())
-            .r#type("TableLocationNoBucket".to_string())
+    let table_location = parse_s3_location(table_location)?;
+    let url_location = &parsed_url.location;
+
+    if !url_location.is_sublocation_of(&table_location) {
+        return Err(ErrorModel::builder()
+            .code(http::StatusCode::FORBIDDEN.into())
+            .message("Request URI does not match table location".to_string())
+            .r#type("RequestUriMismatch".to_string())
```
twuebi, line 365: change to the new interface? 
```
ErrorModel::forbidden("Request URI does not match table location", "RequestUriMismatch", None).append_details(vec![
                 format!("Expected Location: {table_location}"),
                 format!("Actual Location: {url_location}"),
             ]);
```

crates/iceberg-catalog/src/catalog/s3_signer.rs
```diff
@@ -350,184 +348,109 @@ async fn validate_table_method<A: AuthZHandler>(
...
+pub(super) mod s3_utils {
+    use super::{ErrorModel, Result};
+    use crate::service::storage::S3Location;
+    use lazy_regex::regex;
+
+    #[derive(Debug)]
+    pub(super) struct ParsedS3Url {
+        pub(super) location: S3Location,
+        // Used endpoint without the bucket
+        #[allow(dead_code)]
```
twuebi, line 386: from september we can use this instead of allow dead code:

https://github.com/rust-lang/rust/pull/120924

crates/iceberg-catalog/src/service/storage/error.rs
```diff
@@ -16,6 +16,12 @@ pub enum ValidationError {
     },
     #[error(transparent)]
     FileIoError(#[from] FileIoError),
+    #[error("{reason}")]
+    InvalidS3Location {
```
twuebi, line 20: InvalidLocation with maybe an internal marker for what storage it belongs to? Else we may end up with a variant per supported backend

crates/iceberg-catalog/Cargo.toml
c-thiel: Done

crates/iceberg-catalog/src/service/storage/error.rs
c-thiel: Done

crates/iceberg-catalog/src/catalog/s3_signer.rs
c-thiel: Done

twuebi (approved):

PR #257, opened 2024-08-12
fix!: dots can no longer be used in namespace names

### LINKED ISSUE #242: BREAKING CHANGE: Prohibit dots "." in namespace names
The servlet spec changed and does not allow the previous namespace separator anymore.
Current consensus in the community is to make the character configurable.

I believe for now the most consistent way is to reserve "." as a separator. Using "." is discouraged in identifier names anyway as they always require escaping in SQL which is error prone.

Related info: 
New canonicalization: https://jakarta.ee/specifications/servlet/6.0/jakarta-servlet-spec-6.0.html#uri-path-canonicalization
Iceberg Issue: https://github.com/apache/iceberg/issues/10338
Iceberg PR: https://github.com/apache/iceberg/pull/10877/files

twuebi:
closes #242

### DIFF
--- a/crates/iceberg-catalog/src/catalog/namespace.rs
+++ b/crates/iceberg-catalog/src/catalog/namespace.rs
@@ -7,6 +7,7 @@ use crate::request_metadata::RequestMetadata;
 use crate::CONFIG;
 use http::StatusCode;
 use iceberg::NamespaceIdent;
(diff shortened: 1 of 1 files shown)


c-thiel (approved):
Should we add a migration to

PR #262, opened 2024-08-13
chore: Add integration tests for trino

c-thiel:
Trino requires support for `location` on Namespace & Table Level for data access


c-thiel:
Added via https://github.com/hansetag/iceberg-catalog/pull/267

PR #267, opened 2024-08-13
chore: Typesafe(er) Locations

c-thiel:
BEGIN_COMMIT_OVERRIDE
feat: Add support for custom Locations for Namespaces & Tables
chore: Typesafer Locations and configuration attributes
chore: Refactor `load_table` to `load_tables` to serve more use-cases
chore: Move table commit & transaction to `catalog` from `implementations`
chore: Add tests for trino (Vended-Credentials only, Remote Signing is currently not supported)
chore: Use OAuth2 client credential flow for Spark in tests
END_COMMIT_OVERRIDE


crates/iceberg-catalog/src/catalog/namespace.rs
```diff
@@ -82,16 +94,18 @@ impl<C: Catalog, A: AuthZHandler, S: SecretStore>
...
-        A::check_create_namespace(
-            &request_metadata,
-            warehouse_id,
-            request.namespace.parent().as_ref(),
-            state.v1_state.auth,
-        )
-        .await?;
+        let mut t = C::Transaction::begin_read(state.v1_state.catalog.clone()).await?;
+        let warehouse = C::get_warehouse(warehouse_id, t.transaction()).await?;
+        drop(t);
```
twuebi, line 99: why not create a write tx from line 97 and keep using it until line 111?

c-thiel:
@twuebi I added a few ToDo Tobi where your input would be appreciated :)
Configurable locations will come in a separate PR, this lays the foundation and contains a lot of maintenance.

crates/iceberg-catalog/src/catalog/namespace.rs
```diff
@@ -312,3 +338,160 @@ pub(crate) fn validate_namespace_ident(namespace: &NamespaceIdent) -> Result<()>
...
+    removals: Option<Vec<String>>,
+) -> (HashMap<String, String>, UpdateNamespacePropertiesResponse) {
+    let mut properties = previous_properties.unwrap_or_default();
+
+    let mut changes_updated = vec![];
+    let mut changes_removed = vec![];
+    let mut changes_missing = vec![];
+
+    if let Some(removals) = removals {
+        for key in removals {
```
twuebi, line 387: `for key in removals.unwrap_or_default().into_iter()` ?

crates/iceberg-catalog/src/catalog/views/commit.rs
```diff
@@ -274,13 +283,11 @@ pub(crate) async fn commit_view<C: Catalog, A: AuthZHandler, S: SecretStore>(
...
         metadata_location.as_str(),
-        requested_update_metadata,
+        requested_update_metadata.clone(),
         transaction.transaction(),
     )
     .await?;
-    let updated_meta = C::load_view(view_id, transaction.transaction())
-        .await?
-        .metadata;
+    // ToDo Tobi: Why did we need to load the metadata again?
```
twuebi, line 290: not sure, it may be an artifact of when we were doing partial updates on db tables instead of the load-modify-delete-create flow we're doing now

crates/iceberg-catalog/src/catalog/views/create.rs
```diff
@@ -94,20 +94,21 @@ pub(crate) async fn create_view<C: Catalog, A: AuthZHandler, S: SecretStore>(
 
     let view_id: TabularIdentUuid = TabularIdentUuid::View(uuid::Uuid::now_v7());
 
-    let view_location = storage_profile.initial_tabular_location(namespace_id, view_id);
+    let namespace_location = storage_profile.default_namespace_location(namespace_id)?;
+    let view_location = storage_profile.default_tabular_location(&namespace_location, view_id);
     let mut request = request;
-    let metadata_location = storage_profile.initial_metadata_location(
+    let metadata_location = storage_profile.metadata_location(
```
twuebi, line 100: default_metadata_location instead?

twuebi: It was originally called metadata_location but we decided to rename to signal that it may be giving you something else than what's written into the db already, the `initial_` prefix caused some confusion in #156, i saw you used `default` for namespace location already, what do you think?

crates/iceberg-catalog/src/catalog/namespace.rs
c-thiel: fixed

crates/iceberg-catalog/src/service/storage/az.rs
```diff
@@ -47,46 +47,40 @@ pub struct AzdlsProfile {
...
-        if endpoint_suffix.as_deref().is_some_and(|s| s.contains('/')) {
-            return Err(ValidationError::InvalidProfile {
-                source: None,
-                reason: "Storage Profile `endpoint_suffix` must not contain slashes.".to_string(),
-                entity: "EndpointSuffix".to_string(),
-            });
+        if let Some(key_prefix) = self.key_prefix.as_ref() {
+            if key_prefix.is_empty() {
+                self.key_prefix = None;
+            }
```
twuebi, line 79: let's maybe move these fns a bit down in the impl block, I tend to read files top to bottom and I find it easier to understand what's going on if the fns are ordered from high abstraction to lower level fns

crates/iceberg-catalog/src/service/storage/az.rs
```diff
@@ -334,86 +261,34 @@ impl AzdlsProfile {
...
                 SystemTime::now().into(),
                 SystemTime::now()
                     .checked_add(std::time::Duration::from_secs(
                         self.sas_token_validity_seconds.unwrap_or(3600),
                     ))
-                    .unwrap()
+                    .unwrap_or_else(|| {
+                        tracing::error!("Failed to add validity seconds to current time.");
+                        SystemTime::now()
+                    })
```
twuebi, line 283: This would create a sas token that is valid from now() until now() which is expired when we return it

crates/iceberg-catalog/src/service/storage/az.rs
```diff
@@ -536,72 +630,132 @@ fn is_valid_container_name(container: &str) -> Result<(), ValidationError> {
...
+    }
+
+    Ok(())
+}
+
+fn normalize_endpoint_suffix(
+    endpoint_suffix: Option<String>,
+) -> Result<Option<String>, ValidationError> {
+    // If endpoint suffix is Some(""), set it to None.
+    if let Some(endpoint_suffix) = endpoint_suffix {
```
twuebi, line 692: we could reduce nesting a bit by lifting the if let Some into the calling location and making this fn take a String instead of Option<String>

crates/iceberg-catalog/src/service/storage/az.rs
```diff
@@ -536,72 +630,132 @@ fn is_valid_container_name(container: &str) -> Result<(), ValidationError> {
...
+            if endpoint_suffix.contains('/') {
+                return Err(ValidationError::InvalidProfile {
+                    source: None,
+                    reason: "`endpoint_suffix` must not contain slashes.".to_string(),
+                    entity: "EndpointSuffix".to_string(),
+                });
+            }
+
+            // Endpoint suffix must be a valid hostname
+            if Host::parse(&endpoint_suffix).is_err() {
```
twuebi, line 706: should we make endpoint_suffix Host instead?

crates/iceberg-catalog/src/service/storage/mod.rs
```diff
@@ -118,34 +119,39 @@ impl StorageProfile {
...
         match self {
-            StorageProfile::S3(profile) => profile.tabular_location(namespace_id, table_id),
-            StorageProfile::Azdls(profile) => {
-                profile.initial_tabular_location(namespace_id, table_id)
-            }
+            StorageProfile::S3(profile) => profile.base_location().map(Into::into),
+            StorageProfile::Azdls(profile) => profile.base_location(),
             #[cfg(test)]
-            StorageProfile::Test(_) => format!("/tmp/{namespace_id}/{table_id}"),
+            StorageProfile::Test(_) => std::str::FromStr::from_str("file://tmp/").map_err(|_| {
```
twuebi, line 131: Location::from_str(...)?

crates/iceberg-catalog/src/service/storage/mod.rs
```diff
@@ -202,29 +205,143 @@ impl StorageProfile {
...
+                }
+                StorageProfile::Azdls(_) => {
+                    az::validate_vended_credentials(&tbl_config, &test_location, self).await?;
+                }
+                #[cfg(test)]
+                StorageProfile::Test(_) => {}
+            }
+        }
+
+        println!("------------- before remove_all --------------");
```
twuebi, line 286: replace with a trace::debug!("Done with..., doing... now")?

crates/iceberg-catalog/src/service/storage/s3.rs
```diff
@@ -40,6 +39,8 @@ pub struct S3Profile {
     /// Region to use for S3 requests.
     pub region: String,
     /// Path style access for S3 requests.
+    /// If not specified, the client decides. If the underlying S3
```
twuebi, line 42: Which client? The client of the s3 crate?

crates/iceberg-catalog/src/service/storage/s3.rs
```diff
@@ -608,13 +516,44 @@ impl S3Profile {
...
+    let mut builder = iceberg::io::FileIOBuilder::new("s3");
+
+    for key in [
+        iceberg::io::S3_REGION,
+        iceberg::io::S3_ENDPOINT,
+        iceberg::io::S3_ACCESS_KEY_ID,
+        iceberg::io::S3_SECRET_ACCESS_KEY,
+        iceberg::io::S3_SESSION_TOKEN,
+    ] {
+        if let Some(value) = config.get_custom_prop(key) {
```
twuebi, line 531: get_custom_prop removes type-check on read, this means we defer failure to opendal / reqsign

crates/iceberg-ext/src/configs/namespace.rs
```diff
@@ -0,0 +1,129 @@
+use super::{ConfigParseError, ConfigProperty, Location, NotCustomProp, ParseFromStr};
+use std::collections::HashMap;
+use std::fmt::Debug;
+
+#[derive(Debug, PartialEq, Default)]
+#[allow(clippy::module_name_repetitions)]
```
twuebi, line 6: let's maybe put a `#![allow(clippy::module_name_repetitions)]` on `lib.rs` or at least `mod.rs`, I don't think this is a helpful lint, we have a lot of places where we allow it by now

crates/iceberg-catalog/src/service/storage/az.rs
c-thiel: done

crates/iceberg-catalog/src/service/storage/az.rs
```diff
@@ -334,86 +261,34 @@ impl AzdlsProfile {
...
+        path: &Location,
         cred: StorageCredentials,
         permissions: StoragePermissions,
     ) -> Result<String, CredentialsError> {
         let client =
             azure_storage_blobs::prelude::BlobServiceClient::new(self.account_name.as_str(), cred);
         let delegation_key = client
             .get_user_deligation_key(
-                // TODO: should we pull in `time` instead?
+                // ToDo Tobi: Use time or chrono?
```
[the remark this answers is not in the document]
c-thiel: Moved to time

crates/iceberg-catalog/src/service/storage/az.rs
c-thiel: Fixed - throws an error now

crates/iceberg-catalog/src/service/storage/az.rs
c-thiel: Done - call is quite a bit uglier now but I think its worth it.

crates/iceberg-catalog/src/service/storage/mod.rs
c-thiel: I had that first - problem is that FromStr is not `use`ed by default, leading to a conditional use which I like even less.

crates/iceberg-catalog/src/service/storage/s3.rs
c-thiel: Every client does not get this info. Thus, most clients would use bucket style access unless the endpoint is not a valid domain name (i.e. an IP Address)

crates/iceberg-catalog/src/service/storage/s3.rs
c-thiel: This is already typechecked from insertion - I switched to our own types though.

crates/iceberg-ext/src/configs/namespace.rs
c-thiel: We already have it on lib level - just old habit

crates/iceberg-catalog/src/service/storage/mod.rs
```diff
@@ -202,29 +205,143 @@ impl StorageProfile {
...
+    ) -> Result<(), ValidationError> {
+        let compression_codec = CompressionCodec::Gzip;
+
+        let mut test_file_write =
+            self.metadata_location(test_location, &compression_codec, uuid::Uuid::now_v7());
+        if is_vended_credentials {
+            test_file_write.pop().push("test");
+            tracing::debug!("Validating access to: {}", test_file_write);
+        } else {
+            test_file_write.push("test_vended_credentials");
```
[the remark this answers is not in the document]
c-thiel: Fixed

crates/iceberg-catalog/src/catalog/namespace.rs
```diff
@@ -312,3 +336,158 @@ pub(crate) fn validate_namespace_ident(namespace: &NamespaceIdent) -> Result<()>
...
+            "LocationCannotBeRemoved".to_string(),
+            None,
+        )
+        .into());
+    }
+
+    if let Some(location) = updates.as_ref().and_then(|u| u.get(Location::KEY)) {
+        return Err(ErrorModel::bad_request(
+            "Namespace property `location` cannot be updated.",
+            "LocationCannotBeUpdated".to_string(),
```
twuebi, line 434: the .to_string() can probably go

crates/iceberg-catalog/src/catalog/tables.rs
```diff
@@ -993,16 +920,96 @@ impl<C: Catalog, A: AuthZHandler, S: SecretStore>
...
-            .build()
+fn determine_table_ident(
+    parameters_ident: TableIdent,
+    request_ident: Option<TableIdent>,
+) -> Result<TableIdent> {
+    // When querying a branch, spark sends something like the following as part of the `parameters`:
+    // namespace: (<my>, <namespace>, <table_name>)
+    // table_name: branch_<branch_name>
+
+    if let Some(identifier) = request_ident {
```
twuebi, line 931: can we somehow reduce nesting here and maybe clarify the table_name_candidate part a bit, perhaps with a comment?

crates/iceberg-catalog/src/service/storage/az.rs
```diff
@@ -334,86 +225,38 @@ impl AzdlsProfile {
...
             .get_user_deligation_key(
-                // TODO: should we pull in `time` instead?
-                SystemTime::now().into(),
-                SystemTime::now()
-                    .checked_add(std::time::Duration::from_secs(
-                        self.sas_token_validity_seconds.unwrap_or(3600),
+                start,
+                start
+                    .checked_add(time::Duration::seconds(
+                        i64::try_from(self.sas_token_validity_seconds.unwrap_or(3600)).map_err(
```
twuebi, line 242: i think we could make this effectively infallible by clamping `self.sas_token_validity_seconds` to i64::MAX

crates/iceberg-catalog/src/service/storage/s3.rs
```diff
@@ -572,49 +407,169 @@ impl S3Profile {
...
+        let key = table_location.key().join("/");
+        let key = format!("{key}/");
+
+        Ok(format!(
             r#"{{
         "Version": "2012-10-17",
         "Statement": [
             {{
-                "Sid": "{}",
+                "Sid": "TableAccess",
```
twuebi, line 436: I vaguely recall a uniqueness requirement on Sid, could be that it's not relevant for inline policies, are you aware of this?

c-thiel: It only needs to be unique inside one policy.

Docs:
You can use the Sid value as a description for the policy statement. In services that let you specify an ID element, such as SQS and SNS, the Sid value is just a sub-ID of the policy document ID. In IAM, the Sid value must be unique within a JSON policy.

crates/iceberg-catalog/src/catalog/namespace.rs
c-thiel: Fixed

crates/iceberg-catalog/src/catalog/tables.rs
c-thiel: Do you have ideas?
I extended the comment.

twuebi: Something like this?:

```rust
fn determine_table_ident(
    parameters_ident: TableIdent,
    request_ident: Option<TableIdent>,
) -> Result<TableIdent> {
    // When querying a branch, spark sends something like the following as part of the `parameters`:
    // namespace: (<my>, <namespace>, <table_name>)
    // table_name: branch_<branch_name>
    let Some(identifier) = request_ident else {
        return Ok(parameters_ident);
    };

    if identifier == parameters_ident {
        return Ok(identifier);
    }

    let ns_parts = parameters_ident.namespace.clone().inner();
    let table_name_candidate = if ns_parts.len() >= 2 {
        NamespaceIdent::from_vec(ns_parts.iter().take(ns_parts.len() - 1).cloned().collect())
            .ok()
            .map(|n| TableIdent::new(n, ns_parts.last().cloned().unwrap_or_default()))
    } else {
        None
    };

    if table_name_candidate != Some(identifier.clone()) {
        return Err(ErrorModel::builder()
            .code(StatusCode::BAD_REQUEST.into())
            .message(
                "Table identifier in path does not match the one in the request body".to_string(),
            )
            .r#type("TableIdentifierMismatch".to_string())
            .build()
            .into());
    }

    Ok(identifier)
}
```

c-thiel: Added

crates/iceberg-catalog/src/service/storage/az.rs
c-thiel: Done - we still fail on overflow. This is clearer than using `time::DateTimeOffset::max()` I think?

c-thiel:
I am happy enough now from my side. @twuebi  maybe another check from you would be good.

Trino currently does not support Materialized Views for the Rest Catalog, hence no tests.

twuebi (approved):

twuebi (approved):

PR #287, opened 2024-08-21
feat(tabular): drop-purge 

twuebi:
wip


crates/iceberg-catalog/src/catalog/tables.rs
```diff
@@ -274,7 +280,8 @@ impl<C: Catalog, A: AuthZHandler, S: SecretStore>
         // ------------------- BUSINESS LOGIC -------------------
         let table_id = require_table_id(&table, table_id)?;
         let mut t = C::Transaction::begin_read(state.v1_state.catalog).await?;
-        let mut metadatas = C::load_tables(warehouse_id, vec![table_id], t.transaction()).await?;
+        let mut metadatas =
+            C::load_tables(warehouse_id, vec![table_id], false, t.transaction()).await?;
```
c-thiel, line 284: I think it would be clearer to add a variable let include_deleted instead of the false here. We can re-use it also for the ListFlags above.

crates/iceberg-catalog/src/catalog/tables.rs
```diff
@@ -349,7 +356,10 @@ impl<C: Catalog, A: AuthZHandler, S: SecretStore>
         let table_id = C::table_ident_to_id(
             warehouse_id,
             &table_ident,
-            include_staged,
+            ListFlags {
+                include_staged,
+                include_deleted: false,
```
c-thiel, line 361: Same here, lets put it in a var and re-use for C::load_tables

crates/iceberg-catalog/src/catalog/tables.rs
```diff
@@ -474,7 +484,10 @@ impl<C: Catalog, A: AuthZHandler, S: SecretStore>
         let table_id = C::table_ident_to_id(
             warehouse_id,
             &table,
-            include_staged,
+            ListFlags {
+                include_staged,
+                include_deleted: false,
```
c-thiel, line 489: Var

crates/iceberg-catalog/src/catalog/tables.rs
```diff
@@ -722,7 +741,10 @@ impl<C: Catalog, A: AuthZHandler, S: SecretStore>
         let table_ids = C::table_idents_to_ids(
             warehouse_id,
             identifiers,
-            include_staged,
+            ListFlags {
+                include_staged,
+                include_deleted: false,
```
c-thiel, line 746: Var

crates/iceberg-catalog/src/implementations/postgres/tabular.rs
```diff
@@ -50,11 +50,15 @@ where
         AND n.warehouse_id = $3
         AND w.status = 'active'
         AND t.typ = $4
+        AND (t.deleted_at IS NULL OR $5)
+        AND (t.metadata_location IS NOT NULL OR $6)
```
c-thiel, line 54: This filter should be removed - the `if staged && !list_flags.include_staged` statement below can now no longer be hit - or we remove the statement below?

crates/iceberg-catalog/src/service/catalog.rs
```diff
@@ -363,6 +367,12 @@ where
     ) -> Result<()>;
 }
 
+#[derive(Debug, Clone, Default)]
```
c-thiel, line 370: PartialEq? Copy?

twuebi: good point

crates/iceberg-catalog/src/implementations/postgres/tabular.rs
twuebi: do you see value in the statement below?

twuebi: It'd probably still have to stay since the metadata_location in the returned struct is optional and we'd like to make sure, right?

c-thiel (approved):

PR #320, opened 2024-09-02
deps: Starrocks integration tests

c-thiel:


twuebi (approved):
We should look into consolidating some of the sql between the tests, i feel like we'll end up having a lot of drift between coverage on various compute engines, it looks like spark has the biggest coverage

Not sure how easy that is since they have slightly different syntax

c-thiel:
> We should look into consolidating some of the sql between the tests, i feel like we'll end up having a lot of drift between coverage on various compute engines, it looks like spark has the biggest coverage
> 
> Not sure how easy that is since they have slightly different syntax

Not easy without a ORM which doesn't exist well maintained for spark.
I agree that we should increase coverage for all of them. I'll re-do starrocks once path-style-access is supported.

Maybe follow-up to improve tests?

twuebi (approved):

PR #330, opened 2024-09-09
feat(catalog): expiration queue configuration

### LINKED ISSUE #328: make tabular expiration delay configurable by warehouse

### LINKED ISSUE #329: make drop behavior configurable per warehouse
hard vs soft
drop vs drop purge

twuebi:
closes #329 
closes #328


README.md
```diff
@@ -146,15 +146,15 @@ Following options are global and apply to all warehouses:
...
+| `ICEBERG_REST__LISTEN_PORT`                        | `8080`                                 | Port the server listens on. Default: `8080`                                                                                                                                                                                    |
+| `ICEBERG_REST__SECRET_BACKEND`                     | `postgres`                             | The secret backend to use. If `kv2` is chosen, you need to provide additional parameters found under []() Default: `postgres`, one-of: [`postgres`, `kv2`]                                                                     |
+| `ICEBERG_DEFAULT_TABULAR_EXPIRATION_DELAY_SECONDS` | `86400`                          | Default time after which a tabular, i.e. View or Table which has been dropped is going to be deleted. Can be overwritten on warehouse level. Default: `86400`                                                                  |
```
c-thiel, line 157: Just a nit: maybe put behind the default in square brackets how many days?

crates/iceberg-catalog/migrations/20240909120857_configurable_drops.sql
```diff
@@ -0,0 +1,9 @@
+create type tabular_delete_mode as enum ('soft', 'hard');
+
+alter table warehouse
+    add column tabular_expiration_seconds bigint;
+alter table warehouse
+    add column tabular_delete_mode tabular_delete_mode not null default 'soft';
```
c-thiel, line 6: I am not sure if we should declare a default here. Wouldn't it be better to force us to set this explicitly in the code? Not that what the code does and what we define here diverges because someone forgot to specify?

c-thiel (dismissed):
Just a few nits.

Do we need additional tests?

crates/iceberg-catalog/migrations/20240909120857_configurable_drops.sql
twuebi: mostly for backwards compat, i'll add & remove the default in the same transaction then

PR #331, opened 2024-09-10
chore(ci): use tox-uv

twuebi:
[uv](https://github.com/astral-sh/uv) is much faster than pip and likely stable enough by now


twuebi:
looking at times in ci it doesn't make a difference for us since we're building the app which largely dwarfs anything else

PR #334, opened 2024-09-10
feat(cache): cache metadata location in signer

twuebi:
Adds a cache for get_table_metadata by location, we're only caching the table_id since that's not going to change.

partially addresses #6


crates/iceberg-catalog/src/catalog/s3_signer/sign.rs
```diff
@@ -68,12 +70,12 @@ impl<C: Catalog, A: AuthZHandler, S: SecretStore>
         // We are looking for the path in the database, which allows us to also work with AuthN solutions
         // that do not support custom data in tokens. Perspectively, we should
         // try to get per-table signer.uri support in Spark.
-        let (table_id, table_metadata) = if let Ok(table_id) = require_table_id(table.clone()) {
-            (table_id, None)
+        let table_id = if let Ok(table_id) = require_table_id(table.clone()) {
```
c-thiel, line 73: I am not sure why we where cloning before, but I don't think we have to.

crates/iceberg-catalog/src/config.rs
```diff
@@ -2,6 +2,7 @@
 use anyhow::anyhow;
 use std::collections::HashSet;
 use std::convert::Infallible;
+use std::num::NonZeroUsize;
```
c-thiel, line 5: Fancy!

crates/iceberg-catalog/src/implementations/postgres/tabular/table.rs
```diff
@@ -114,13 +114,16 @@ pub(crate) async fn create_table(
...
-    })?;
+    let location = location
+        .ok_or_else(|| {
+            ErrorModel::builder()
+                .code(StatusCode::CONFLICT.into())
+                .message("Table location is required".to_string())
+                .r#type("CreateTableLocationRequired".to_string())
+                .build()
+        })?
+        .trim_end_matches('/')
```
c-thiel, line 125: This shouldn't be part of the Postgres impl, but fixed in catalog.rs so that its fixed for all backends.

crates/iceberg-catalog/src/implementations/postgres/tabular/table.rs
```diff
@@ -334,79 +337,84 @@ pub(crate) async fn get_table_metadata_by_id(
     })
 }
 
-pub(crate) async fn get_table_metadata_by_s3_location(
+pub(crate) async fn get_table_id_by_s3_location(
     warehouse_id: WarehouseIdent,
     location: &str,
```
c-thiel, line 342: Could we change the type to configs::Location (basically URL?).
I think just by splitting location at :// it is not guaranteed to be a valid URL. Location type is a bit more typesafe and allows easier access to protocol and (sub)-paths by `path_segemnts`.
We might need to expose more attributes of Location.
Maybe having sublocation part of configs.Location makes sense? Then we could share it across impls again.

crates/iceberg-catalog/src/config.rs
twuebi: it's somewhat of a pain to work with.. there's no const constructor which would be infallible that I found

crates/iceberg-catalog/src/implementations/postgres/tabular/table.rs
twuebi: I'm also adding it to determine_tabular_location, it's also here since most tests do not create tables via the catalog but instead directly on postgres impl level

crates/iceberg-catalog/src/implementations/postgres/tabular/table.rs
twuebi: changed to use location instead

c-thiel (approved):

PR #339, opened 2024-09-13
feat(management): users

twuebi:
Running the integration test auto-populates the user table with a service user, it currently UPSERTS a user per call to `config`:

```
postgres=# select * from users;
                  id                  | name |           display_name            | email | origin | deleted_at |          created_at           | updated_at
--------------------------------------+------+-----------------------------------+-------+--------+------------+-------------------------------+------------
 cf7903ea-433f-4316-a8b0-fc9718a3247b |      | service-account-iceberg-rest-test |       | config |            | 2024-09-17 11:59:30.048282+00 |
(1 row)
```


crates/iceberg-catalog/src/service/token_verification.rs
```diff
@@ -59,7 +78,8 @@ pub(crate) async fn auth_middleware_fn(
                 tracing::debug!("Failed to verify token: {:?}", err);
                 return IcebergErrorResponse::from(err).into_response();
             }
-        }
+        };
+        metadata.openid_config = Some(verifier.config.clone());
```
twuebi, line 82: entra id doesn't have a token info endpoint, there's only userinfo but userinfo cannot be called with the provided token, we either need service creds or do some oidc flow which results in a id token

crates/iceberg-catalog/src/api/management/v1/warehouse.rs
```diff
@@ -159,8 +162,36 @@ impl axum::response::IntoResponse for CreateWarehouseResponse {
...
+                "MissingUserId",
+                None,
+            ))?;
+        let user_id = auth.user_id().ok_or(ErrorModel::bad_request(
+            "User ID not found in auth details",
+            "MissingUserId",
+            None,
+        ))?;
+
+        let user_id = Uuid::new_v5(&user_id, auth.issuer().as_bytes());
```
twuebi, line 183: This is future proofing to avoid clashing user_ids across idps. Not sure if necessary, it also obfuscates lookups by user_id in our db. Should probably kick this in favor of a new idp table which we'll add once we go to multiple idps, then the unique constraint could become `(user_id, idp_id)` which would make the problem go away.

crates/iceberg-catalog/migrations/20240913122911_users.sql
```diff
@@ -0,0 +1,14 @@
+create table users
```
twuebi, line 1: one table with a bunch of nullable columns or one for natural persons and another for service users?

crates/iceberg-catalog/src/service/token_verification.rs
```diff
@@ -18,24 +18,107 @@ use serde::de::DeserializeOwned;
...
+    pub fn user_id(&self) -> Option<Uuid> {
+        match self {
+            // TODO: sub is a string, we don't want to deal with that in our db, can we prescribe
+            //       uuid in the jwt?
+            // should we use oid / id? Airflow is doing that:
+            //  - google:   "username": "google_" + data.get("id", ""),
+            //  - azure:    "username": me["oid"],
+            //  - keycloak: "username": data.get("preferred_username", ""),
+            // this seems to conflate the username with the user id though
+            Self::JWT(claims) => claims.sub.parse().ok(),
```
twuebi, line 41: @c-thiel

crates/iceberg-catalog/src/catalog/config.rs
```diff
@@ -142,6 +133,30 @@ impl<
...
+    state: <D as Catalog>::State,
+) -> Result<()> {
+    if let Some(user_id) = request_metadata.user_id() {
+        // If the user is authenticated, create a user in the catalog
+        let user = D::create_user(
+            user_id,
+            request_metadata.user_display_name(),
+            request_metadata.user_name().as_deref(),
+            request_metadata.email(),
+            UserOrigin::ImplicitRegistration("config".to_string()),
```
twuebi, line 147: maybe change to uri of request or remove the inner string instead, would make the db side of things cleaner, currently we're getting a text column instead of an enum there.

crates/iceberg-catalog/src/service/token_verification.rs
c-thiel: I don't think we can prescribe uuids - zitadel for example is not using uuids.

oid is the most correct thing we can use I believe - according to the Azure docs its a global identifier:
https://learn.microsoft.com/en-us/entra/identity-platform/id-token-claims-reference.

We need to make sure though that its also part of OAuth2 tokens and not only the identity token

crates/iceberg-catalog/src/implementations/postgres/user.rs
```diff
@@ -0,0 +1,93 @@
+use super::dbutils::DBErrorHandler;
+use crate::api::management::v1::{User, UserOrigin};
+use crate::service::{ErrorModel, Result};
+use http::StatusCode;
+use sqlx::PgConnection;
+use uuid::Uuid;
+
+#[expect(dead_code)]
+pub(crate) async fn get_user(user_id: Uuid, connection: &mut PgConnection) -> Result<User> {
```
c-thiel, line 9: We need search additionally for the UI - and also a List so that users can be deleted if they no longer want to be searchable

crates/iceberg-catalog/src/service/token_verification.rs
twuebi: [access token docs](https://learn.microsoft.com/en-us/entra/identity-platform/access-token-claims-reference#payload-claims) also say sub is suitable as key in DB:


sub | String | The principal associated with the token. For example, the user of an application. This value is immutable, don't reassign or reuse. The subject is a pairwise identifier that's unique to a particular application ID. If a single user signs into two different applications using two different client IDs, those applications receive two different values for the subject claim. Using the two different values depends on architecture and privacy requirements. See also the oid claim, which does remain the same across applications within a tenant. | This value can be used to perform authorization checks, such as when the token is used to access a resource, and can be used as a key in database tables.
-- | -- | -- | --
oid | String, a GUID | The immutable identifier for the requestor, which is the verified identity of the user or service principal. This ID uniquely identifies the requestor across applications. Two different applications signing in the same user receive the same value in the oid claim. The oid can be used when making queries to Microsoft online services, such as the Microsoft Graph. The Microsoft Graph returns this ID as the id property for a given user account. Because the oidallows multiple applications to correlate principals, to receive this claim for users use the profile scope. If a single user exists in multiple tenants, the user contains a different object ID in each tenant. Even though the user logs into each account with the same credentials, the accounts are different. | This value can be used to perform authorization checks, such as when the token is used to access a resource, and can be used as a key in database tables.






Regarding UUID vs String, I was wondering about the data type of the column in the db, somehow this has to become our primary key, and we'll need to make it a composite of issuer & oid/sub since oid/sub won't be unique across idps, so i understand, we'd be concatenating them and make it a text? For entra-id, this'd be in the 110 bytes range per key if we assume a uuid as oid/sub.

crates/iceberg-catalog/src/api/iceberg/types.rs
```diff
@@ -7,6 +7,11 @@ use serde::{Deserialize, Deserializer, Serialize, Serializer};
 pub struct Prefix(pub(crate) String);
 
 impl Prefix {
+    #[must_use]
+    pub fn new(item: String) -> Self {
```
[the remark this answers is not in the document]
twuebi: tests, can mark it as such

crates/iceberg-catalog/src/api/management/mod.rs
```diff
@@ -70,13 +71,51 @@ pub mod v1 {
...
+    }
+
+    impl IntoResponse for User {
+        fn into_response(self) -> axum::response::Response {
+            (http::StatusCode::CREATED, Json(self)).into_response()
+        }
+    }
+
+    /// Create a new user
+    #[utoipa::path(
```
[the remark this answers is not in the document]
twuebi: do we want to support user deletion vs deactivation? If we support user deletion, we'll have to be careful not to use user_ids as foreign keys in other tables that should outlast the user, e.g. some change log.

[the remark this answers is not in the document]
twuebi: I'd also like to add those as follow ups to not blow up scope too much

crates/iceberg-catalog/src/api/management/mod.rs
```diff
@@ -70,13 +71,51 @@ pub mod v1 {
...
     }
 
+    #[derive(Debug, Serialize, utoipa::ToSchema)]
+    pub enum UserOrigin {
+        ImplicitRegistration(String),
+        ExplicitRegistration,
+    }
+
+    #[derive(Debug, Serialize, utoipa::ToSchema)]
+    pub struct User {
```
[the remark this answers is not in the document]
twuebi: Can we know? We could use implicit / explicit registration as a proxy but I could also call /config as a first action, right?

crates/iceberg-catalog/migrations/20240913122911_users.sql
twuebi: iirc, name is optional in tokens, should we reject tokens without name?

crates/iceberg-catalog/src/service/token_verification.rs
c-thiel: Decision: oid

c-thiel: Decision: Concatenation with <url-encoded issuer url>/<oid> (Validate that oid contains of alphanumeric chars and -)

crates/iceberg-catalog/src/api/iceberg/types.rs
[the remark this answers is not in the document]
twuebi: removed, cfg(test) items don't show up in integration tests, i.e. a /tests directory next to /src, will just add / remove for my tests

c-thiel:
Closed in favor of https://github.com/hansetag/iceberg-catalog/pull/381

Issue #271, opened 2024-08-15
Support PubSub for Kafka


c-thiel:

brgr-s:
I would like to give this one a shot

brgr-s:
Turns out there is something to discuss right from the get go.

I kind of expected that there would be one clear choice for a kafka rust lib, but at least at first glance, there is not.

There seem to be two more or less mature implementations available:

- [kafka-rust](https://github.com/kafka-rust/kafka-rust)
- [rust-rdkafka](https://github.com/fede1024/rust-rdkafka)

### kafka-rust

Seems to be a pure rust implementation. A first look at the examples shows that it seems to be pretty easy to use. There a few things to mention, though

- Last release is [0.10.0](https://crates.io/crates/kafka/0.10.0), which is 11 months old. It has no tag in the GitHub repo (last tag is 0.9.0 from may 2022). Repo seems to be active, though
- They do not regard their lib as production ready: ["Use it in production at your own risk."](https://crates.io/crates/kafka/0.10.0)
- There seems to be no async client/producer implemented

### rust-rdkafka

... is actually "just" a safe interface to [librdkafka](https://github.com/confluentinc/librdkafka)

- Last release is [0.36.2](https://github.com/fede1024/rust-rdkafka/releases/tag/v0.36.2), which is 7 months old< librdkafka of course is actively maintained, the last release is 2 months old.
- async producer available
- They do not mention if the lib is production ready, however since it is based on librdkafka, one might assume that it should be
- rust-rdkafka can be enabled for Cloud Events SDK via [feature flat](https://github.com/cloudevents/sdk-rust?tab=readme-ov-file#crate-structure). I am not sure if this is relevant, as the same is true for nats, however that feature flag is not used in TIP
- Reliance on a c lib has of course caveats.
  - librdkafka can be linked statically when the lib is build, using cmake (which is required on the build system); if SSL is required, libssl-dev is required.... [and so on](https://github.com/cloudevents/sdk-rust?tab=readme-ov-file#crate-structure)
  - librdkafka could be linked dynamically, introducing a dependency on the system that is running TIP

### what next?

I am not sure what is the best choice here. If introducing a c lib is not an option, rust-kafke seems to be the only choice. If it is ok to schlepp around a c lib, rust-rdkafka is also async and seems to be "endorsed" by Cloud Events SDK.

Or maybe I am overlooking "that other kafka rust lib", that has less downsides than kafka-rust or rust-rdkafke :smile:

[...]

PR #340, opened 2024-09-13
feat: add kafka support #271

Grongrilla:


brgr-s:
This is a WIP PR

### tested

Locally running kafka (docker) recieves the same CloudEvents messages as the NATS events backend on the configured topic. The configuration of the initial brokers with a single host:port entry works.

### open or tbd

- [x] ~~Final decision to use `FutureProvider` or `BasicProvider`; remove unused provider to fix clippy errors~~ done
- [x] ~~No authentication supported~~ default allows SASL and SASL-SSL (untested: OATUHBEARER)
- [x] ~~No fine grained control over producer configuration~~ reads configuration from env variable map, which is passed to rdkafka
- [x] ~~README.MD not yet upated with new configuration environment variables~~ done
- [x] ~~Configuration of multiple initial brokers not tested~~ tested with confluence free plan
- [x] ~~No SSL support~~ enabled by default with vendored openssl from rdkafka
- [x] ~~Decide if current location for vendored code is fine, or move to better location (maybe kafka module below service module~~ restructured event_provider in submodules. vendored code lifes in kafka submodule
- [x] Allow reading (librd)kafka config from a file

Cargo.toml
```diff
@@ -1,8 +1,8 @@
 [workspace]
 members = [
```
twuebi, line 2: Could we get rid of the formatting changes here? It makes it a bit more difficult to discern actual changes from whitespace changes & we'll likely end up with a bit of ping pong with formatting between PRs. We could consider adopting https://github.com/tamasfe/taplo to format tomls.

crates/iceberg-catalog-bin/src/serve.rs
```diff
@@ -69,11 +70,21 @@ pub(crate) async fn serve(bind_addr: std::net::SocketAddr) -> Result<(), anyhow:
...
+        running_with_publisher = true;
+    }
+    if let Some(kafka_brokers) = &CONFIG.kafka_brokers {
+        let kafka_publisher = build_kafka_producer(kafka_brokers)?;
+        cloud_event_sinks
+            .push(Arc::new(kafka_publisher) as Arc<dyn CloudEventBackend + Sync + Send>);
+        running_with_publisher = true;
+    }
+
+    if !running_with_publisher {
```
twuebi, line 87: This could become `if cloud_event_sinks.is_empty()`

crates/iceberg-catalog-bin/src/serve.rs
```diff
@@ -160,3 +171,23 @@ async fn build_nats_client(nat_addr: &Url) -> Result<NatsBackend, Error> {
...
+        .fold(String::new(), |acc, e| format!("{acc},{e}"));
+    let kafka_backend = KafkaBackend {
+        producer: rdkafka::ClientConfig::new()
+            .set("bootstrap.servers", &kafka_brokers_csv)
+            .create()?,
+        topic: CONFIG
+            .kafka_topic
+            .clone()
+            .ok_or(anyhow::anyhow!("Missing kafka topic."))?,
+    };
```
twuebi, line 187: Should we accept a `Vec<(SocketAddr, Topic)>`? Are the kafka brokers instances of the same deployment or is it to support separate brokers? If it's separate brokers, then I'd think we'd want flexibility to specify a topic for each of those

crates/iceberg-catalog-bin/src/serve.rs
```diff
@@ -160,3 +171,23 @@ async fn build_nats_client(nat_addr: &Url) -> Result<NatsBackend, Error> {
     };
     Ok(nats_publisher)
 }
+
+fn build_kafka_producer(kafka_brokers: &Vec<SocketAddr>) -> Result<KafkaBackend, Error> {
+    let kafka_brokers_csv = kafka_brokers
+        .iter()
+        .fold(String::new(), |acc, e| format!("{acc},{e}"));
```
twuebi, line 178: Doesn't this prepend a comma to the string?

Initially, `acc` is empty, so the first iteration would yield: `,<some-address>`

I think we already depend on itertools, so you could just use https://docs.rs/itertools/latest/itertools/fn.join.html instead

crates/iceberg-catalog/src/vendor/cloudevents/binding/mod.rs
```diff
@@ -0,0 +1,40 @@
+// Vendored from
```
twuebi, line 1: I think we need to include the license header here

crates/iceberg-catalog/src/vendor/cloudevents/binding/mod.rs
```diff
@@ -0,0 +1,40 @@
...
+
+fn header_prefix(prefix: &str, name: &str) -> String {
+    if name == "datacontenttype" {
+        CONTENT_TYPE.to_string()
+    } else {
+        [prefix, name].concat()
+    }
+}
+
+#[macro_export]
```
twuebi, line 31: Where's this macro being used? If only within this module, you can get around the macro_export by putting a `use header_value_to_str;` below the definition.

brgr-s:
@twuebi are you Ok with the way I introduced the vendored code or would you prefer something different?

twuebi:
> @twuebi are you Ok with the way I introduced the vendored code or would you prefer something different?

Maybe we promote `service::event_publishers` to a directory module and then have submodules for nats & kafka where kafka has a submodule `vendored`? IMO the vendored stuff is a detail of the kafka module and we could just stuff it in there, what do you think?

brgr-s:
> > @twuebi are you Ok with the way I introduced the vendored code or would you prefer something different?
> 
> Maybe we promote `service::event_publishers` to a directory module and then have submodules for nats & kafka where kafka has a submodule `vendored`? IMO the vendored stuff is a detail of the kafka module and we could just stuff it in there, what do you think?

The modules would be very small, especially after vendored is gone (which will happen sometimes, I hope). I still like this idea better than putting the vendored code from Cloudevents SDK in its' own module. I will restructure the code as you suggested :)

brgr-s:
Regarding the discussion about brokers and topics, I found this helpfull: [Kafka broker](https://www.redpanda.com/guides/kafka-architecture-kafka-broker) and [rdkafka Configuration](https://github.com/confluentinc/librdkafka/blob/master/CONFIGURATION.md).

So to my understanding, a Kafka deployment can have multiple brokers, but they all know the same topics. The configuration parameter of rdkafka seems to expect the list of brokers to belong to the same deployment - it is not explicetly stated as such, but there is no way to configure a topic->broker mapping in rdkafka. The topic is part of the message, not the producer instance.

It could be possible that a TIP users wants to publish to multiple topics, which might life on multiple kafka deployments. The same is probably true for NATS.... but this sounds like an exception to me that we probably do not need to support out of the box.

But on taht topic, another question: do we want to test kafka support with "just" kafka, or also with redpanda?

twuebi:
> But on taht topic, another question: do we want to test kafka support with "just" kafka, or also with redpanda?

is one much more work than the other? In general, testing against more implementations is nice, on the other hand, I'd defer it to a follow-up if it makes the scope of this PR bigger.

brgr-s:
Commit #2729dd9 changed to configuration to use a HashMap and use the key-value-pairs as-is to set properties on `rdkafka FutureProducer`. Error handling is defered to rdkafka apart from checking for the existance of one minimally required key (but not checking the value). librdkafka has a lot of [configration parameters](https://github.com/confluentinc/librdkafka/blob/master/CONFIGURATION.md) and it seems to be unreasonable to make a sensible selection.

brgr-s:
@twuebi @c-thiel 

If you have some time to take a look at the current WIP state of this PR, I'd be gratefull.

Baiscally all configuration parameters for librdkafka are just taken as String->String map. Only sensitive parameters are extracted (the ones I found, I might have overlooked something).

I also extracted "enable.idempotence" and set it to "true" as default, as per your suggesion @c-thiel 

I am debating on setting "client.id" to "Lakekeeper" as default. The deafult is "rdkafka" if nothing is set.

Also there is the question of setting a "key" when sending cloudevents. Currently it is a configuration parameter. We could set it to a value depending on the message, f. ex. warehouse-id or warehouse-id-catalog, or something like that.

I opted agains extracing a whole bunch of parameters and introducing types (f. ex. enums) or using the proper type of that parameter (f. ex. integer). This could be usefull if we did any sort of validation or if we wanted to restrict some values, but on the other hand, all parameters are evaluated by librdkafka and rust-rdkafka wrapes the resulting error code into an `Error`. I feel like this is fine for the beginning.

Some parameter values will lead to a required dependency on rust-rdkafka (certain SASL options need libsaslv2, there is libzstd for compression). We could try if we can actually support a certain configuration property with the current feature selection... but again, it seem slike overkill.

crates/iceberg-catalog-bin/src/serve.rs
```diff
@@ -160,3 +167,59 @@ async fn build_nats_client(nat_addr: &Url) -> Result<NatsBackend, Error> {
...
+    topic: &String,
+    key: &Option<String>,
+) -> Result<KafkaBackend, Error> {
+    if !(kafka_config.conf.contains_key("bootstrap.servers")
+        || kafka_config.conf.contains_key("metadata.broker.list"))
+    {
+        return Err(anyhow!(
+            "Kafka config map does not conain 'bootstrap.servers' or 'metadata.broker.list'. You need to provide either of those, in addition with any other parameters you need."
+        ));
+    }
```
twuebi, line 182: if these are mandatory configs, we should probably make them well-known values?

crates/iceberg-catalog-bin/src/serve.rs
```diff
@@ -160,3 +167,59 @@ async fn build_nats_client(nat_addr: &Url) -> Result<NatsBackend, Error> {
...
+fn build_kafka_producer(
+    kafka_config: &KafkaConfig,
+    topic: &String,
+    key: &Option<String>,
+) -> Result<KafkaBackend, Error> {
+    if !(kafka_config.conf.contains_key("bootstrap.servers")
+        || kafka_config.conf.contains_key("metadata.broker.list"))
+    {
+        return Err(anyhow!(
+            "Kafka config map does not conain 'bootstrap.servers' or 'metadata.broker.list'. You need to provide either of those, in addition with any other parameters you need."
```
twuebi, line 180: s/conain/contain

brgr-s:
I feel this is pretty close to beeing done. Comments and suggestions are very welcome.

One test related to openfga is failing. I don't think it has anything to do with my code. I it does I am happy to fix it.

brgr-s:
@c-thiel With commit 03b6d38 "enable.idempotence" is no longer set to "true". This requires other parameters to be set to the correct value, depending on how kafka is configured. I felt it was a bit intrusive.

crates/iceberg-catalog/src/config.rs
```diff
@@ -35,9 +36,14 @@ fn get_config() -> DynAppConfig {
...
     let prefixes = &["LAKEKEEPER_TEST__"];
 
+    let file_keys = &["kafka_config"];
+
     let mut config = figment::Figment::from(defaults);
     for prefix in prefixes {
-        config = config.merge(figment::providers::Env::prefixed(prefix).split("__"));
+        let env = figment::providers::Env::prefixed(prefix).split("__");
+        config = config
+            .merge(figment_file_provider_adapter::FileAdapter::wrap(env.clone()).only(file_keys))
```
twuebi, line 45: this adds another layer of configuration where it looks for `kafka_config` in current dir?

crates/iceberg-catalog/src/service/event_publisher/kafka/mod.rs
```diff
@@ -0,0 +1,79 @@
+pub(crate) mod vendor;
+
+use std::collections::HashMap;
+
+#[cfg(feature = "kafka")]
```
twuebi, line 5: you can just put `#[cfg(feature = "kafka")]` on the mod `kafka` in event_publisher instead

brgr-s:
@twuebi @c-thiel problem with kafka in test container is resolved

What remains open is the discussion about using a key for kafka backed cloudevents. Should we extract one from the message, for example warehouse, namespace, table-name, a combination?

twuebi:
> @twuebi @c-thiel problem with kafka in test container is resolved

Awesome!

> What remains open is the discussion about using a key for kafka backed cloudevents. Should we extract one from the message, for example warehouse, namespace, table-name, a combination?

tabular_id sounds sensible, what do you think @c-thiel?

twuebi:
> > @twuebi @c-thiel problem with kafka in test container is resolved
> 
> Awesome!
> 
> > What remains open is the discussion about using a key for kafka backed cloudevents. Should we extract one from the message, for example warehouse, namespace, table-name, a combination?
> 
> tabular_id sounds sensible, what do you think @c-thiel?

@Grongrilla discussed this with @c-thiel, we think `tabular_id` is the way

twuebi (dismissed):

twuebi (dismissed):

twuebi (approved):

Issue #234, opened 2024-08-04
Support GCS


c-thiel:

corleyma:
FYI: I added an issue for this here #204.  Feel free to close mine out in favor of this one!

c-thiel:
Closed in favor of https://github.com/hansetag/iceberg-catalog/issues/204

PR #361, opened 2024-09-20
feat(storage): support gcs

twuebi:
Still figuring out how token downscoping is working

https://cloud.google.com/storage/docs/access-control/signed-urls
https://cloud.google.com/storage/docs/uniform-bucket-level-access
https://cloud.google.com/apis/design/resource_names
https://cloud.google.com/iam/docs/reference/sts/rest/v1/Options

https://cloud.google.com/iam/docs/downscoping-short-lived-credentials#exchange-credential

https://github.com/salrashid123/downscoped_token

#234


crates/iceberg-catalog/src/service/storage/gcs/mod.rs
```diff
@@ -0,0 +1,373 @@
...
+            }
+        }
+
+        Ok(config)
+    }
+
+    fn normalize_key_prefix(&mut self) -> Result<(), ValidationError> {
+        if let Some(key_prefix) = self.key_prefix.as_mut() {
+            *key_prefix = key_prefix.trim_matches('/').to_string();
+            if key_prefix.starts_with(".well-known/acme-challenge/") {
```
c-thiel, line 185: out of curiosity, why is this a problem?

crates/iceberg-catalog/src/service/storage/gcs/mod.rs
```diff
@@ -0,0 +1,373 @@
...
+                self.key_prefix = None;
+            }
+        }
+
+        // GCS supports a max of 1024 chars and we need some buffer for tables.
+        if let Some(key_prefix) = self.key_prefix.as_ref() {
+            if key_prefix.len() > 896 {
+                return Err(ValidationError::InvalidProfile {
+                    source: None,
+                    reason: "Storage Profile `key_prefix` must be less than 1024 characters."
```
c-thiel, line 205: Less than 896 chars?

crates/iceberg-catalog/src/service/storage/error.rs
```diff
@@ -45,6 +45,10 @@ impl From<TableConfigError> for ValidationError {
                     entity: "TableConfig".to_string(),
                 }
             }
+            TableConfigError::StsError(_, _) => ValidationError::Internal {
```
c-thiel, line 48: I think this should not be internal, it happens on validation of a Warehouse. Its typically a misconfiguration on client side.

crates/iceberg-catalog/src/service/storage/error.rs
```diff
@@ -150,6 +159,9 @@ impl From<TableConfigError> for IcebergErrorResponse {
             e @ TableConfigError::Misconfiguration(_) => {
                 ErrorModel::bad_request(e.to_string(), "Misconfiguration", Some(Box::new(e))).into()
             }
+            e @ TableConfigError::StsError(_, _) => {
+                ErrorModel::internal(e.to_string(), "StsError", Some(Box::new(e))).into()
```
c-thiel, line 163: Also here - bad_request?

crates/iceberg-catalog/src/service/storage/gcs/sts.rs
```diff
@@ -0,0 +1,226 @@
...
+            TableConfigError::StsError("Downscoping failed".to_string(), Some(Box::new(e)))
+        })
+}
+
+#[derive(Deserialize, veil::Redact)]
+pub(crate) struct STSResponse {
+    #[redact(partial)]
+    pub(crate) access_token: String,
+    pub(crate) expires_in: Option<usize>,
+    token_type: String,
```
c-thiel, line 74: Should we introduce an enum, which we could also re-use in the request?

crates/iceberg-catalog/src/service/storage/gcs/sts.rs
```diff
@@ -0,0 +1,226 @@
...
+            // A JSON-format Credential Access Boundary, encoded with percent encoding.
+            // idk?
+            options: percent_encoding::utf8_percent_encode(&op, percent_encoding::NON_ALPHANUMERIC)
+                .to_string(),
+        })
+    }
+}
+
+#[derive(Serialize, Deserialize)]
+struct Options {
```
c-thiel, line 122: Shouldn't we also add an audience to scope it to the storage service?
https://cloud.google.com/iam/docs/reference/sts/rest/v1beta/Options

crates/iceberg-catalog/src/service/storage/gcs/sts.rs
```diff
@@ -0,0 +1,226 @@
...
+            token_info_url: None,
+            service_account_impersonation_url: None,
+            service_account_impersonation: None,
+            delegates: None,
+            credential_source: None,
+            quota_project_id: None,
+            workforce_pool_user_project: None,
+        }
+    }
+}
```
c-thiel, line 222: I think we should add a test here that tests the negative case - i.e. that a file outside of the downscoped prefix cannot be accessed. The validate method only tests the positive case I think.

crates/iceberg-catalog/src/service/storage/gcs/mod.rs
twuebi: https://cloud.google.com/storage/docs/objects#naming

> Object names cannot start with .well-known/acme-challenge/.

There are also a few others, not sure if we should be checking them:

- Object names can contain any sequence of valid Unicode characters, of length 1-1024 bytes when UTF-8 encoded.
- Object names cannot contain [Carriage Return or Line Feed characters](https://en.wikipedia.org/wiki/Newline#Unicode).
- Objects cannot be named . or ...

crates/iceberg-catalog/src/service/storage/gcs/mod.rs
twuebi: copy paste error from s3, also fixing it over there 😄

crates/iceberg-catalog/src/service/storage/error.rs
twuebi: it's 50/50, I'll split the error type then

crates/iceberg-catalog/src/service/storage/error.rs
twuebi: When getting table config it's more on the internal error, maybe failed dependency, there's nothing wrong with the GET table request we received, we either have a bad state internally or some external cloud provider config has changed

crates/iceberg-catalog/src/service/storage/gcs/sts.rs
twuebi: we don't really use the token_type, we can remove it

crates/iceberg-catalog/src/service/storage/gcs/sts.rs
twuebi: Not sure what's going on there but the endpoint rejected that:

```
downscoping response: "{\"error\":\"invalid_request\",\"error_description\":\"Unsupported argument: `options.audiences`\"}"
```

twuebi: even google's own python implementation is not exposing that parameter, https://github.com/googleapis/google-auth-library-python/blob/484c8db151690a4ae7b6b0ae38db0a8ede88df69/google/auth/downscoped.py#L170

crates/iceberg-catalog/src/service/storage/gcs/sts.rs
twuebi: added tests for vended creds in `service/storage/mod.rs`

crates/iceberg-catalog/src/service/storage/gcs/sts.rs
twuebi: i think we should keep it, seems useful for debugging, i think it's not the same enum as in the request, in the request only a single string is valid anyways

crates/iceberg-catalog/src/service/storage/gcs/mod.rs
```diff
@@ -0,0 +1,373 @@
...
+    pub bucket: String,
+    /// Subpath in the bucket to use.
+    /// The same prefix can be used for multiple warehouses.
+    pub key_prefix: Option<String>,
+}
+
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize, utoipa::ToSchema)]
+#[serde(tag = "credential-type", rename_all = "kebab-case")]
+#[schema(rename_all = "kebab-case")]
+pub enum GcsCredential {
```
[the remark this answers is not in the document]
twuebi: there's one in storage.md, I'll add an example here too, it's the json you get when creating a gcp service account

crates/iceberg-catalog/src/service/storage/gcs/mod.rs
c-thiel: Hm, should be fine. Lets just accept an error in those cases

crates/iceberg-catalog/src/service/storage/gcs/sts.rs
c-thiel: Well, then let's forget it. Strange

c-thiel (approved):

PR #381, opened 2024-10-01
authz + users

twuebi:


crates/iceberg-catalog/src/api/management/mod.rs
```diff
@@ -392,6 +598,9 @@ pub mod v1 {
                     "/warehouse/:warehouse_id/deleted_tabulars",
                     get(list_deleted_tabulars),
                 )
+                .route("/user", post(register_user).get(list_users))
+                .route("user/:issuer/:id", delete(delete_user).put(update_user))
```
twuebi, line 602: TODO: remove issuer from id, just use id

crates/iceberg-catalog/migrations/20240913122911_users.sql
```diff
@@ -0,0 +1,24 @@
+create type user_origin as enum ('explicit-via-register-call', 'implicit-via-config-call');
+
+create table users
```
c-thiel, line 3: I think we need an additional column project_id for the search endpoint.
We need to store this information from the start, or we won't be able to add the project-scoped search endpoint later.
Same for roles I believe.

Maybe we should split project into a separate table first.

c-thiel: A user entry can exist multiple times for multiple projects. Not sure if we should add another mapping table instead? It would allow us to update a single user id in all projects simultaneously. Otherwise, if we ever do support multiple IdPs, having it separated per project in the first place would make sense.

crates/iceberg-catalog/src/implementations/postgres/user.rs
```diff
@@ -0,0 +1,270 @@
...
+        users,
+        next_page_token: None,
+    })
+}
+
+pub(crate) async fn delete_user<'c, 'e: 'c, E: sqlx::Executor<'c, Database = sqlx::Postgres>>(
+    id: UserId,
+    connection: E,
+) -> Result<()> {
+    // TODO: @christian soft delete or hard delete? I'm leaning towards soft delete to avoid
```
c-thiel, line 90: I am not sure - we must have a mechanism to delete users completely because of DSGVO.
So we need to find a way to implement it anyway. I tend towards hard delete or remove all PII data additionally - i.e. removing name and such.

Dangling references in AuthZ wouldn't be much of a problem.

crates/iceberg-catalog/src/implementations/postgres/user.rs
```diff
@@ -0,0 +1,270 @@
...
+        r#"
+        SELECT
+            id,
+            name,
+            origin as "origin: DbUserOrigin",
+            email,
+            created_at,
+            updated_at
+        FROM users
+        where (deleted_at is null OR $1) AND (name ILIKE $2)
```
c-thiel, line 68: I don't think using ILIKe on user provided input is a good idea.
We should at least limit search string length and make sure it doesn't contain (or escape) "%" (and other reserved chars?).
We should also add %<search-term>% or something like this by default.
If I am looking for a part of a last name I would still expect to find the person even if I don't specify a full regex.

Maybe there is a certain index in Postgres for free text search?

crates/iceberg-catalog/src/api/management/mod.rs
```diff
@@ -71,12 +73,216 @@ pub mod v1 {
...
+        (
+            http::StatusCode::OK,
+            Json(ApiServer::<C, A, S>::delete_user(api_context, metadata, id).await),
+        )
+            .into_response()
+    }
+
+    #[derive(Debug, Deserialize, utoipa::ToSchema)]
+    pub struct CreateRoleRequest {
+        pub id: String,
```
c-thiel, line 220: This should be treated as the external_id and it should be optional.

PR #405, opened 2024-10-14
Cors & UI and workflows

twuebi:


crates/iceberg-catalog/src/config.rs
```diff
@@ -80,6 +81,12 @@ pub struct DynAppConfig {
     /// reverse proxy before routing to the catalog service.
     /// Example value: `{warehouse_id}`
     prefix_template: String,
+    /// CORS allowed origins. If not set, CORS is disabled.
+    #[serde(
+        deserialize_with = "deserialize_origin",
+        serialize_with = "serialize_origin"
+    )]
+    pub allow_origin: Option<Vec<HeaderValue>>,
```
c-thiel, line 89: As there is more than one, should we name it allowed_origins?

c-thiel: Do we ever need serialization of the config? can we get rid of those functions?

Regarding name - maybe allow_origin is OK as this is also how tower calls it.

examples/self-contained/docker-compose-build.yaml
```diff
@@ -7,6 +7,8 @@ services:
     build:
       context: ../../
       dockerfile: docker/full.Dockerfile
+    environment:
+      - ICEBERG_REST__ALLOW_ORIGIN=http://localhost:3002
```
c-thiel, line 11: Why do we need it only in -build.yaml? Shouldn't it be part of the main docker-compose as well?

Ah - because UI is only here?

c-thiel (approved):

crates/iceberg-catalog/src/config.rs
twuebi: Iirc figment needs it

examples/self-contained/docker-compose-build.yaml
twuebi: Yep

PR #450, opened 2024-10-30
feat: pagination without empty pages

twuebi:
closes https://github.com/lakekeeper/lakekeeper/issues/457


crates/iceberg-catalog/src/api/iceberg/mod.rs
```diff
@@ -67,46 +68,135 @@ pub mod v1 {
...
+            self.ordering
+                .into_iter()
+                .zip(self.next_page_tokens)
+                // we can unwrap here since the only way of adding items is via insert which ensures that every
+                // entry in self.ordering is also a key into self.tabulars.
+                .map(move |(key, next_p)| {
+                    let v = self
+                        .entities
+                        .remove(&key)
+                        .expect("keys have to be in tabulars if they are in self.ordering");
```
c-thiel, line 162: tabulars -> entities

crates/iceberg-catalog/src/api/iceberg/mod.rs
```diff
@@ -67,46 +68,135 @@ pub mod v1 {
...
+            self.next_page_tokens.remove(idx);
+            self.entities.remove(key)
+        }
+
+        #[allow(clippy::missing_panics_doc)]
+        pub fn into_iter_with_page_tokens(mut self) -> impl Iterator<Item = (T, V, String)> {
+            self.ordering
+                .into_iter()
+                .zip(self.next_page_tokens)
+                // we can unwrap here since the only way of adding items is via insert which ensures that every
```
c-thiel, line 156: I think this can lead to panics currently if the insert "silent override" hits.
In this case we still push ordering & next_page_token, which can contain duplicates as they are Vecs.
While `.remove` the first item works, the second panics

crates/iceberg-catalog/src/api/iceberg/mod.rs
```diff
@@ -67,46 +68,135 @@ pub mod v1 {
...
+        pub(crate) fn with_capacity(capacity: usize) -> Self {
+            Self {
+                entities: HashMap::with_capacity(capacity),
+                next_page_tokens: Vec::with_capacity(capacity),
+                ordering: Vec::with_capacity(capacity),
+            }
+        }
+
+        pub(crate) fn insert(&mut self, key: T, value: V, next_page_token: String) {
+            // TODO: should this become a result instead of a silent overwrite?
```
c-thiel, line 113: I believe this can currently lead to panics, see my comment below.

crates/iceberg-catalog/src/api/iceberg/mod.rs
twuebi: The vecs are guaranteed to have unique items:

We start with an empty collection, the only way to add items is via insert.

1. insert item A: 1 item in hashmap, ordering and next page token
2. insert item B: 2 items in hashmap, ordering and next page token
3. insert item B: we insert into hashmap, hashmap returns a V, we find the index of K in both vecs and remove them before pushing again

since there is no other way of adding items, we know that the vecs have a uniqueness guarantee.

I've added some tests, could you check if your case is covered?

c-thiel (approved):

PR #477, opened 2024-11-04
feat: Extend user search to email field

c-thiel:


crates/iceberg-catalog/migrations/20241009122911_users_and_roles.sql
```diff
@@ -17,7 +17,7 @@ create table users
 call add_time_columns('users');
 select trigger_updated_at('users');
 
-CREATE INDEX users_name_gist_idx ON users USING gist (name gist_trgm_ops(siglen=256));
+CREATE INDEX users_name_email_gist_idx ON users USING gist ( (name || ' ' || email) gist_trgm_ops(siglen=256));
```
twuebi, line 20: we shouldn't edit migrations in-place, it'll break sqlx migrate due to changed SHA for existing deployments

crates/iceberg-catalog/src/implementations/postgres/user.rs
```diff
@@ -212,7 +212,7 @@ pub(crate) async fn search_user<'e, 'c: 'e, E: sqlx::Executor<'c, Database = sql
 ) -> Result<SearchUserResponse> {
     let users = sqlx::query!(
         r#"
-        SELECT id, name, name <-> $1 AS dist, user_type as "user_type: DbUserType"
+        SELECT id, name, email, (name || ' ' || email) <-> $1 AS dist, user_type as "user_type: DbUserType"
```
twuebi, line 215: why search in the concatenation?

PR #507, opened 2024-11-16
feat: Default to hard deletion

c-thiel:
New warehouses are by default created with hard deletion enabled.
This avoids some unexpected behaviors, i.e. that namespaces cannot be deleted if a soft-deleted tabular is still present, unless explicitly enabled.


crates/iceberg-catalog/src/api/management/v1/warehouse.rs
```diff
@@ -79,10 +79,12 @@ pub enum TabularDeleteProfile {
     #[schema(title = "TabularDeleteProfileHard")]
     Hard {},
     #[schema(title = "TabularDeleteProfileSoft")]
+    #[serde(rename_all = "kebab-case")]
     Soft {
         #[serde(
             deserialize_with = "seconds_to_duration",
-            serialize_with = "duration_to_seconds"
+            serialize_with = "duration_to_seconds",
+            alias = "expiration_seconds"
```
twuebi, line 87: why the alias?

twuebi (approved):

PR #517, opened 2024-11-17
feat: undrop 

twuebi:


crates/iceberg-catalog/src/api/management/v1/warehouse.rs
```diff
@@ -217,6 +218,20 @@ impl axum::response::IntoResponse for CreateWarehouseResponse {
     }
 }
 
+#[derive(Deserialize, Debug, ToSchema)]
+#[serde(rename_all = "kebab-case", untagged)]
+pub enum Undroppable {
```
twuebi, line 223: Come up with better name

crates/iceberg-catalog/src/api/management/v1/warehouse.rs
```diff
@@ -641,6 +656,108 @@ pub trait Service<C: Catalog, A: Authorizer, S: SecretStore> {
...
+    ) -> Result<()> {
+        // ------------------- AuthZ -------------------
+        let catalog = context.v1_state.catalog;
+        let authorizer = context.v1_state.authz;
+        // TODO: authz
+        let mut futs = vec![];
+        match &request.undrop {
+            Undroppable::Tabulars { tabulars: tabs } => {
+                for t in tabs {
+                    match t {
```
twuebi, line 673: move to fn

crates/iceberg-catalog/src/api/management/v1/warehouse.rs
```diff
@@ -641,6 +656,108 @@ pub trait Service<C: Catalog, A: Authorizer, S: SecretStore> {
...
+                        None,
+                    )
+                    .into());
+                };
+            }
+        }
+        if !futures::future::try_join_all(futs)
+            .await?
+            .into_iter()
+            .all(|t| t)
```
twuebi, line 715: in namespace case, we end up with empty futs, all on an empty iter evaluates to true

crates/iceberg-catalog/src/api/management/v1/warehouse.rs
```diff
@@ -641,6 +656,108 @@ pub trait Service<C: Catalog, A: Authorizer, S: SecretStore> {
...
+                        page_size: Some(1000),
+                    },
+                )
+                .await?
+                .into_iter()
+                .map(|t| TableIdentUuid::from(*t.0))
+                .collect_vec()
+            }
+        };
+        let task_id = C::undrop_tabular(&tabs, transaction.transaction()).await?;
```
twuebi, line 750: undrop_tabulars

crates/iceberg-catalog/src/api/management/v1/warehouse.rs
```diff
@@ -641,6 +656,108 @@ pub trait Service<C: Catalog, A: Authorizer, S: SecretStore> {
...
+        context
+            .v1_state
+            .queues
+            .cancel_tabular_expiration(TaskFilter::TaskIds(task_id))
+            .await?;
```
twuebi, line 755: emit event? @c-thiel

twuebi: #572

crates/iceberg-catalog/src/service/authz/implementations/openfga/migration/mod.rs
```diff
@@ -50,58 +51,100 @@ pub(crate) async fn migrate(
         return Ok(());
     }
 
+    let written_active_model = write_active_model(client, &store).await?;
```
twuebi, line 54: move into loop

crates/iceberg-catalog/src/service/authz/implementations/openfga/migration/mod.rs
```diff
@@ -50,58 +51,100 @@ pub(crate) async fn migrate(
...
-            ModelVersion::V1 => Ok(()),
+        for int_id in max_applied.as_monotonic_int()..written_active_model.version_as_int() {
+            let model_version = ModelVersion::from_monotonic_int(int_id)
+                .ok_or(OpenFGAError::UnknownModelVersionApplied(int_id))?;
+            tracing::info!("Applying migration for model version {}", model_version);
+            match model_version {
+                ModelVersion::V1 => {
+                    // no migration to be done, we start at v1
+                }
+                ModelVersion::V2 => v2::migrate(client, &store).await,
```
twuebi, line 69: mark as applied

crates/iceberg-catalog/src/implementations/postgres/task_queues/mod.rs
```diff
@@ -66,26 +66,30 @@ async fn queue_task(
...
+                idempotency_key,
+                warehouse_id,
+                suspend_until)
         VALUES ($1, $2, 'pending', $3, $4, $5, $6)
-        ON CONFLICT ON CONSTRAINT unique_idempotency_key DO NOTHING
+        ON CONFLICT ON CONSTRAINT unique_idempotency_key
+        DO UPDATE SET
+            status = EXCLUDED.status,
+            suspend_until = EXCLUDED.suspend_until
+        WHERE task.status = 'cancelled'
```
twuebi, line 81: This is basically resume task, we could also use the same for retries. 

Alternatively, we do not update cancelled tasks but instead we create a new task if a previously restored table is to be dropped again.

twuebi: @c-thiel

crates/iceberg-catalog/src/api/management/v1/warehouse/undrop.rs
```diff
@@ -0,0 +1,120 @@
...
+    warehouse_ident: WarehouseIdent,
+    request: UndropTabularsRequest,
+    transaction: <C::Transaction as Transaction<C::State>>::Transaction<'_>,
+) -> api::Result<Vec<TableIdentUuid>> {
+    Ok(match request.target {
+        UndropTarget::Tabulars { tabulars: tabs } => {
+            tabs.into_iter().map(|i| TableIdentUuid::from(*i)).collect()
+        }
+        UndropTarget::Namespace { namespace: ns } =>
+        // TODO: do we want to paginate here?
```
c-thiel, line 21: I think it would be the most correct way to paginate. Alternatively for now we could also just document that this undroppes the first 1000 tables so the namespace and the client should retry.

I tend towards implementing pagination now as we will need pagination at least for "drop cascade" later, and it would be strange if one API paginates and the other doesn't. On the other hand this is a sync API and drop might be asynchronous?

crates/iceberg-catalog/src/implementations/postgres/task_queues/mod.rs
c-thiel: I think the way you implemented (retries) is clearer. If we had multiple tasks for a single table it's less clear how to perform the update. One task per job per table seems clearest to me.

tests/python/tests/test_spark.py
```diff
@@ -301,6 +303,81 @@ def test_drop_table_purge_http(spark, warehouse: conftest.Warehouse, storage_con
...
+        [
+            "v1",
+            "warehouse",
+            str(warehouse.warehouse_id),
+            "deleted_tabulars",
+            "undrop",
+        ]
+    ))
+
+    print(json.dumps({
```
c-thiel, line 363: Remove debug print?

c-thiel (commented):
LGTM - just a handful of comments.
Also check my comments on https://github.com/lakekeeper/console/pull/44#pullrequestreview-2491203380 on the error - not sure if we should fix it in rust or the UI.

Apart from that, could you update the docs/docs/concept.md Soft Deletion section to state and remove the comment that its not here yet?
In docs/docs is always the current WIP docs

crates/iceberg-catalog/src/api/management/v1/warehouse/undrop.rs
twuebi: removed namespace undrop target as discussed

c-thiel (approved):

PR #518, opened 2024-11-17
feat: split table metadata: migration

twuebi:


crates/iceberg-catalog/src/implementations/postgres/catalog.rs
```diff
@@ -576,6 +576,7 @@ impl Catalog for super::PostgresCatalog {
             &catalog_state.read_pool(),
             None,
             pagination_query,
+            false,
```
c-thiel, line 579: If we hook into the proc macro of migration, we could remove it now.
This would enable users to skip 0.5 on migrations

c-thiel (approved):

PR #524, opened 2024-11-19
fix: prepend a version count to metadata files

### LINKED ISSUE #522: fix: create new metadata with version prepended
Python uses:

```python
version_str = f"{new_version:05d}"
return f"{location}/metadata/{version_str}-{uuid.uuid4()}.metadata.json"
```

Java example: 00000-d0407fb2-1112-4944-bb88-c68ae697e2b4.gz.metadata.json

twuebi:
we're going for the string parsing route here, we could also store the version number in the db, wdyt @c-thiel?

Python is parsing strings:

```py
    @staticmethod
    def _parse_metadata_version(metadata_location: str) -> int:
        """Parse the version from the metadata location.

        The version is the first part of the file name, before the first dash.
        For example, the version of the metadata file
        `s3://bucket/db/tb/metadata/00001-6c97e413-d51b-4538-ac70-12fe2a85cb83.metadata.json`
        is 1.
        If the path does not comply with the pattern, the version is defaulted to be -1, ensuring
        that the next metadata file is treated as having version 0.

        Args:
            metadata_location (str): The location of the metadata file.

        Returns:
            int: The version of the metadata file. -1 if the file name does not have valid version string
        """
        file_name = metadata_location.split("/")[-1]
        if file_name_match := TABLE_METADATA_FILE_NAME_REGEX.fullmatch(file_name):
            try:
                uuid.UUID(file_name_match.group(2))
            except ValueError:
                return -1
            return int(file_name_match.group(1))
        else:
            return -1
```

Currently, this PR will fail any metadata file that's not a uuid (len 36) /  uuid with prepended parsable usize

closes #522

### DIFF
--- a/crates/iceberg-catalog/src/catalog/tables.rs
+++ b/crates/iceberg-catalog/src/catalog/tables.rs
@@ -537,7 +537,7 @@ impl<C: Catalog, A: Authorizer + Clone, S: SecretStore>
             updates,
         )?;
         let location = previous_table.metadata_location;
-        let next_metadata_count = if let Some(loc) = location {
+        let next_metadata_count = if let Some(loc) = location.as_ref() {
             extract_count_from_metadata_location(loc)? + 1
         } else {
             0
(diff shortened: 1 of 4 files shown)


crates/iceberg-catalog/src/catalog/tables.rs
```diff
@@ -1149,6 +1168,34 @@ impl<C: Catalog, A: Authorizer + Clone, S: SecretStore>
...
+        None,
+    ))?;
+    usize::from_str(front).map_err(|e| {
+        ErrorModel::internal(
+            format!("Failed to parse integer counter in filename: {e}"),
+            "InvalidTableFileName",
+            Some(Box::new(e)),
+        )
+        .into()
+    })
```
twuebi, line 1196: alternatively, we just do unwrap_or(..) here

c-thiel (approved):

PR #531, opened 2024-11-19
feat: server now refuses to start if migrations are missing

twuebi:


crates/iceberg-catalog-bin/src/wait_for_db.rs
```diff
@@ -51,8 +51,8 @@ pub(crate) async fn wait_for_db(
 
             counter += 1;
             if counter > retries {
-                tracing::error!("Ran out of retries while waiting for migrations.");
-                anyhow::bail!("Ran out of retries while waiting for migrations.");
+                tracing::error!("Database is not up to date with binary.");
+                anyhow::bail!("Database is not up to date with binary.");
```
c-thiel, line 55: Lets offer a solution in the message

c-thiel (approved):

PR #543, opened 2024-11-21
feat: Add operator role

c-thiel:


crates/iceberg-catalog/src/api/management/v1/bootstrap.rs
```diff
@@ -24,6 +24,9 @@ pub enum AuthZBackend {
 pub struct BootstrapRequest {
     /// Set to true if you accept LAKEKEEPER terms of use.
     pub accept_terms_of_use: bool,
+    /// If set to true, the calling user is treated as an operator and obtain
+    /// a corresponding role. If not specified, the user is treated as a human.
+    pub is_operator: Option<bool>,
```
twuebi, line 29: Let's make this not an option but either just bool with a serde(default) or an enum with serde(default)

twuebi (approved):

PR #581, opened 2024-11-27
chore: Fix release pipe / Cross

c-thiel:


c-thiel:
Let's move to ARM builders. This won't work.

Issue #623, opened 2024-12-05
Unresponsive During High-Volume Data Ingestion


menne95:
## Summary

When attempting to ingest a large volume of data (~3 million records per second), lakekeeper becomes unresponsive (it seems lakekeeper is stuck like not doing anything) while my data ingestor is waiting for lakekeeper to respond.

During these tests, the logs show repeated messages like the one below (meantime data was getting ingested)

```txt
{"timestamp":"2024-12-05T07:35:23.518967Z","level":"INFO","message":"No task found","target":"iceberg_catalog::implementations::postgres::task_queues::tabular_expiration_queue","filename":"crates/iceberg-catalog/src/implementations/postgres/task_queues/tabular_expiration_queue.rs","line_number":118,"span":{"name":"pick_new_task"},"spans":[{"name":"pick_new_task"}]}
```

however, we have seen many logs like the below one which caused the ingestor apps to slowly reduce the amount of data ingested per second (or even for newly deployed ingestor apps, to simply wait for the lakekeeper to reply)

```txt
{"timestamp":"2024-12-04T12:38:22.815238Z","level":"WARN","message":"acquired connection, but time to acquire exceeded slow threshold","aquired_after_secs":5.125707121,"slow_acquire_threshold_secs":2.0,"target":"sqlx::pool::acquire","filename":"/home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/sqlx-core-0.8.2/src/pool/inner.rs","line_number":308,"span":{"method":"POST","request_id":"2de3d474cf90ba29b84023183cc38540","uri":"/catalog/v1/9ef83972-b237-11ef-8ad7-a7d83451cc2d/namespaces/testing-full-load-v5/tables/my-table","version":"HTTP/1.1","name":"request"},"spans":[{"method":"POST","request_id":"2de3d474cf90ba29b84023183cc38540","uri":"/catalog/v1/9ef83972-b237-11ef-8ad7-a7d83451cc2d/namespaces/testing-full-load-v5/tables/my-table","version":"HTTP/1.1","name":"request"}]}
```

From a resource perspective, lakekeeper pod is using almost no CPU and memory. 

## Configuration

We are deploying lakekeeper using the helm chart version [0.1.3](https://github.com/lakekeeper/lakekeeper-charts/tree/lakekeeper-0.1.3) with the following custom values

```yaml
       catalog:
          replicas: 1
          resources:
            limits:
              cpu: 1
              memory: 4Gi
[...]

PR #638, opened 2024-12-07
fix: potential deadlock also for views

twuebi:
further addresses #623 

Same as #636 but for views, even though we were committing here, we still do it after reading secret which could hang.

### DIFF
--- a/crates/iceberg-catalog/src/catalog/views/load.rs
+++ b/crates/iceberg-catalog/src/catalog/views/load.rs
@@ -74,7 +74,8 @@ pub(crate) async fn load_view<C: Catalog, A: Authorizer + Clone, S: SecretStore>
 
     let view_location = parse_view_location(&view_metadata.location)?;
 
-    // We don't commit the transaction yet, first we need to write the metadata file.
+    t.commit().await?;
+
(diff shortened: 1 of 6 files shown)


crates/iceberg-catalog/src/catalog/views/load.rs
```diff
@@ -74,7 +74,8 @@ pub(crate) async fn load_view<C: Catalog, A: Authorizer + Clone, S: SecretStore>
 
     let view_location = parse_view_location(&view_metadata.location)?;
 
-    // We don't commit the transaction yet, first we need to write the metadata file.
+    t.commit().await?;
```
c-thiel, line 77: If we commit here, the rest of the function must be infallible. The commit succeeded so we must return an ok value to the client. Lets also double check tables. Alternatively we can get the secret earlier?

twuebi: This is a load fn, a read only transaction, committing it should have no sideffects but dropping the connection

c-thiel: Ah - you are right! Should have used a proper screen.

c-thiel (approved):

Issue #657, opened 2024-12-11
non 20x status codes are not documented on openapi


twuebi:
Our openapi docs do not document non-20x status code responses. This shouldn't be hard to do for us since we use the same model for all errors.

We'll have to implement `utoipa::ToSchema` on `IcebergErrorResponse`. That won't be straight forward since it contains a `pub source: Option<Box<dyn std::error::Error + Send + Sync + 'static>>,` which isn't a thing we can send via json. 

We should probably split the internal error model from the external one. Initially, we can simply copy the internal one, remove the source field and just convert it  

It appears that there's no router level annotation which says all 4XX status codes are `ErrorModel`. Gladly, there's an option to specify ranges, so we can write `4XX` instead of enumerating all client error codes in the handler annotations uf utoipa.

PR #658, opened 2024-12-11
feat(openapi): document error models in openapi

twuebi:
partially addresses #657

### DIFF
--- a/crates/iceberg-ext/src/catalog/rest/error.rs
+++ b/crates/iceberg-ext/src/catalog/rest/error.rs
@@ -20,7 +20,6 @@ macro_rules! impl_into_response {
 #[cfg(feature = "axum")]
 pub(crate) use impl_into_response;
 use typed_builder::TypedBuilder;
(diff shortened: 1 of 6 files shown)


crates/iceberg-ext/src/catalog/rest/error.rs
```diff
@@ -234,6 +235,30 @@ impl ErrorModel {
     }
 }
 
+#[derive(Debug, Serialize, Deserialize, ToSchema)]
+pub struct ApiErrorModel {
```
c-thiel, line 239: I am wondering if we should ditch the `iceberg-ext` `ErrorModel` completely eventually.

c-thiel (approved):

crates/iceberg-ext/src/catalog/rest/error.rs
twuebi: As in move it into iceberg-catalog?

c-thiel (approved):

PR #659, opened 2024-12-12
fix: remove unused relation from openfga schema

twuebi:


c-thiel (commented):
We must also remove it from rust, i.e. https://github.com/lakekeeper/lakekeeper/blob/44c43aea5891eea3e15d11eef151657c30f68826/crates/iceberg-catalog/src/service/authz/mod.rs#L98

c-thiel (approved):

PR #675, opened 2024-12-16
feat(tables): load table credentials

twuebi:


crates/iceberg-catalog/src/service/authz/mod.rs
```diff
@@ -131,6 +131,12 @@ impl TableUuid for TableIdentUuid {
     }
 }
 
+impl TableUuid for TabularDetails {
```
c-thiel, line 134: Should we move the `TableUuid` trait out of authz?
Feels odd to implement it in the authz module for TabularDetails

crates/iceberg-catalog/src/catalog/tables.rs
```diff
@@ -460,15 +427,78 @@ impl<C: Catalog, A: Authorizer + Clone, S: SecretStore>
...
+        let TableParameters { prefix, table } = parameters;
+        let warehouse_id = require_warehouse_id(prefix)?;
+
+        let list_flags = ListFlags {
+            include_active: true,
+            include_staged: false,
+            include_deleted: false,
+        };
+
+        let (mut t, table_id, storage_permissions) = Self::load_prelude(
```
c-thiel, line 463: I don't like the scope of the load_prelude function. Maybe we can remove transaction creation from it, add the NoStoragePermissions Error to it and give it a better name, such as `require_data_action`.
I was missing authorization here and wouldn't expect it to happen in a prelude.

crates/iceberg-catalog/src/service/authz/mod.rs
twuebi: moved TabularDetails, it's also in service/mod.rs now next to TableIdentUuid, would you like the trait to move there as well?

c-thiel (approved):

Issue #560, opened 2024-11-24
Check if cargo hakari speeds up builds


twuebi:
https://crates.io/crates/cargo-hakari claims it can speed up builds in large workspaces, also in ci, we should try it

twuebi:
didn't seem to have an effect after removing cargo chef from ci from-scratch builds

PR #687, opened 2024-12-17
chore(ci): speed up builds with cargo hakari

twuebi:
#560


twuebi:
after removing cargo chef from ci in #699 this doesn't seem to have an impact anymore

PR #691, opened 2024-12-19
Fix API page creation

bitsondatadev:
I've added Redocly as the new API page, as well as, relocated the api pages under the versioned docs. I also added the base lakekeeper colors in css variables. 

A few things to note are that the font styles and such are still a bit off, but I did modify the API docs pages to take up all the space other than the navigation tab. This change will temporarily be in HTML `<style>` tags in the header until @munabedan has the time to do this the proper mkdocs style way. I tried to take a stab at it and always am reminded how much css and scss workflows confound me, so I'll leave it to the pros.

I removed the `neoteroi.mkdocsoad` plugin as it is not well maintained and doesn't render a lot of elements needed. Similarly for the other community mkdocs swagger plugin, it didn't render well for all items and would have required a lot of fixing on our part.

Redocly seems to be the accepted UI that most use to render open api user interfaces.

Here's a preview of Redocly embeds:
redocly


c-thiel:
Both looking much better than what we had before!
Any preference from your side?

Could you change the following paths to the new locations as well?:

- In the rust crate, you can see it here in the failing test: https://github.com/lakekeeper/lakekeeper/actions/runs/12407088250/job/34636544730?pr=691
- In the main just file that re-creates the openapi specs (update-openfga, update-management-openapi, add-return-uuid-to-rest-openapi)
- In `.github/unittests.yaml` where we validate that the openapi spec is up-to-date and try to generate the UI client from it. 


I believe after that tests should start working again

site/mkdocs.yml
```diff
@@ -6,6 +6,7 @@ edit_uri: blob/main/docs/
 
 theme:
   name: material
+  custom_dir: overrides
```
c-thiel, line 9: The overrides dir is missing in this PR - serve fails with:
```
ERROR   -  Config value 'theme': The path set in custom_dir ('/Users/cht/code/ht/iceberg-rest/iceberg-catalog/site/overrides')
           does not exist.
```

c-thiel (changes requested):

site/mkdocs.yml
bitsondatadev: Ah yip...removed the only file inside the directory that I was toying with and ergo removed the directory from GitHub. I'll remove this for now. Likely will be added later when Muna does his magic.

site/mkdocs.yml
```diff
@@ -42,7 +42,6 @@ theme:
   features:
     - navigation.tabs
     - navigation.tabs.sticky
-    - navigation.path
```
bitsondatadev, line 45: This is an [MkDocs Material Insiders](https://squidfunk.github.io/mkdocs-material/insiders/) feature. Removing it to avoid confusion.

c-thiel (dismissed):

c-thiel (approved):

PR #694, opened 2024-12-19
chore(ci): link with lld on linux and windows

twuebi:


twuebi:
doesn't seem to have a mesaurable impact on here

PR #706, opened 2024-12-24
feat: Check Endpoint for single permissions

c-thiel:


crates/iceberg-catalog/src/api/management/mod.rs
```diff
@@ -926,13 +926,12 @@ pub mod v1 {
         )
     )]
     async fn undrop_tabulars<C: Catalog, A: Authorizer + Clone, S: SecretStore>(
-        Path(warehouse_id): Path<uuid::Uuid>,
+        Path(_warehouse_id): Path<uuid::Uuid>,
```
twuebi, line 929: why keep this if its unused? Do you want to introduce undrop for a whole warehouse later?

crates/iceberg-catalog/src/service/authz/implementations/openfga/api.rs
```diff
@@ -1285,6 +1289,9 @@ async fn update_role_assignments_by_id<C: Catalog, S: SecretStore>(
         update_warehouse_assignments_by_id,
     ),
     components(schemas(
+        CheckAction,
```
twuebi, line 1292: I think utoipa >0.5 can collect schemas on its own now

crates/iceberg-catalog/src/api/management/mod.rs
c-thiel: I think we should keep it in the path so that we can unrop a table by name later.
Query engines don't know IDs and thus need this translation layer eventually.

crates/iceberg-catalog/src/service/authz/implementations/openfga/check.rs
```diff
@@ -0,0 +1,644 @@
...
+use iceberg::{NamespaceIdent, TableIdent};
+use openfga_rs::CheckRequestTupleKey;
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, utoipa::ToSchema)]
+#[serde(rename_all = "kebab-case", tag = "type")]
+/// Represents an action on an object
+pub(super) enum CheckAction {
+    Server {
+        action: ServerAction,
```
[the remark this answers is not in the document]
c-thiel: I had two reasons to use snake case:
1. Permissions could be part of SQL eventually - kebab case is very unhandy there as it needs to be escaped
2. OpenFGA uses snake case internally

We could use mappings for both cases of course as well. Not sure if this is worth a breaking change.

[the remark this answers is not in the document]
twuebi: let's leave it then, wanted to make sure this wasn't accidental

twuebi (approved):

PR #746, opened 2025-01-08
chore: remove solved issue from deny

twuebi:


twuebi:
we still depend on it via iceberg-rust -> rust-decimal -> borsh

Issue #572, opened 2024-11-26
Emit cloudevents on undrop tabulars


twuebi:

brgr-s:
I would like to take a look at this one.

twuebi:
done

PR #759, opened 2025-01-13
WIP: emit event for undropTabulars #572

Grongrilla:
WIP version has a rather expensive call to `Catalog` to get namespace and tabular name. There are certainly more efficient ways to get those informations (based on the single supported catalog backend of postgresql, atm), especially because we call into the `Catalog` for each "undropped tabular", but I did not want to extend the trait _just_ to emit an event in this single case.


crates/iceberg-catalog/src/api/management/v1/warehouse/mod.rs
```diff
@@ -676,7 +679,37 @@ pub trait Service<C: Catalog, A: Authorizer, S: SecretStore> {
...
-        // TODO: emit event
+        let num_tabulars = tabs.len();
+        let list_flags = ListFlags {
+            include_active: true,
+            include_staged: true,
+            include_deleted: true,
+        };
+        for (i, t) in tabs.iter().enumerate() {
+            if let Ok(Some(table_metadata)) =
+                C::get_table_metadata_by_id(warehouse_id, *t, list_flags, catalog.clone()).await
```
twuebi, line 690: IMO, we should change the undrop interface in the catalog trait to return the fields instead of calling the DB for every table we get here.

brgr-s:
Closing PR as one commit is broken and it was easier to redo it.

PR #775, opened 2025-01-21
feat: Register table endpoint

c-thiel:


crates/iceberg-catalog/src/catalog/io.rs
```diff
@@ -66,18 +67,34 @@ pub(crate) async fn read_file(file_io: &FileIO, file: &Location) -> Result<Vec<u
...
             .await
             .map_err(|e| IoError::FileRead(Box::new(e)))
             .map(Into::into)
     })
-    .await
+    .await?;
+
+    if file.as_str().ends_with(".gz.metadata.json") {
+        let codec = CompressionCodec::Gzip;
+        let content = codec.decompress(&content)?;
```
twuebi, line 85: spawn_blocking?

crates/iceberg-catalog/src/catalog/io.rs
```diff
@@ -66,18 +67,34 @@ pub(crate) async fn read_file(file_io: &FileIO, file: &Location) -> Result<Vec<u
...
+        Ok(content)
+    }
+}
+
+pub(crate) async fn read_metadata_file(
+    file_io: &FileIO,
+    file: &Location,
+) -> Result<TableMetadata, IoError> {
+    let content = read_file(file_io, file).await?;
+    serde_json::from_slice(&content).map_err(IoError::TableMetadataDeserialization)
```
twuebi, line 97: metadata files can be quite large, do you think parsing can lead to blocking here? maybe we just wrap both read_file & from_slice in spawn_blocking?

crates/iceberg-catalog/src/catalog/io.rs
c-thiel: `write_metadata_file` should then also spawn_blocking right?

crates/iceberg-catalog/src/catalog/tables.rs
```diff
@@ -1422,6 +1513,21 @@ pub(super) fn parse_location(location: &str, code: StatusCode) -> Result<Locatio
         .map_err(Into::into)
 }
 
+pub(super) fn require_allowed_location(
+    location: &Location,
+    storage_profile: &StorageProfile,
+) -> Result<()> {
+    if !storage_profile.is_allowed_location(location) {
+        return Err(ErrorModel::bad_request(
+            format!("Specified table location is not allowed: {location}"),
```
twuebi, line 1522: add a small explanation what allowed locations are?

c-thiel: I added a docstring to the `is_allowed_location` function.
I don't want to comment it in `table.rs`, because it might change in the future. That's what we have the function for. Currently its only sublocations of the one base location of a storage profile, eventually we might allow multiple locations per Storage Profile.

twuebi (approved):

PR #802, opened 2025-02-05
fix: parsing of pg sslmode should be case-insensitive

c-thiel:


crates/iceberg-catalog/src/config.rs
```diff
@@ -59,7 +59,6 @@ fn get_config() -> DynAppConfig {
     // Fail early if the base_uri is not a valid URL
     config.s3_signer_uri_for_warehouse(WarehouseIdent::from(uuid::Uuid::new_v4()));
     config.base_uri_catalog();
-    config.base_uri_management();
```
twuebi, line 62: why remove this?

c-thiel: We don't need it - I removed the whole method. I don't like that we need catalog either - its only required because of the design of the config endpoint

c-thiel: I think ideally we would get rid of the base url altogether.
We might consider parsing it from the URL or the X-forwarded-for header.

crates/iceberg-catalog/src/config.rs
```diff
@@ -612,16 +617,22 @@ mod test {
     #[allow(unused_imports)]
     use super::*;
 
+    #[test]
+    fn test_pg_ssl_mode_case_insensitive() {
+        figment::Jail::expect_with(|jail| {
+            jail.set_env("LAKEKEEPER_TEST__PG_SSL_MODE", "DISABLED");
```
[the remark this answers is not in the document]
c-thiel: Added tests - especially the good case of "disabled" is kind of important :D

twuebi (approved):

PR #811, opened 2025-02-10
feat: warehouse stats

twuebi:


corleyma:
Any plans to expose these stats as (maybe optional) prometheus metrics?  I think we've discussed something similar in the past and deferred due to concerns about cardinality, but rich but optional instrumentation could be useful for folks who aren't as concerned about cardinality in their deployments and want easy integration of telemetry with their existing observability stack.

twuebi:
Hi @corleyma, the cardinality woes still exist, we'd be fine with exposing these stats via prometheus as opt-in.

twuebi:
https://github.com/lakekeeper/lakekeeper/issues/817

crates/iceberg-catalog/migrations/20250108131004_statistics.sql
```diff
@@ -0,0 +1,117 @@
...
+CREATE OR REPLACE FUNCTION get_stats_interval() RETURNS interval AS
+$$
+BEGIN
+    RETURN interval '1 ' || get_stats_interval_unit();
+END;
+$$ LANGUAGE plpgsql;
+
+create table warehouse_statistics
+(
+    stats_id         int generated always as identity primary key,
```
c-thiel, line 24: I think we should use a UUID instead (`uuid_generate_v1mc()`).
Identity can feel weird as they are not consecutive in the domain of interest - the `warehouse_id`.

Do we gain anything from identity?

crates/iceberg-catalog/src/api/management/mod.rs
```diff
@@ -834,6 +837,71 @@ pub mod v1 {
...
+
+    /// Get warehouse statistics
+    ///
+    /// Get statistics about the warehouse, currently table and view counts over time.
+    ///
+    /// We lazily create a new statistics entry every hour, in between hours, the existing entry is
+    /// being updated. If there's a change at created_at + 1 hour, a new entry is created.
+    /// If there's been no change, no new entry is created, meaning there may be gaps.
+    ///
+    /// Example:
```
c-thiel, line 869: Lets add the information that the examples below are excluding the date information which is present in the message

crates/iceberg-catalog/src/api/management/mod.rs
```diff
@@ -834,6 +837,71 @@ pub mod v1 {
...
+    /// - 00:16:32: warehouse created:
+    ///     - created_at: 00:16:32, valid_until: 01:00:00, updated_at: null, 0 tables, 0 views
```
c-thiel, line 871: As discussed, lets move `valid_until` to the front and give it a more generic name like `timestamp`or so

crates/iceberg-catalog/src/api/management/v1/warehouse/mod.rs
```diff
@@ -230,6 +230,33 @@ impl axum::response::IntoResponse for CreateWarehouseResponse {
...
+    ///
+    /// We lazily create a new statistics entry every hour, in between hours, the existing entry
+    /// is being updated. If there's a change at `created_at` + 1 hour, a new entry is created. If
+    /// there's no change, no new entry is created.
+    pub valid_until: chrono::DateTime<chrono::Utc>,
+}
+
+#[derive(Debug, Serialize, ToSchema)]
+#[serde(rename_all = "kebab-case")]
+pub struct WarehouseStatsResponse {
```
c-thiel, line 251:
```suggestion
pub struct WarehouseStatisticsResponse {
```

crates/iceberg-catalog/src/api/management/v1/warehouse/mod.rs
```diff
@@ -377,6 +404,31 @@ pub trait Service<C: Catalog, A: Authorizer, S: SecretStore> {
...
+        context: ApiContext<State<A, C, S>>,
+        request_metadata: RequestMetadata,
+    ) -> Result<WarehouseStatsResponse> {
+        // ------------------- AuthZ -------------------
+        let authorizer = context.v1_state.authz;
+        authorizer
+            .require_warehouse_action(
+                &request_metadata,
+                warehouse_id,
+                &CatalogWarehouseAction::CanGetMetadata,
```
c-thiel, line 419: On second thought, I am not sure if guests, that have permissions only via inheritance, should be able to see stats. Lets leave it for now, but we might want to restrict this later if stats reveal more sensitive information about warehouses - i.e. occurances of table properties etc.

For those more sensitive stats we should introduce a new action that requires direct assignment to the warehouse or entities above - and can not "bubble up". 
We could use `describe` for it.

crates/iceberg-catalog/migrations/20250108131004_statistics.sql
twuebi: not really, was looking for something that'd retain at least some locality with lack of uuid v7 support in pg, didnt' realize that v1mc is somewhat like a v7, changing it now.

crates/iceberg-catalog/src/api/management/mod.rs
twuebi: renamed to timestamp

crates/iceberg-catalog/src/api/management/v1/warehouse/mod.rs
twuebi: renamed

crates/iceberg-catalog/src/api/management/mod.rs
twuebi: mention that dates are omitted

c-thiel (dismissed):

PR #814, opened 2025-02-12
feat!: Migrate Authentication to Limes, Support Unlimited Authenticators, Customizable Authentication

c-thiel:
Todo: prefer oid over sub


Cargo.toml
```diff
@@ -81,7 +81,7 @@ kube = { version = "0.98", default-features = false, features = [
     "rustls-tls",
     "client",
 ] }
-k8s-openapi = { version = "0.24.0", features = ["v1_28"] }
+k8s-openapi = { version = "0.24.0", features = ["v1_32"] }
```
twuebi, line 84: do we still need this in lakekeeper?

crates/iceberg-catalog-bin/src/serve.rs
```diff
@@ -144,9 +144,120 @@ pub(crate) async fn serve(bind_addr: std::net::SocketAddr) -> Result<(), anyhow:
...
+        None
+    };
+
+    let authn_oidc = if let Some(uri) = CONFIG.openid_provider_uri.clone() {
+        let mut authenticator = limes::jwks::JWKSWebAuthenticator::new(
+            uri.as_ref(),
+            Some(std::time::Duration::from_secs(3600)),
+        )
+        .await?
+        .set_idp_id("oidc");
```
twuebi, line 181: are we using this elsewhere? Should we make it a const / env?

crates/iceberg-catalog/src/service/authz/mod.rs
```diff
@@ -153,6 +153,9 @@ pub enum NamespaceParent {
 
 #[async_trait::async_trait]
 /// Interface to provide AuthZ functions to the catalog.
+/// The provided `Actor` arguments of the all methods except `check_actor`
+/// are assumed to be valid. Please ensure to call `check_actor` before, preferably
+/// during Authentication.
```
twuebi, line 158: maybe specify what you mean by valid in this context?

Cargo.toml
c-thiel: Yes - won't compile otherwise as the k8s dependency pulled in by limes doesn't know which k8s API Version to build for.
Good point actually - let me try to only depend on it in the bin crate.

c-thiel: Actually no, because we fixed it in limes.
I think removing it there should be done in a later version. Removing it from lakekeeper for now

crates/iceberg-catalog/src/service/authz/mod.rs
c-thiel: All other occurances are in tests - no other places

crates/iceberg-catalog/src/catalog/config.rs
```diff
@@ -27,22 +27,20 @@ impl<A: Authorizer + Clone, C: Catalog, S: SecretStore>
...
                 .ok_or_else(|| {
-                    // ToDo Christian: Split Project into separate endpoint, Use name
                     ErrorModel::bad_request(
-                        "No project provided. Please provide warehouse as: <project-name>/<warehouse-name>",
-                        "GetConfigNoProjectProvided", None)
+                        "No project provided. Please provide the `x-project-id` header",
+                        "GetConfigNoProjectProvided",
+                        None,
+                    )
                 })?;
```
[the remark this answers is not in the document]
c-thiel: Done

crates/iceberg-catalog/src/service/authz/mod.rs
twuebi: I meant for future reference it could make sense to expand a bit on validity, i.e. it has been checked that the currently authenticated user is allowed to use this actor?

crates/iceberg-catalog-bin/src/serve.rs
c-thiel: Done - its in the bin though, so really only usable there.

twuebi (approved):

PR #818, opened 2025-02-13
feat(stats): endpoints call stats

twuebi:


crates/iceberg-catalog/src/api/endpoints.rs
```diff
@@ -0,0 +1,275 @@
...
+
+static MAP: LazyLock<HashMap<&str, Endpoints>> = LazyLock::new(|| {
+    Endpoints::iter()
+        .filter(|e| {
+            !matches!(
+                e,
+                Endpoints::ManagementGetPermissions
+                    | Endpoints::ManagementPostPermissions
+                    | Endpoints::ManagementHeadPermissions
+                    | Endpoints::ManagementDeletePermissions
```
c-thiel, line 95: Can we add a comment here why they are special?

crates/iceberg-catalog/src/api/iceberg/mod.rs
```diff
@@ -203,33 +203,10 @@ pub mod v1 {
...
-        "POST /v1/{prefix}/transactions/commit".into(),
-        "GET /v1/{prefix}/namespaces/{namespace}/views".into(),
-        "POST /v1/{prefix}/namespaces/{namespace}/views".into(),
-        "GET /v1/{prefix}/namespaces/{namespace}/views/{view}".into(),
-        "POST /v1/{prefix}/namespaces/{namespace}/views/{view}".into(),
-        "DELETE /v1/{prefix}/namespaces/{namespace}/views/{view}".into(),
-        "HEAD /v1/{prefix}/namespaces/{namespace}/views/{view}".into(),
-        "POST /v1/{prefix}/views/rename".into(),
-    ]
+    crate::api::endpoints::Endpoints::catalog()
```
c-thiel, line 206: I think this still contains the /catalog prefix which is not allowed here. Paths should start with /v1.
Let's write a test that checks if a handful of expected strings are present.

crates/iceberg-catalog/src/api/management/mod.rs
```diff
@@ -1099,6 +1118,10 @@ pub mod v1 {
                     "/warehouse/{warehouse_id}/deleted_tabulars/undrop",
```
c-thiel, line 1118: We don't have a good way to mark this deprecated right?

crates/iceberg-catalog/src/api/management/v1/project.rs
```diff
@@ -230,6 +231,76 @@ pub trait Service<C: Catalog, A: Authorizer, S: SecretStore> {
...
+            context.v1_state.catalog,
+        )
+        .await
+    }
+}
+
+#[derive(Deserialize, Serialize, Debug, ToSchema)]
+pub struct EndpointStatistic {
+    pub count: i64,
+    pub http_string: String,
```
c-thiel, line 267: Maybe `http_path` or so? Maybe chatGPT has a better idea. I don't like having the type as part of the name.

crates/iceberg-catalog/src/api/management/v1/project.rs
```diff
@@ -230,6 +231,76 @@ pub trait Service<C: Catalog, A: Authorizer, S: SecretStore> {
...
+                .unwrap_or_else(|| chrono::Duration::days(1)),
+            request.status_codes.as_deref(),
+            context.v1_state.catalog,
+        )
+        .await
+    }
+}
+
+#[derive(Deserialize, Serialize, Debug, ToSchema)]
+pub struct EndpointStatistic {
```
c-thiel, line 265: Lets add some docs for all fields

crates/iceberg-catalog/src/api/management/v1/project.rs
```diff
@@ -230,6 +231,76 @@ pub trait Service<C: Catalog, A: Authorizer, S: SecretStore> {
...
+#[derive(Deserialize, Serialize, Debug, ToSchema)]
+pub struct EndpointStatisticsResponse {
+    pub timestamps: Vec<chrono::DateTime<chrono::Utc>>,
+    pub stats: Vec<Vec<EndpointStatistic>>,
+    pub previous_page_token: Option<String>,
+    pub next_page_token: Option<String>,
+}
+
+#[derive(Debug)]
+pub enum RangeSpecifier {
```
c-thiel, line 284: Is this used anywhere?

crates/iceberg-catalog/migrations/20250207162731_endpoint_stats.sql
```diff
@@ -0,0 +1,89 @@
...
+    'management-post-warehouse-deleted-tabulars-undrop1',
+    'management-post-warehouse-deleted-tabulars-undrop2',
+    'management-post-warehouse-delete-profile',
+    'management-get-permissions',
+    'management-post-permissions',
+    'management-head-permissions',
+    'management-delete-permissions');
+
+
+create table endpoint_statistics
```
c-thiel, line 73: Just a comment that we don't for get to add the primary access as index.

c-thiel (commented):
Got halfway through, rest tomorrow.

crates/iceberg-catalog/src/implementations/postgres/endpoint_statistics.rs
```diff
@@ -0,0 +1,322 @@
...
+                        )
+                    },
+                )
+                .multiunzip::<_>();
+            let whn = warehouse_names
+                .iter()
+                .filter_map(|w| w.as_deref().map(ToString::to_string))
+                .collect::<Vec<_>>();
+            let warehouse_ids = sqlx::query!(
+                r#"select warehouse_name, warehouse_id from warehouse where warehouse_name = any($1)"#,
```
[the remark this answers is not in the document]
c-thiel: By adding the project ID we can also handle another case more explicitly:
What do we do if we receive calls to non-existent projects.
It would be good if we could detect this, so that we can log it out explicitly.
If this happens, it probably should not be an error log.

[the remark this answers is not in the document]
c-thiel: I think we shouldn't have this lookup in every loop iteration. Could we just look up all warehouse_name / project_id combinations in the received stats in a single transaction?

crates/iceberg-catalog/src/service/endpoint_statistics.rs
```diff
@@ -0,0 +1,271 @@
...
+                EndpointStatisticsMessage::EndpointCalled {
+                    request_metadata,
+                    response_status,
+                    path_params,
+                    query_params,
+                } => {
+                    let warehouse = Self::maybe_get_warehouse_ident(&path_params);
+
+                    let Some(matched_path) = request_metadata.matched_path() else {
+                        tracing::error!("No path matched.");
```
c-thiel, line 173: Is this worth an error log? I assume that happens quite often with typos etc.
I am not sure that we want to print this as error.
I think I would use something below info.

crates/iceberg-catalog/src/api/management/mod.rs
twuebi: not really, we also cannot easily drop the route with `deleted_tabulars` from the openapi specs as that would be a breaking change. We could duplicate the handler function, mark the deprecated route as deprecated in the api doc & remove it in a later release.

crates/iceberg-catalog/migrations/20250207162731_endpoint_stats.sql
twuebi: added 3 indexes

crates/iceberg-catalog/src/service/endpoint_statistics.rs
```diff
@@ -0,0 +1,291 @@
...
+            statistic_sinks: stat_sinks,
+            flush_interval,
+            flush_mode,
+        }
+    }
+
+    pub async fn run(mut self) {
+        let mut last_update = tokio::time::Instant::now();
+        loop {
+            if !matches!(self.flush_mode, FlushMode::Manual)
```
c-thiel, line 151: Should we open a span for all the traces in this function?
This doesn't run as part of an API span, so adding one for this call of the sheduled funciton would be helpful to group all related messages.
We are doing the same in middle for a refresh run.

crates/iceberg-catalog/src/service/endpoint_statistics.rs
```diff
@@ -0,0 +1,291 @@
...
+    pub status_code: StatusCode,
+    pub warehouse: Option<WarehouseIdent>,
+    // probably only relevant for config calls
+    pub warehouse_name: Option<String>,
+}
+
+#[derive(Debug, PartialEq, Copy, Clone)]
+pub enum FlushMode {
+    Automatic,
+    Manual,
```
c-thiel, line 119: We use manual only in tests.
Maybe worth cfg(test) it?
Would need to change a match arm below as well.

docs/docs/api/management-open-api.yaml
```diff
@@ -2110,6 +2158,113 @@ components:
...
+            Format: `METHOD /path/to/endpoint`
+        status_code:
+          type: integer
+          format: int32
+          description: The status code of the response.
+          minimum: 0
+        updated_at:
+          type:
+          - string
+          - 'null'
```
c-thiel, line 2195: Is this OK? Shouldn't it just be a string that is not required?
There are more like this though.

crates/iceberg-catalog/src/api/management/v1/project.rs
```diff
@@ -230,6 +232,144 @@ pub trait Service<C: Catalog, A: Authorizer, S: SecretStore> {
...
+    pub previous_page_token: String,
+    /// Token to get the next page of results.
+    ///
+    /// Inverse of `previous_page_token`, see its documentation above.
+    pub next_page_token: String,
+}
+
+#[derive(Debug, Serialize, Deserialize, ToSchema, Clone)]
+#[serde(rename_all = "kebab-case", tag = "type")]
+pub enum RangeSpecifier {
```
c-thiel, line 327: This will be a global type name in the API spec  and thus the name should be a bit more precise.
We might introduce different range filters in the future.

My idea:

```rust
#[derive(Debug, Serialize, Deserialize, ToSchema, Clone)]
#[serde(rename_all = "kebab-case", tag = "type")]
pub enum TimeWindowSelector {
    Window {
        /// End timestamp of the time window
        window_end: chrono::DateTime<chrono::Utc>,
        /// Duration/span of the time window
        /// 
        /// The effective time range = `window_end` - `window_duration` to `window_end`
        window_duration: chrono::Duration,
    },
    PageToken {
        /// Opaque Token from previous response for paginating through time windows
        /// 
        /// Use the `next_page_token` or `previous_page_token` from a previous response
        token: String,
    },
}
```

crates/iceberg-catalog/src/api/endpoints.rs
```diff
@@ -0,0 +1,346 @@
...
+    }
+
+    pub fn from_method_and_matched_path(method: &Method, inp: &str) -> Option<Self> {
+        if inp.starts_with("/management/v1/permissions") {
+            return match method.as_str() {
+                "GET" => Some(Endpoints::ManagementGetPermissions),
+                "POST" => Some(Endpoints::ManagementPostPermissions),
+                "HEAD" => Some(Endpoints::ManagementHeadPermissions),
+                "DELETE" => Some(Endpoints::ManagementDeletePermissions),
+                _ => None,
```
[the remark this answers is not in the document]
twuebi:
```
                    let Some(uri) = Endpoints::from_method_and_matched_path(
                        request_metadata.request_method(),
                        matched_path,
                    ) else {
                        tracing::error!(
                            "Could not parse endpoint from matched path: '{} {}'. This is likely a bug which will affect the statistics collection.",
                            request_metadata.request_method(),
                            matched_path
                        );
                        continue;
                    };
```

we're logging it in the message receiver, we're also authenticated at this point so no bots or random calls should show up. We're also only logging matched-paths, so we'll only ever see routes that exist being called by authenticated users.

crates/iceberg-catalog/src/api/management/mod.rs
c-thiel: UI Issue: https://github.com/lakekeeper/console/issues/101

c-thiel (dismissed):

c-thiel (approved):

c-thiel:
Currently something is very noisy in the logs - doesn't seem right.

c-thiel (approved):

c-thiel (approved):

PR #819, opened 2025-02-13
feat: add kafka support for CloudEvents publishing #271

Grongrilla:
Attempt number 2 to add kafka support


brgr-s:
To start of trying to get this to work, I just merge main into my branch. No work done on trying to get rid of cross build, yet.

brgr-s:
@twuebi ... at long last?

.github/workflows/release.yml
```diff
@@ -135,6 +139,24 @@ jobs:
...
+          DOCKER_BUILDKIT=1 docker build -t localhost/build-lakekeeper:latest \
+            -f docker-build/build-with-alpine3.21.Dockerfile .
+
+      - name: Build [Docker Build]
+        if: matrix.docker-build == true
+        run: |
+          docker run \
+          --mount type=bind,src=.,target=/build \
+          localhost/build-lakekeeper:latest \
+          cargo build --release --all-features --locked --target ${{ matrix.target }}
```
twuebi, line 158: I tried executing this locally but it fails due to no Cargo.toml being found in /build, it did work with 

`-w /build --mount type=bind,src=$(pwd),target=/build `

Grongrilla: I will have a second look at this. I tested locally, then with a custom workflow, and basically just copy-pasted the latter into release.yml.

However, your suggestion is "more stable", I think we should go with that.

PR #821, opened 2025-02-16
feat: Add Opt-In to S3 Variant prefixes (s3a, s3n)

c-thiel:


crates/iceberg-catalog/migrations/20250216105917_variant_protocols.sql
```diff
@@ -0,0 +1,61 @@
...
+   FROM active_tabulars t
+  WHERE typ = 'view'::tabular_type;
+
+-- Drop old index - old column is going to be dropped after
+-- migration period.
+drop index tabular_namespace_id_location_idx;
+ALTER TABLE tabular drop column location;
+
+-- Create a new index on the new columns
+create index tabular_namespace_id_location_idx on tabular (namespace_id, fs_location);
```
twuebi, line 61: shouldn't the index also include the protocol? Maybe based on an expression which groups s3a/s3 together

c-thiel: Currently the lookup in the s3 signer does not provide the protocol. We would only need to include the protocol, if we assume that different protocols could correspond to different locations in the query. Because we don't have this scenario currently, I excluded the protocol both in the index and the query.

crates/iceberg-catalog/src/implementations/postgres/migrations/split_table_metadata.rs
```diff
@@ -233,178 +228,4 @@ mod test {
         }
         trx.commit().await.unwrap();
     }
-
-    #[sqlx::test(fixtures("metadata_migration"))]
```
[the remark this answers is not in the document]
c-thiel: I was thinking about the same. We already removed some of the tests in the past I believe.
I would personally prefer to keep the already tested function for another 1-2 minor releases before we remove it.

There is no easy way with SQLx to run this tests against a specific DB version?

crates/iceberg-catalog/src/service/storage/s3.rs
```diff
@@ -742,33 +807,48 @@ impl TryFrom<Location> for S3Location {
...
-impl From<S3Location> for Location {
-    fn from(location: S3Location) -> Self {
-        location.location
+    /// Always returns `s3://` prefixed location.
+    pub(crate) fn into_normalized_location(self) -> Location {
+        self.location
     }
 }
 
+// impl From<S3Location> for Location {
```
twuebi, line 844: remove the comment block?

tests/python/tests/conftest.py
```diff
@@ -504,6 +511,7 @@ def spark(warehouse: Warehouse, storage_config):
         f"spark.sql.catalog.{catalog_name}.scope": settings.openid_scope,
         f"spark.sql.catalog.{catalog_name}.oauth2-server-uri": f"{settings.token_endpoint}",
     }
+    print(f"Configuration: {configuration}")
```
twuebi, line 514: could there be credentials in configuration?

crates/iceberg-catalog/migrations/20250216105917_variant_protocols.sql
twuebi: hm, I was wondering about `s3://some/location` vs `abfss://some_location` but I guess `namespace_id` has us covered there

crates/iceberg-catalog/src/implementations/postgres/migrations/split_table_metadata.rs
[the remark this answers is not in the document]
twuebi: I meant the fixtures-migration at `crates/iceberg-catalog/src/implementations/postgres/migrations/fixtures`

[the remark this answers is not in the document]
c-thiel: Done

crates/iceberg-catalog/src/service/storage/s3.rs
c-thiel: Done

tests/python/tests/conftest.py
c-thiel: There could be - shouldn't be here

twuebi (dismissed):

twuebi (approved):

PR #829, opened 2025-02-19
feat(tasks): add unit to poll interval config

twuebi:
Improves testing ergonomics, being able to set poll interval in ms range means we can sleep much shorter in integration tests which is helpful for the upcoming trino integration tests


c-thiel (commented):
What do you think about using: https://crates.io/crates/iso8601
ISO 8601 timestamps are also being discussed in other places in the iceberg ecosystem.

c-thiel (dismissed):

c-thiel (approved):

PR #834, opened 2025-02-19
feat: use x-forwarded-for/host headers to generate links

twuebi:
makes trino tests a lot easier & also eliminates some problems users had that ran lakekeeper in docker and spark from host.


crates/iceberg-catalog/src/request_metadata.rs
```diff
@@ -11,13 +11,19 @@ use uuid::Uuid;
...
+const X_FORWARDED_PROTO_HEADER: &str = "x-forwarded-proto";
+const X_FORWARDED_PORT_HEADER: &str = "x-forwarded-port";
 
 /// A struct to hold metadata about a request.
 #[derive(Debug, Clone)]
 pub struct RequestMetadata {
     request_id: Uuid,
     project_id: Option<ProjectIdent>,
     authentication: Option<Authentication>,
+    host: Option<String>,
```
c-thiel, line 26: I think we should rename this everywhere to `base_url` or so - a host comes without protocol, just like the HOST header doesn't have the protocol prepended.
This would then also use the same naming as the config option that can be used to override the host.

crates/iceberg-catalog/src/request_metadata.rs
```diff
@@ -45,12 +51,14 @@ impl RequestMetadata {
...
     }
 
+    #[cfg(test)]
     #[must_use]
     pub fn new_unauthenticated() -> Self {
         Self {
             request_id: Uuid::now_v7(),
             project_id: None,
             authentication: None,
+            host: Some("localhost".to_string()),
```
c-thiel, line 61: The host as we currently use it has the protocol prepended, I would change this example to match a typical value: `http://localhost:8181`.

crates/iceberg-catalog/src/catalog/config.rs
```diff
@@ -60,16 +60,25 @@ impl<A: Authorizer + Clone, C: Catalog, S: SecretStore>
...
         config
             .overrides
             .insert("prefix".to_string(), CONFIG.warehouse_prefix(warehouse_id));
 
-        config
-            .overrides
-            .insert("uri".to_string(), CONFIG.base_uri_catalog().to_string());
+        config.overrides.insert(
+            "uri".to_string(),
+            request_metadata
```
c-thiel, line 76: We should switch the priority: Explicit config comes before Header parsing for a few reasons:
1. Backwards compatibility: This change could break pretty much all current deployments that are behind a proxy that is not configured to set the headers. `BASE_URL` is set correctly for those deployments. Who knows if headers are set correctly.
2. Deployments don't always have the proxy configuration in their hands, but they can configure Lakekeeper freely. Thus users need a way to force something on Lakekeeper side - our `BASE_URL` env var.

When changing this, we also need to change the docs.

docs/docs/configuration.md
```diff
@@ -6,15 +6,15 @@ For most deployments, we recommend to set at least the following variables: `LAK
...
+| Variable                                         | Example                                | Description                                                                                                                                                                                                                                                               |
+|--------------------------------------------------|----------------------------------------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| <nobr>`LAKEKEEPER__BASE_URI`</nobr>              | <nobr>`https://example.com:8181`<nobr> | Fallback base URL where the catalog is externally reachable. Default: `https://localhost:8181`. Lakekeeper uses x-forwarded-for & host before falling back to this option.                                                                                                |
```
c-thiel, line 11: Let's name all 3 variables that we respect explicitly. Helps for search :).

Lets also describe what we do if they are not set.

c-thiel (changes requested):

crates/iceberg-catalog/src/catalog/config.rs
twuebi: Then we'll either have to remove the default of base_url which is a breaking change, or we'll have to add a flag which disables the base_location default which also doesn't sound too pretty

PR #858, opened 2025-02-25
Add Front page to Lakekeeper site

bitsondatadev:
Adding Muna's first looks at the Lakekeeper landing page.

I did a quick overview of Muna's work (though mostly looking at the wording). I wanted to get the front page visible for you all to see and get Muna's CLA signed.


bitsondatadev:
@munabedan, you'll need to sign the CLA!

c-thiel:
@munabedan , @bitsondatadev  just had a quick peek - this looks awesome :)
I'll go over it in more detail in the next days and also aggregate Tobi's and Viktor's thoughts.

munabedan:
Thanks @c-thiel  , I look forward to your feedback 🙂

c-thiel:
@munabedan so, we gathered a bit of feedback. We really like the reactivity and everything, it's really good already. Just some nits that we could improve I think:

1. The front page image on the right: I really like having an image there, and I also like that the bear is kind of in the same fractal design as the logo. The bear however looks really sad with all its ice being melted away and only this very tiny iceberg remaining :D . It's OK if the image is simpler as well. It looks AI generated to me - a much simpler svg would also do it I think. It would be great if the Bear had our colors.
1. For the text, the K in Lakekeeper is lowercase
1. I am not sure if we can use the official rust logo. I remember that being quite restrictive. @bitsondatadev  not sure if you could maybe check permissions on that?
1. We like the headline font, but I think we are using it in too many places currently. It seems a little bit harder to read than the regular font. So it's great as an accent, but maybe we can reduce the overall usage a bit?

I'll work a bit on the texts and PR it into your PR. Probably next Thursday in the plane :)

c-thiel (approved):

c-thiel (approved):

PR #876, opened 2025-03-03
feat: Migrate to OpenFGA Client

c-thiel:


authz/openfga/v2/changed.diff
```diff
@@ -1,18 +0,0 @@
---- authz/openfga/v1/schema.fga	2024-12-09 12:10:28
```
twuebi, line 1: why get rid of this file? I thought it was nice to see the changes like this in PRs etc

crates/iceberg-catalog/src/service/authz/implementations/openfga/client.rs
```diff
@@ -83,79 +59,37 @@ pub(crate) async fn new_client_from_config(
...
+    let store_name = store_name.unwrap_or(AUTH_CONFIG.store_name.clone());
+    let auth_model_id = get_active_auth_model_id(&mut client, Some(store_name.clone())).await?;
+    let store = client
         .get_store_by_name(&store_name)
         .await?
-        .ok_or(OpenFGAError::StoreNotFound { store: store_name })?
-        .id;
-    let authorization_model_id =
-        get_auth_model_id(&mut client, store_id.clone(), active_model_version).await?;
+        .unwrap_or_else(|| panic!("OpenFGA store with name {store_name} exists"));
```
twuebi, line 86: do we want to panic here?

crates/iceberg-catalog/src/service/authz/implementations/openfga/error.rs
```diff
@@ -1,71 +1,47 @@
...
+    // InvalidBearerToken(InvalidMetadataValue),
+    // #[error("Store not found: {store}")]
+    // StoreNotFound { store: String },
+    // #[error("Too many authorization models in database. Max allowed pages: {0}")]
+    // TooManyAuthorizationModels(u32),
+    // #[error("Too many pages")]
+    // TooManyPages {
+    //     max_pages: u32,
+    //     tuple: ReadRequestTupleKey,
+    // },
```
twuebi, line 33: remove those?

crates/iceberg-catalog/src/service/authz/implementations/openfga/client.rs
c-thiel: I made an error from it. Probably cleaner.
Eventually I might pipe it though all the way from openfga_client

crates/iceberg-catalog/src/service/authz/implementations/openfga/error.rs
c-thiel: Gone :)

authz/openfga/v2/changed.diff
c-thiel: The principal is now a bit different:
All backwards compatible changes can be made right there in the `store.fga.yaml` and the folder version increased.
Increasing the folder version is just a failsafe to prevent us from not updating rust.

Only for breaking changes, that require tuple migrations, we will create a new folder.
Thus, for backwards compatible changes we can now simply use the git history to get changes.
Happy to add the diff file for new majors.

twuebi (approved):

PR #878, opened 2025-03-04
fix: /default-project route to match openapi def

xugy99:


c-thiel (dismissed):
LGTM! Will add this to 0.8.0 as this is a breaking change (although its right of course!)

twuebi (dismissed):

c-thiel (dismissed):

c-thiel (approved):

PR #880, opened 2025-03-04
feat!: Change Project ID type from UUID to String (alphanumeric + hyphen + underscore)

c-thiel:
To integrate better with external systems, we enlarge the ProjectID type from UUID to a String.
Project-IDs may only contain alphanumeric characters, hyphens and underscores

BEGIN_COMMIT_OVERRIDE
feat!: Change Project ID type from UUID to String (alphanumeric + hyphen + underscore)
feat!: Remove `project_id` from `RenameProjectRequest`. Use `/v1/project/{project_id}/rename` instead.
END_COMMIT_OVERRIDE


xugy99 (approved):
looks good to me 👍

crates/iceberg-catalog/src/implementations/postgres/tabular/table/mod.rs
```diff
@@ -128,77 +128,6 @@ impl From<DbTableFormatVersion> for FormatVersion {
     }
 }
 
-pub(crate) async fn load_tables_old(
```
twuebi, line 131: We should also remove `only_unmigrated: bool,` from `tabular::list_tabulars`

crates/iceberg-catalog/src/service/mod.rs
```diff
@@ -131,10 +139,55 @@ impl From<ProjectId> for uuid::Uuid {
...
+            Ok(Self(id))
+        } else {
+            Err(ErrorModel::builder()
+                .code(StatusCode::BAD_REQUEST.into())
+                .message(format!(
+                    "Project IDs may only contain alphanumeric characters, hyphens and underscores. Got: `{id}`",
+                ))
+                .r#type("MalformedProjectID".to_string())
+                .build()
+                .into())
```
twuebi, line 169: Let's use the variant constructor, eventually, I'd like us to get rid of the builder altogether

docs/docs/api/management-open-api.yaml
```diff
@@ -5,7 +5,7 @@ info:
   license:
     name: Apache-2.0
     identifier: Apache-2.0
-  version: 0.6.2
```
twuebi, line 8: Can we somehow make release-plz do this for us?

crates/iceberg-catalog/src/implementations/postgres/tabular/table/mod.rs
c-thiel: Done, I also removed the column from `tabulars`

twuebi (approved):

PR #881, opened 2025-03-04
feat: Split OpenFGA Model into components

c-thiel:


crates/iceberg-catalog/src/service/authz/implementations/openfga/mod.rs
```diff
@@ -1,11 +1,14 @@
-use std::{collections::HashSet, fmt::Debug, sync::Arc};
+#![allow(clippy::ref_option)] // For lazy static
```
twuebi, line 1: maybe move this one just above the offending line instead of disabling it for the whole mod?

twuebi (approved):

PR #891, opened 2025-03-11
feat: recursive namespace deletes

twuebi:
Adds an endpoint `/management/v1/warehouse/{warehouse_id}/namespace/{namespace_id}?force={force}&purge={purge}` which deletes a namespace and everything it contained. Something like `rm -r` would do.

If the namespace is within a warehouse that has soft-deletions active, the `force` query parameter has to be `true` otherwise a 400 Bad Request is returned. The `purge` query parameter controls whether the catalog should just forget about the table or if it should also clean up any data in the respective object store.


crates/iceberg-catalog/src/api/management/mod.rs
```diff
@@ -837,6 +838,49 @@ pub mod v1 {
...
+    /// namespace will be deleted. The namespace itself will also be deleted. If the warehouse
+    /// containing the namespace is configured with a soft-deletion profile, the `force` flag has to
+    /// be provided. The deletion will not be a soft-deletion. Every table, view and namespace will
+    /// be gone as soon as this call returns. Depending on whether the `purge` flag was set to true,
+    /// the data will be queued for deletion too. Any pending `tabular_expiration` will be cancelled.
+    /// If there's a running `tabular_expiration`, this call will fail with a `409 Conflict` error.
+    #[utoipa::path(
+        delete,
+        tag = "warehouse",
+        path = "/management/v1/warehouse/{warehouse_id}/namespace/{namespace_id}",
```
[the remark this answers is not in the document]
c-thiel: 1) would be very similar to `rm`. You can have `rm`, `rm -r`, 'rm -rf'

crates/iceberg-catalog/src/implementations/postgres/namespace.rs
```diff
@@ -685,13 +761,50 @@ pub(crate) mod tests {
...
+
+        let mut transaction = PostgresTransaction::begin_write(state.clone())
+            .await
+            .unwrap();
+        let namespace_id =
+            namespace_to_id(warehouse_id, &table.namespace, transaction.transaction())
+                .await
+                .unwrap()
+                .expect("Namespace not found");
+        drop_namespace(warehouse_id, namespace_id, true, transaction.transaction())
```
c-thiel, line 787: Should we also assert that the returned drop info is what we expect?

crates/iceberg-catalog/src/implementations/postgres/namespace.rs
```diff
@@ -709,21 +822,63 @@ pub(crate) mod tests {
...
 
-        let result = drop_namespace(warehouse_id, response.0, transaction.transaction())
+        let result = drop_namespace(warehouse_id, response.0, false, transaction.transaction())
             .await
             .unwrap_err();
 
         assert_eq!(result.error.code, StatusCode::CONFLICT);
 
-        drop_namespace(warehouse_id, response2.0, transaction.transaction())
+        drop_namespace(warehouse_id, response2.0, false, transaction.transaction())
```
c-thiel, line 831: Also here and below, let's assert the return.

crates/iceberg-catalog/migrations/20250311111248_cascade_delete.sql
```diff
@@ -0,0 +1,16 @@
+-- TODO: should these become an enum?
```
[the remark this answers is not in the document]
twuebi: Was thinking if this could have more states than on & off which is a bit harder to extend when not an enum

twuebi:
Protected:

image

image

docs/docs/api/rest-catalog-open-api.yaml
```diff
@@ -3258,6 +3353,12 @@ components:
           nullable: true
           items:
             type: string
+        protection-status:
+          type: array
+          uniqueItems: true
```
c-thiel, line 3358: Items are not unique - its an array of booleans

docs/docs/api/rest-catalog-open-api.yaml
```diff
@@ -3274,6 +3375,12 @@ components:
...
```
c-thiel, line 3380: Items are not unique - its an array of booleans

docs/docs/api/rest-catalog-open-api.yaml
```diff
@@ -544,6 +602,16 @@ paths:
 
 
 
+
```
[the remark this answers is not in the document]
twuebi: somehow yq inserts those every time i run it

crates/iceberg-catalog/src/catalog/tables.rs
```diff
@@ -894,6 +907,43 @@ impl<C: Catalog, A: Authorizer + Clone, S: SecretStore>
...
+                &request_metadata,
+                warehouse_id,
+                CatalogWarehouseAction::CanUse,
+            )
+            .await?;
+        authorizer
+            .require_table_action(
+                &request_metadata,
+                Ok(Some(table_id)),
+                CatalogTableAction::CanCommit,
```
c-thiel, line 933: We should also use `CanDelete` here. Otherwise people can set protection but not unset it.

crates/iceberg-catalog/src/catalog/tables.rs
```diff
@@ -894,6 +907,43 @@ impl<C: Catalog, A: Authorizer + Clone, S: SecretStore>
...
+            )
+            .await
+            .map_err(set_not_found_status_code)?;
```
c-thiel, line 936: What is the motivation of this mapping?
If someone has read access on the table and tries to get protection, and sees a 404 - this is unexpected.
I think we mostly return 401s currently if an object does not exist, unless its in the Catalog API because of the way Java handles client codes.

c-thiel (dismissed):

c-thiel (dismissed):

c-thiel (approved):

PR #905, opened 2025-03-14
feat(s3): make url-style detection configurable

twuebi:


crates/iceberg-catalog/src/catalog/s3_signer/sign.rs
```diff
@@ -180,6 +184,45 @@ impl<C: Catalog, A: Authorizer + Clone, S: SecretStore>
     }
 }
 
+async fn s3_url_style_detection<C: Catalog>(
+    state: C::State,
+    warehouse_id: WarehouseIdent,
+) -> Result<S3UrlStyleDetectionMode, IcebergErrorResponse> {
+    let mut tx = C::Transaction::begin_read(state).await?;
```
c-thiel, line 191: Is there a benefit to starting the transaction here instead of having it as part of the async block?
Beginning a transaction contacts the DB which we now do even if the cache hits.

crates/iceberg-catalog/src/service/storage/s3.rs
```diff
@@ -78,6 +79,37 @@ pub struct S3Profile {
...
+    /// Virtual host:
+    ///   - <https://foo.s3.endpoint.com/bar/a/key>
+    ///   - <https://foo.s3-eu-central-1.amazonaws.com/file>
+    ///
+    /// Path style:
+    ///   - <https://s3.endpoint.com/foo/bar/a/key>
+    ///   - <https://s3.us-east-1.amazonaws.com/foo/file>
+    #[serde(default)]
+    #[builder(default, setter(strip_option))]
+    pub s3_url_detection_mode: Option<S3UrlStyleDetectionMode>,
```
c-thiel, line 100: If we have a default on Auto, why do we need the Option here?

crates/iceberg-catalog/src/catalog/s3_signer/sign.rs
twuebi: no, good catch

c-thiel (approved):

c-thiel (approved):

PR #912, opened 2025-03-17
feat(az): support azure shared keys

### LINKED ISSUE #906: feat(az): Shared key authentication
Currently, we only support client credentials, some orgs forbid them and require use of shared access keys. We should add support for that.

reqsign & opendal seem to already support this, we'd have to add it to our iceberg-rust fork. Maybe we'll also have to update our opendal fork in the process.

twuebi:
closes #906


c-thiel (approved):
LGTM! I think we should add the UI change as part of this PR as well?

PR #937, opened 2025-03-24
feat: add kafka support #271

Grongrilla:


brgr-s:
@twuebi I added a `#[allow(dead_code)]` in `io.rs` line 225 for `test_remove_all`. This does not really belong in this PR, I just added it to get the unittests workflow running. Happy to remove it as soon as a fix for this problem is in `main`.

brgr-s:
~~@twuebi Needed to convert this back to draft.~~

~~Cmake v4.0 released a few days ago. It deprecates CMake < 3.5. librdkafka sets a minimum required version of CMake (3.2) without setting a max version (which is also deprecated). On CMake 4.0 this leads to the configuration(!) of the librdkafka build failing .
I need some time to make sure that in the build env CMake never gets updated to 4.x until librdkafka fixes their CMakeList.txt. Unfortunately we cannot get rid of CMake and build with mklove instead, as this breaks the -gnu build.~~

~~librdkafka maintainers should be aware of this issue: https://github.com/confluentinc/librdkafka/issues/4594~~

twuebi (dismissed):

twuebi (approved):

PR #952, opened 2025-03-28
fix!: drop duplicate sequence numbers in table snapshots & add index

twuebi:
Caution: this is a destructive change. Currently, it's possible through concurrent writes to get duplicate `table_id, sequence_number` tuples into `table_snapshot` by:

1. node A: load table
2. node B: load table
3. node A: add snapshot with sequence number incremented by one
4. node B: add snapshot with sequence number incremented by one
5. node A: commit
6. node B: commit

In general, this should not have happened often, since it's only possible when not sending a `RefSnapshotIdMatch` requirement:

```rust
#[serde(rename = "assert-ref-snapshot-id")]
    RefSnapshotIdMatch {
        /// The reference of the table to assert.
        r#ref: String,
        /// The snapshot id of the table to assert.
        /// If the id is `None`, the ref must not already exist.
        #[serde(rename = "snapshot-id")]
        snapshot_id: Option<i64>,
    },
```

### DIFF
--- a/crates/iceberg-catalog/migrations/20250328131139_fix_sequence_number.sql
+++ b/crates/iceberg-catalog/migrations/20250328131139_fix_sequence_number.sql
@@ -1,9 +1,25 @@
 DELETE
 FROM table_snapshot t1
     USING table_snapshot t2
-WHERE t1.created_at > t2.created_at
+        JOIN "table" ON t2.table_id = "table".table_id
+WHERE t1.created_at < t2.created_at
   AND t1.table_id = t2.table_id
-  AND t1.sequence_number = t2.sequence_number;
+  AND t1.sequence_number = t2.sequence_number
+  AND "table".table_format_version != '1';
 
-alter table table_snapshot
-    add constraint unique_table_snapshot_sequence_number unique (table_id, sequence_number);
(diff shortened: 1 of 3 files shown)


crates/iceberg-catalog/migrations/20250328131139_fix_sequence_number.sql
```diff
@@ -0,0 +1,25 @@
...
+FROM table_snapshot t1
+    USING table_snapshot t2
+        JOIN "table" ON t2.table_id = "table".table_id
+WHERE t1.created_at < t2.created_at
+  AND t1.table_id = t2.table_id
+  AND t1.sequence_number = t2.sequence_number
+  AND "table".table_format_version != '1';
+
+-- table fmt version 1 has no sequence number, hence we only enforce the unique constraint via a partial unique index
+CREATE OR REPLACE FUNCTION is_table_format_v2(table_id_param uuid)
```
c-thiel, line 11:
```suggestion
CREATE OR REPLACE FUNCTION is_not_table_format_v1(table_id_param uuid)
```

The function does not what the body describes - it would also return true for table_format_version 3.
I know its an enum, but I would surely forget to update the function.

c-thiel (approved):

PR #956, opened 2025-04-01
fix: table ref deletion

### LINKED ISSUE #954: Can't drop Tag/Branch permanently, it automatically reappear again after time
### Describe the bug
I'm encountering unexpected behavior with Tag/Branch functionality.

First, I create a new branch using the latest version of my table. Then, I attempt to drop the branch using the command `ALTER TABLE DROP BRANCH ...`, and this operation is successful.

As expected, if I immediately try to query the deleted branch/tag, it raises an error. However, the strange behavior is that after a few minutes (approximately three minutes in my case), I can query the deleted branch/tag normally again.

Furthermore, when I check the metadata in the Lakekeeper UI, the deleted branch is still present.

I'm unsure what's happening, but it seems the `ALTER TABLE DROP BRANCH` command isn't deleting the reference branch/tag in the new metadata file, which is why it becomes queryable again after a delay.(*I captured screenshot after dropping tab/branch*)
Image

**BOTH Branch and Tag have same issue.**

```python
import time
spark.sql("ALTER TABLE lakekeeper.risk_mon_out.test_data_sample CREATE TAG `historial-tag-20250331`")

spark.sql("select * from lakekeeper.risk_mon_out.test_data_sample.refs where type ='TAG'").show(truncate=False)

try:
    spark.sql("ALTER TABLE lakekeeper.risk_mon_out.test_data_sample DROP TAG `historial-tag-20250331`")
    print("Drop tag successfully")
except Exception as e:
    print(e)
# Wait 180s
time.sleep(180)
spark.sql("select * from lakekeeper.risk_mon_out.test_data_sample.`tag_historial-tag-20250331`").show(3)


try:
    spark.sql("ALTER TABLE lakekeeper.risk_mon_out.test_data_sample CREATE BRANCH `test-branch`")
    print("Created")
    spark.sql("ALTER TABLE lakekeeper.risk_mon_out.test_data_sample DROP BRANCH `test-branch`")
    print("Drop branch successfully")
except Exception as e:
    print(e)
time.sleep(180)
spark.sql("select * from lakekeeper.risk_mon_out.test_data_sample.refs").show(truncate=False)


```
How can I fix that?

### To Reproduce
Steps to reproduce the behavior, preferably starting from one of our docker-compose examples.
1. Create branch/tag
2. Test drop and wait
3. Try to select the deleted branch/tag

### Expected behavior
The deleted branch/tag can't query any more.

### Screenshots
- Drop tag, wait and still able to query after waiting time, but will raises error if query immediately
Image
- Metadata file still appear deleted tag/branch
Image

### System Information:
- Deployment Type: Helm
- Lakekeeper Version: 0.7.4
- Authentication Provider: Keycloak
- Authorization Provider: OpenFGA
- OS: Ubuntu
- Spark 3.5.2
- Iceberg 1.7.1

twuebi:
fixes #954


crates/iceberg-catalog/src/implementations/postgres/tabular/table/common.rs
```diff
@@ -406,8 +406,7 @@ pub(super) async fn insert_snapshot_refs(
     let _ = sqlx::query!(
         r#"
         WITH deleted AS (
-            DELETE FROM table_refs
-            WHERE table_id = $1 AND table_ref_name = ANY($2::TEXT[])
```
c-thiel, line 410: Shouldn't we better do `WHERE table_id = $1 AND table_ref_name not in ...`?
This would currently always re-write all refs. Refs are updated in almost every commit, we would then need to touch multiple rows. Also the last_updated_timestamp always gets updated now for all references.

twuebi: yes, much better, thanks!

c-thiel (approved):

PR #975, opened 2025-04-05
fix: Consistency checks for READ Committed transaction level

c-thiel:


crates/iceberg-catalog/src/catalog/tables.rs
```diff
@@ -68,6 +68,9 @@ const PROPERTY_METADATA_DELETE_AFTER_COMMIT_ENABLED: &str =
     "write.metadata.delete-after-commit.enabled";
 const PROPERTY_METADATA_DELETE_AFTER_COMMIT_ENABLED_DEFAULT: bool = false;
 
+pub(crate) const CONCURRENT_UPDATE_ERROR_TYPE: &str = "ConcurrentUpdateError";
+const MAX_RETRIES_ON_CONCURRENT_UPDATE: usize = 2;
```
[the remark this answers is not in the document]
c-thiel: Lets stick to in-code now, I think we might use this to Warehouse configs in the future if it proofs to be helpful

crates/iceberg-catalog/src/catalog/views/commit.rs
```diff
@@ -81,56 +86,137 @@ pub(crate) async fn commit_view<C: Catalog, A: Authorizer + Clone, S: SecretStor
...
 
+    // Verify assertions (only needed once)
     check_asserts(requirements.as_ref(), view_id)?;
 
+    // Prepare for event publishing (serialize body once)
+    let body = maybe_body_to_json(&request);
+
+    // Start the retry loop
+    let mut attempt = 0;
+    loop {
```
twuebi, line 103: I'm wondering if we can reduce nesting here a bit if we'd use tryhard which we already use within the io module

twuebi: (same at the table loop)

twuebi (approved):

PR #977, opened 2025-04-07
fix(stats): unique constraint violation on warehouse deletion

twuebi:


c-thiel:
On Warehouse Statistics (tables, views) we have `ON DELETE CASCADE`. Let's discuss which behavior we want. I think it should be consistent.

c-thiel (approved):

PR #1004, opened 2025-04-11
Additional notes on machine user provisioning for EntraID OAuth2

george-zubrienko:


c-thiel:
Thank you very much for the PR @george-zubrienko !

PR #1046, opened 2025-04-24
fix: Drop Warehouse should not work with open tasks

c-thiel:


c-thiel:
Closed in favor of https://github.com/lakekeeper/lakekeeper/pull/1052

PR #1059, opened 2025-04-30
feat: configurable warehouse validation

twuebi:


crates/iceberg-catalog/src/service/storage/mod.rs
c-thiel, line 1: I think there are a few problems here:
1. Trace in line 347 might be wrong, let's updated it with what actually happens.
2. I don't think it works, because cleanup always runs (line 391) - did I overlook something? Did you test it with read-only credentials?

c-thiel:
Blocker: Information about read-only needs to be persistet in DB. Its not just the validation that needs to be skipped.
Will be handled by Federation eventually.

PR #1075, opened 2025-05-06
feat: add endpoint hooks to inject plugins from main

twuebi:


crates/iceberg-catalog-bin/src/serve.rs
```diff
@@ -120,18 +121,19 @@ pub(crate) async fn serve(bind_addr: std::net::SocketAddr) -> Result<(), anyhow:
             CONFIG.queue_config.clone(),
         )?),
     );
+    let hooks = Hooks::new(vec![]);
```
c-thiel, line 124: Let's give the class a more expressive name like "EndpointHooks". Maybe copilot has a good idea as well.

crates/iceberg-catalog/src/service/hooks.rs
```diff
@@ -0,0 +1,280 @@
...
+    pub fn new(hooks: Vec<Arc<dyn EndpointHooks>>) -> Self {
+        Self(hooks)
+    }
+}
+
+#[async_trait::async_trait]
+impl EndpointHooks for Hooks {
+    async fn commit_table(
+        &self,
+        _warehouse_id: WarehouseIdent,
```
c-thiel, line 41: Why are we using underscores here when the parameter is used below?

crates/iceberg-catalog/src/service/hooks.rs
```diff
@@ -0,0 +1,280 @@
...
+        _prefix: Option<Prefix>,
+        _request: Arc<CommitTransactionRequest>,
+        _table_ident_map: Arc<HashMap<TableIdent, TableIdentUuid>>,
+        _request_metadata: Arc<RequestMetadata>,
+    ) {
+        for hook in &self.0 {
```
c-thiel, line 47: Its already a for loop, can we just join the futures and run them in parallel?

crates/iceberg-catalog/src/service/hooks.rs
```diff
@@ -0,0 +1,280 @@
...
+
+#[async_trait::async_trait]
+pub trait EndpointHooks: Send + Sync {
+    async fn commit_table(
+        &self,
+        _warehouse_id: WarehouseIdent,
+        _prefix: Option<Prefix>,
+        _request: Arc<CommitTransactionRequest>,
+        _table_ident_map: Arc<HashMap<TableIdent, TableIdentUuid>>,
+        _request_metadata: Arc<RequestMetadata>,
```
c-thiel, line 192: We should also add the actual resulting metadata - as this contains information required for the decision if for example snapshot expiration is required.
Also for all other endpoints where we have it in memory anyway.

crates/iceberg-catalog/src/service/hooks.rs
```diff
@@ -0,0 +1,280 @@
...
+    ) {
+        // Default implementation does nothing
+    }
+
+    #[allow(clippy::too_many_arguments)]
+    async fn commit_view(
+        &self,
+        _warehouse_id: WarehouseIdent,
+        _view_ident_uuid: ViewIdentUuid,
```
c-thiel, line 252: Lets called all `xxx_ident_uuid` parameters simply `xxx_id`, i.e. `view_id`.
I already fixed project_id globally, but not yet the view / table ids. Then we just have:
`xxx_ident`: Iceberg identifier
`xxx_id`: Unique id

crates/iceberg-catalog/src/service/hooks.rs
```diff
@@ -0,0 +1,280 @@
...
+        }
+    }
+}
+
+#[async_trait::async_trait]
+pub trait EndpointHooks: Send + Sync {
+    async fn commit_table(
+        &self,
+        _warehouse_id: WarehouseIdent,
+        _prefix: Option<Prefix>,
```
c-thiel, line 189: Why do we have the prefix in some endpoints but not all?
My tendency is to remove it everywhere and just use `warehouse_id` as an arugment everywhere.

crates/iceberg-catalog/src/service/hooks.rs
twuebi: courtesy of intellij.. clippy complains about unused vars, refactor -> rename apparently ends up renaming it across trait impls as well

crates/iceberg-catalog/src/service/endpoint_hooks.rs
```diff
@@ -0,0 +1,354 @@
...
+        Self(hooks)
+    }
+}
+
+impl Display for EndpointHookCollection {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "EndpointHookCollection with [")?;
+        for idx in 0..self.0.len() {
+            if idx == self.0.len() - 1 {
+                write!(f, "{:?}", self.0[idx])?;
```
c-thiel, line 51: Shouldn't this use Dispalay as well?

crates/iceberg-catalog/src/service/endpoint_hooks.rs
```diff
@@ -0,0 +1,354 @@
...
+/// The `EndpointHooks` are passed into the services via the `EndpointHookCollection`. If you want
+/// to provide your own implementation, you'll have to fork and modify the main function to include
+/// your hooks.
+#[async_trait::async_trait]
+pub trait EndpointHooks: Send + Sync + Debug + Display {
+    async fn commit_transaction(
+        &self,
+        _warehouse_id: WarehouseId,
+        _request: Arc<CommitTransactionRequest>,
+        _responses: Arc<Vec<CommitContext>>,
```
c-thiel, line 259: This isn't quite the response of the endpoint.
Maybe: ` _commits: Arc<Vec<CommitContext>>,`

crates/iceberg-catalog/src/service/endpoint_hooks.rs
```diff
@@ -0,0 +1,354 @@
...
+    async fn register_table(
+        &self,
+        _warehouse_id: WarehouseId,
+        _parameters: NamespaceParameters,
+        _request: Arc<RegisterTableRequest>,
+        _metadata: Arc<TableMetadata>,
+        _request_metadata: Arc<RequestMetadata>,
+    ) {
+    }
+    async fn create_table(
```
c-thiel, line 284: We should include metadata location for most endpoints, including the old metadata location if it changed. We already have this for the `commit_transaction` endpoint, but not for views.

crates/iceberg-catalog/src/service/endpoint_hooks.rs
```diff
@@ -0,0 +1,354 @@
...
+        _table_ident_map: Arc<HashMap<TableIdent, TableId>>,
+        _request_metadata: Arc<RequestMetadata>,
+    ) {
+        // Default implementation does nothing
+    }
+
+    async fn drop_table(
+        &self,
+        _warehouse_id: WarehouseId,
+        _parameters: TableParameters,
```
c-thiel, line 269: We should not expose any type here that contains the prefix, such as `TableParameters` or `NamespaceParameters`. The identifier is enough. Implementations of Hooks must never rely on what is in the prefix, as this is determined by Lakekeeper itself. So Lakekeeper is also responsible to parse the prefix and just forward usable information (such as the `warehouse_id`) to clients.

docs/docs/api/management-open-api.yaml
```diff
@@ -4134,6 +4134,32 @@ components:
               type: string
               enum:
                 - soft
+    TabularId:
```
c-thiel, line 4137: This should not change - it should still be `TabularIdentOrUuid`

crates/iceberg-catalog/src/service/endpoint_hooks.rs
twuebi: We don't offer API endpoints to do lookups via identifier, do you plan to expose these externally?

c-thiel (dismissed):

c-thiel (approved):

c-thiel (approved):

Issue #1051, opened 2025-04-28
Implement MCP Server


c-thiel:

brgr-s:
@c-thiel 

I had a look at the [official Rust SDK](https://github.com/modelcontextprotocol/rust-sdk/) and [a popular alternative](https://github.com/rust-mcp-stack/rust-mcp-sdk).

Both offer a "MCP core", which is basically the[ JSON RPC spec for MCP](https://modelcontextprotocol.io/specification/2025-06-18/basic#schema) translated into `trait`s and/or `struct`s, along with some helpers.

On top of that, both SDKs offer a "batteries included" implementation of the [MCP lifecycle spec](https://modelcontextprotocol.io/specification/2025-06-18/basic/lifecycle), [MCP transport spec](https://modelcontextprotocol.io/specification/2025-06-18/basic/transports) and [MCP authZ spec](https://modelcontextprotocol.io/specification/2025-06-18/basic/authorization), with the possibility to implement clients and serves.

That "batteries included" approach makes it very hard (maybe impossible) to integrate lifecycle and transport (authentication we wouldn't need) into an existing, axum based API that requires a shared state - like Lakekeeper. For the official SDK there is even an [issue](https://github.com/modelcontextprotocol/rust-sdk/issues/352) asking for that. Every abstraction provided in the SDKs that might help is unfortunately so "thight", that we cannot simply add state.

So in order to implement an "MCP endpoint" (there is only one, the rest is handled via JSON RPC) we can only use "MCP core" and have to implement everything ourselves. Of course there might be crates out there that help with lifecycle and - if we want to - support for streaming and/or SSE.

Two other alternatives come to mind:

1) Proxy Catalog (and Management) API calls, i.e. call ourselves via HTTP for tool/resource requests from inside lakekeeper (which is what the OP in above mentioned issue is doing)
2) Implement a "Lakekeeper-Aware Iceberg REST MCP Server" using the official SDK or some other SDK we like
3) Wait for improvements of official SDK or another SDK regarding axums' shared state


The _next_ thing to consider would be what kind of tools or resources to actually implement. The [linked example](https://github.com/ryft-io/iceberg-mcp) implements a handfull of tools, where I am not even sure, if those are

1) usefull to anyone
2) really MCP tools, and not MCP resources (tools are expected to "do something" like "Send an email", "Make a reservation", "Schedule a meeting")

[...]

PR #1078, opened 2025-05-07
feature: mcp server scaffolding

HerrMuellerluedenscheid:
Implementing #1051
Opening the merge request early for transparency and to discuss questions.

### DIFF
--- a/crates/iceberg-catalog/src/api/endpoints.rs
+++ b/crates/iceberg-catalog/src/api/endpoints.rs
@@ -227,6 +227,10 @@ generate_endpoints! {
         Delete(DELETE, "/management/v1/permissions"),
         Put(PUT, "/management/v1/permissions"),
     }
+
+    enum MCPV1 {
+        ServerInfo(GET, "/mcp/v1/info"),
+    }
 }
 
 impl Endpoint {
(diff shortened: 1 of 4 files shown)


c-thiel:
Thank you for taking this!
If you need Feedback or anything or just someone to exchange thoughts, just let me know. Happy to jump in a call as well :)

HerrMuellerluedenscheid:
Thanks so much @c-thiel ! Yeah, maybe a call might be helpful. I'll shoot you a message on discord.
I was mostly reading APIs and docs today, familiarizing with the tech stack at use. I found the [tower handler](https://github.com/modelcontextprotocol/rust-sdk/blob/09474d47694225dc7cf828776d934fbd02414cba/crates/rmcp/src/service/tower.rs) which might be the simplest way to extend the existing axum server. But that might be something worthy to discuss :)

HerrMuellerluedenscheid:
After wrangling a bit with the `mrcp` crate I think that [this example](https://github.com/HerrMuellerluedenscheid/axum-mcp/blob/main/src/main.rs) is comprised of the functionality that we need: A stateful mcp tool and a working router. It requires making the handlers of the protocol at use public for now. But that is something that I think we can live with for the time being.

PR #1085, opened 2025-05-08
feat(tasks): Extendable Task Queue System

### LINKED ISSUE #748: feat(tasks): add heart beat
Currently, we rely on max-age to pick up stale tasks. For our current tasks, this is probably fine. For longer running tasks, this may be problematic, we should add a heart beat mechanism which periodically informs the world that it's still alive.

Possible solutions:

- add `last_heartbeat` + `runner_id` columns to `task` & update periodically from runner code
- acquire an advisory lock for the task which would be dropped if the e.g. instance is restarted

### LINKED ISSUE #807: feat(tasks): make task configuration configurable at runtime per queue
We currently have a single global queue config with the following config options:
* `max_retries`
* `max_age`
* `poll_interval`

I believe that at least `max_age` should be quite different between queues - compaction should be significantly higher than TabularExpiration, which should probably time out after a minute already.

I am thinking about making individual task queues configurable via API, so that we can offer those settings via the UI. This would make configuration much easier compared to Env-vars and would be easier extandable to add queue specific configurations.

_Originally posted by @c-thiel in https://github.com/lakekeeper/lakekeeper/pull/763#discussion_r1948053048_

### LINKED ISSUE #810: feat(tasks): cancel running
Eventually, we want to be able to cancel running tasks. This is a somewhat difficult undertaking. Currently, we don't expose on which tasks are running on a specific Lakekeeper instance. Exposing this would require a refactoring of our task runners so that they have a handle to each running task.

Once that exists, we need to think how running tasks can be cancelled, currently, tasks are just blackboxes and their internal state is unclear so cancellation may, depending on task, leave the system in an unknown state.

twuebi:
initial work to support external task queues

closes #748 
closes #807 
closes #810


crates/iceberg-catalog/migrations/20250508101407_tasks_store_their_state.sql
```diff
@@ -0,0 +1,97 @@
...
+(
+    task_id            uuid primary key                                           not null,
+    warehouse_id       uuid references warehouse (warehouse_id) on delete cascade not null,
+    queue_name         text                                                       not null,
+    state              jsonb                                                      not null,
+    status             task_outcomes                                              not null,
+    entity_id          uuid                                                       not null,
+    entity_type        entity_type                                                not null,
+    last_error_details text
+);
```
c-thiel, line 57: Timestamps are missing

crates/iceberg-catalog/src/implementations/postgres/task_queues.rs
```diff
@@ -0,0 +1,1114 @@
...
+    }
+
+    Ok(())
+}
+
+use crate::service::task_queue::{EntityId, TaskInput, TaskMetadata};
+
+/// Cancel pending tasks for a warehouse
+/// If `task_ids` are provided in `filter` which are not pending, they are ignored
+pub(crate) async fn cancel_pending_tasks(
```
c-thiel, line 321: Shouldn't we also cancel running tasks via this function?
My idea is (doesn't need to be this PR):
* User requests cancellation (are there other reasons to cancel currently?) -> set status `cancellation_requested`
* Running task can periodically check its own status via a trait method. If it sees `cancellation_requested` state, it cancels its run and `record_cancellation`
What do you think?
If the task doesn't check cancellation, then the task will stay in `cancellation_requested` until max age is reached.

crates/iceberg-catalog/src/implementations/postgres/task_queues.rs
```diff
@@ -0,0 +1,1114 @@
...
+    }
+}
+
+#[async_trait::async_trait]
+impl TaskQueue for PgQueue {
+    async fn pick_new_task(&self, queue_name: &str) -> crate::api::Result<Option<Task>> {
+        pick_task(&self.read_write.write_pool, queue_name, &self.max_age).await
+    }
+
+    async fn record_success(&self, id: Uuid, message: Option<&str>) -> crate::api::Result<()> {
```
c-thiel, line 47: Lets name the arguments `task_id` instead of just id to avoid confusion with entity_id.
Should we introduce a `TaskId` type?

crates/iceberg-catalog/src/service/task_queue/mod.rs
```diff
@@ -1,132 +1,219 @@
...
+        loop {
+            for (queue_name, queue_tx) in &self.queues {
+                match self.queue.pick_new_task(queue_name).await {
+                    Ok(Some(task)) => {
+                        if let Err(e) = tryhard::retry_fn(|| queue_tx.send(task.clone()))
+                            .retries(5)
+                            .fixed_backoff(Duration::from_millis(100))
+                            .await
+                        {
+                            // TODO: Do we even want to record_failure for this?
```
c-thiel, line 86: I think we should just log an error.
The task might be important. We don't want to progress its attempts and remove it from the active tasks table if we haven't actually tried to run it due to internal problems.

crates/iceberg-catalog/src/service/task_queue/mod.rs
```diff
@@ -1,132 +1,219 @@
...
+                    Ok(None) => {
+                        tracing::trace!("No task available for queue '{queue_name}'");
+                        // No task available
+                    }
+                    Err(e) => {
+                        tracing::error!("Failed to pick new task: '{}'", e.error);
+                    }
+                }
+            }
+            tokio::time::sleep(poll_interval).await;
```
c-thiel, line 104: This is re-introducing the problem we fixed in https://github.com/lakekeeper/lakekeeper/pull/1025.
If there are many jobs waiting to be polled, and the poll_interval is significantly larger than the capacity (tasks / second basically) of the worker, then we starve the worker, simply because we don't poll often enough.
What the sleep here should do is prevent unneccessary DB load if there is nothing to fetch.

I would propose something like:
* Run in parallel, one loop per task type
* If a poll was successful, immediately get the next task if the local queue contains less ten `number_of_queue_workers  + 1` items.
* If a poll retured no task or polling errored, then sleep

I guess there is also no need of polling if the local queue is too full anyway. This is also solved by the approach I described.
Currently, lets say one worker is completly occupied with running compaction for a day - the queue is still filled up with more and more tasks every `poll_interval`. Other workers might finish but don't get any tasks until they reach their `max_age`.

crates/iceberg-catalog/src/service/task_queue/mod.rs
```diff
@@ -1,132 +1,219 @@
...
+            }
+            self.queues.insert(name, tx);
+        }
+        let feeder = tokio::task::spawn(self.run(poll_interval));
+        tokio::select! {
+            res = futures::future::select_all(queue_tasks) => {
+                let (res, index, _) = res;
+                if let Err(e) = res {
+                    tracing::error!("Task queue {index} panicked: {e}");
+                    return Err(anyhow::anyhow!("Task queue {index} panicked: {e}"));
```
c-thiel, line 138: With the display implementation of e we might loose information. Lets try to keep the full error it somewhere for the user.

crates/iceberg-catalog-bin/src/serve.rs
```diff
@@ -414,7 +385,7 @@ async fn serve_inner<A: Authorizer, N: Authenticator + 'static>(
     let stats_handle = tokio::task::spawn(tracker.run());
 
     tokio::select!(
-        _ = queues.spawn_queues::<PostgresCatalog, _, _>(catalog_state, secrets_state, authorizer) => tracing::error!("Tabular queue task failed"),
+        _ = queues.spawn_queues::<PostgresCatalog, _, _>(catalog_state, secrets_state, authorizer, CONFIG.queue_config.poll_interval) => tracing::error!("Tabular queue task failed"),
```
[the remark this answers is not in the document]
twuebi: it helps a lot with testing instrumentation, I'd like to avoid having to put a figment jail closure around every test that needs to set the polling interval to a small value

crates/iceberg-catalog/migrations/20250508101407_tasks_store_their_state.sql
```diff
@@ -0,0 +1,112 @@
+create type entity_type as enum ('tabular');
+create type task_final_status as enum ('failure', 'cancellation', 'success');
```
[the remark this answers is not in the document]
twuebi: they are proper nouns describing an end result, we can go back to adjectives.

crates/iceberg-catalog/migrations/20250508101407_tasks_store_their_state.sql
```diff
@@ -0,0 +1,112 @@
...
+
+alter table task
+    alter column task_data set not null,
+    alter column entity_type set not null,
+    alter column entity_id set not null;
+
+create table task_config
+(
+    warehouse_id uuid references warehouse (warehouse_id) on delete cascade not null,
+    queue_name   text                                                       not null,
```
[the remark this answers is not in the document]
twuebi: i'd not obscure that this is a queue by removing queue from its name. In my understanding, the whole thing is a queue and the elements within the queues are tasks, I'd prefer to keep queue somewhere in it.

crates/iceberg-catalog/src/implementations/postgres/tabular/mod.rs
```diff
@@ -470,9 +470,9 @@ where
         FROM tabular t
         INNER JOIN namespace n ON t.namespace_id = n.namespace_id
         INNER JOIN warehouse w ON n.warehouse_id = w.warehouse_id
-        LEFT JOIN tabular_expirations te ON t.tabular_id = te.tabular_id
-        LEFT JOIN task tt ON te.task_id = tt.task_id
-        WHERE n.warehouse_id = $1
+        LEFT JOIN task tt ON (t.tabular_id = tt.entity_id AND tt.entity_type = 'tabular' AND queue_name = 'tabular_expiration')
+        WHERE (tt.queue_name = 'tabular_expiration' OR tt.queue_name is NULL)
```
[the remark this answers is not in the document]
twuebi: When there's a tabular without a tabular_expiration we end up with NULLs in all of `tt` columns no?

crates/iceberg-catalog/src/implementations/postgres/task_queues.rs
```diff
@@ -0,0 +1,1114 @@
...
+        record_success(id, &self.read_write.write_pool, message).await
+    }
+
+    async fn record_failure(&self, id: Uuid, error_details: &str) -> crate::api::Result<()> {
```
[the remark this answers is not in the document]
twuebi: not sure i understand what method we are missing here

crates/iceberg-catalog/migrations/20250508101407_tasks_store_their_state.sql
```diff
@@ -0,0 +1,134 @@
...
+    config       jsonb                                                      not null,
+    max_age      interval,
+    -- TODO: add poll interval / max_age etc here?
```
c-thiel, line 51: Remove todo

crates/iceberg-catalog/migrations/20250508101407_tasks_store_their_state.sql
```diff
@@ -0,0 +1,134 @@
...
+    alter column task_data set not null,
+    alter column entity_type set not null,
+    alter column entity_id set not null;
+
+create table task_config
+(
+    warehouse_id uuid references warehouse (warehouse_id) on delete cascade not null,
+    queue_name   text                                                       not null,
+    config       jsonb                                                      not null,
+    max_age      interval,
```
c-thiel, line 50: What do you think about `task_instance_heartbeat_timeout` instead of `max_age` now that we have repurposed it?
Airflow has many different timeouts, the most important ones are:
1. [`task_instance_heartbeat_timeout`](https://airflow.apache.org/docs/apache-airflow/stable/configurations-ref.html#task-instance-heartbeat-timeout) (Timeout since Heartbeat)
1. [`execution_timeout`](https://airflow.apache.org/docs/apache-airflow/stable/core-concepts/tasks.html#timeouts) (Timeout since pickup)
1. [`timeout`](https://airflow.apache.org/docs/apache-airflow/stable/core-concepts/tasks.html#timeouts) Timeout since task creation.

I would assiciate `max_age` more with 2) or 3) than with 1). I think in Lakekeeper we need 1) and 2).

crates/iceberg-catalog/migrations/20250508101407_tasks_store_their_state.sql
```diff
@@ -0,0 +1,134 @@
+create type entity_type as enum ('tabular');
+create type task_final_status as enum ('failed', 'cancelled', 'completed');
```
c-thiel, line 2: Could we make `completed` -> `success` for my sanity :D ?

crates/iceberg-catalog/migrations/20250508101407_tasks_store_their_state.sql
c-thiel: Specifying 2) is much more important for users than 1) typically.

crates/iceberg-catalog/src/api/management/mod.rs
```diff
@@ -1492,30 +1553,125 @@ pub mod v1 {
...
     )]
     #[serde(rename_all = "kebab-case")]
     pub enum DeleteKind {
         Default,
         Purge,
     }
 
-    #[must_use]
-    pub fn api_doc<A: Authorizer>() -> utoipa::openapi::OpenApi {
+    pub fn api_doc<A: Authorizer>(
```
c-thiel, line 1582: If we build the APIDoc now dynamically, is there a place where we should cache it? Or should this method be cached?

twuebi:
@c-thiel https://github.com/lakekeeper/lakekeeper/issues/1128

PR #1092, opened 2025-05-13
feat: Improve S3 signer performance via Table specific signer URIs

c-thiel:
Needs https://github.com/apache/iceberg-python/pull/2005 to be released


crates/iceberg-catalog/src/catalog/s3_signer/sign.rs
```diff
@@ -39,12 +39,13 @@ impl<C: Catalog, A: Authorizer + Clone, S: SecretStore>
...
     async fn sign(
         prefix: Option<Prefix>,
-        _namespace: Option<String>,
-        table: Option<String>,
+        tabular_id: Option<uuid::Uuid>,
         request: S3SignRequest,
         state: ApiContext<State<A, C, S>>,
         request_metadata: RequestMetadata,
     ) -> Result<S3SignResponse> {
+        tracing::info!("Signing prefix: {prefix:?} - tabular_id: {tabular_id:?}");
```
twuebi, line 47: wanna keep this on info!?

twuebi, line 47: I may be missing something, but isn't `tabular_id` unused?

c-thiel: Removed the logline, we have one below the branching point now.

crates/iceberg-catalog/src/service/mod.rs
```diff
@@ -84,6 +84,12 @@ pub struct TableId(uuid::Uuid);
 #[serde(transparent)]
 pub struct WarehouseId(pub(crate) uuid::Uuid);
 
+impl Default for WarehouseId {
+    fn default() -> Self {
+        Self(uuid::Uuid::now_v7())
+    }
+}
+
```
twuebi, line 92: I'd prefer not using `default()` as a constructor which returns a random value. Let's maybe use something like `new_random()`. Using `Default` can make it seem like there's a default warehouse id

crates/iceberg-catalog/src/service/storage/s3.rs
```diff
@@ -531,17 +534,17 @@ impl S3Profile {
...
                 remote_signing = true;
             }
         }
 
         if remote_signing {
             config.insert(&s3::RemoteSigningEnabled(true));
-            // Currently per-table signer uris are not supported by Spark.
-            // The URI is cached for one table, and then re-used for another.
-            // let signer_uri = CONFIG.s3_signer_uri_for_table(warehouse_id, namespace_id, table_id);
-            // config.insert("s3.signer.uri".to_string(), signer_uri.to_string());
```
twuebi, line 544: Does this mean we're limited to certain spark versions or does the location fallback help with that? Wouldn't this issue happen with older spark versions?

crates/iceberg-catalog/src/service/storage/s3.rs
```diff
@@ -1181,7 +1184,7 @@ fn validate_bucket_name(bucket: &str) -> Result<(), ValidationError> {
     Ok(())
 }
 
-fn insert_pyiceberg_hack(config: &mut TableProperties) {
+fn push_fsspec_fileio_with_s3v4restsigner(config: &mut TableProperties) {
```
twuebi, line 1187: pyiceberg_hack was a bit more telling about why we're doing this, maybe add a small comment?

crates/iceberg-catalog/src/service/storage/s3.rs
c-thiel: Spark was never the problem - it just didn't support seperate "s3.signer.endpoint", but uris worked. The previous comment is wrong.

crates/iceberg-catalog/src/service/storage/s3.rs
c-thiel: I thought `pyiceberg_hack ` was an aweful name. It didn't say anything about what it was doing and why either. Just that it was for Pyiceberg. Added a comment now.

twuebi (approved):

PR #1097, opened 2025-05-15
feat: Support overwrite option for RegisterTable

c-thiel:
BEGIN_COMMIT_OVERRIDE
feat: Support overwrite option for RegisterTable

chore: Update REST OpenAPI spec
END_COMMIT_OVERRIDE


crates/iceberg-catalog/src/catalog/tables.rs
```diff
@@ -376,43 +377,107 @@ impl<C: Catalog, A: Authorizer + Clone, S: SecretStore>
...
             )
             .await?;
 
-        authorizer
-            .create_table(&request_metadata, tabular_id, namespace_id)
-            .await?;
+        let mut auth_needs_delete = false;
+        // Delete the previous table from authorizer if it exists and differs from the new one
+        if let Some(previous_table_id) = previous_table_id {
+            if previous_table_id != tabular_id {
```
twuebi, line 451: maybe I've missed that detail - how can we be sure that identical ids are in fact the same table?

docs/docs/api/rest-catalog-open-api.yaml
```diff
@@ -629,7 +638,25 @@ paths:
...
+
+
+
+
+
+
+
+
+
+
```
twuebi, line 658: We should get rid of these new line changes

Issue #1093, opened 2025-05-14
feat: Support ETAGs in load table response


c-thiel:
The Iceberg REST spec recently introduced etags / If-None-Match headers in https://github.com/apache/iceberg/commit/e1d24016fa744150af8838e29f52662138671a45.

We want to implement this in Lakekeeper as well.

anandubey:
@c-thiel Please review the PR opened for this feature and let me know if something else needs to be done.

PR #1108, opened 2025-05-20
feat: add `If-None-Match` header for freshness-aware table loading

anandubey:
Closes: #1093


twuebi:
@anandubey thanks for the contribution. I think we're still missing the rust-side implementation here, so far this just changes the openapi spec.

anandubey:
@twuebi Okay, so handling this header in `createTable` and `LoadTable` implementation is what's missing. Correct?

c-thiel:
> @twuebi Okay, so handling this header in `createTable` and `LoadTable` implementation is what's missing. Correct?

Yes exactly. Its actually in the "LoadTableResponse" to be more precise, which means that it should also be returned by other endpoints using this response, such as `register.`

Thanks for taking this!

PR #1132, opened 2025-05-28
add terraform doc snippets

askvinni:
Adds doc snippets for terraform resources for Azure EntraID.

Two important caveats:

- Applications need to be marked with as `enterprise = true` (at least in my organization, it didn't work otherwise)
- Unlike the manual setup, the OpenID aud returned wasn't prefixed with `api://`. I also added a terraform output example that implicitly follows that.


c-thiel (approved):

c-thiel:
Thanks alot @askvinni!

PR #1149, opened 2025-06-03
feat: add quick check path to some `list_...` fns

mooori:
Adds OpenFGA privilege `list_all` and action `can_list_everything_in_namespace` to `namespace`. Based on these, implements a quick check path in `list_namespaces, list_tables, list_views`: If the user has `list_all` on the parent, `can_get_metadata` needs not be checked for the children.

### Tests

- OpenFGA tests have been extended to include the new items.
- Regarding the new logic in `list_*`: It would be nice to have a test hitting the quick check path. Though, this is a private implementation detail mainly re-using existing functionality. Any advice on whether this needs a dedicated test? If yes, a pointer to the testing infra most suitable for that would be appreciated.


mooori:
Tests generated by [`impl_pagination_tests`](https://github.com/lakekeeper/lakekeeper/blob/d3cab68cb5d3e6c6b7fd029831897335cf53dfe6/crates/lakekeeper/src/catalog/mod.rs#L343) pass for namespaces but some of them fail for tables and views. I think this is due to `HidingAuthorizer` not actually checking OpenFGA relation tuples for [`is_allowed_namespace_action`](https://github.com/lakekeeper/lakekeeper/blob/d3cab68cb5d3e6c6b7fd029831897335cf53dfe6/crates/lakekeeper/src/service/authz/mod.rs#L776). Instead it only checks its set of `hidden` objects.

For example: hiding some tables adds strings like `table:id1, table:id2, ...` to `HidingAuthorizer.hidden`. Then `HidingAuthorizer::is_allowed_namespace_action(_, "namespace:id", CanListEverythingInNamespace)` returns true not because `CanListEverythingInNamespace` was granted, but because `namespace:id` is not in the `hidden` set. As a result, the quick check path is executed and hidden tables are shown.

To fix that, I suppose `HidingAuthorizer` needs to be modified? But then I'm struggling to understand what exactly `hidden` means and how it interacts with OpenFGA, since [authz docs](https://docs.lakekeeper.io/docs/nightly/authorization/) don't mention `hidden` items.

authz/openfga/v3.3/components/namespace.fga
```diff
@@ -15,6 +15,7 @@ type namespace
     define pass_grants: [user, role#assignee]
     define manage_grants: [user, role#assignee] or (ownership but not managed_access_inheritance from parent) or manage_grants from parent
     define describe: [user, role#assignee] or ownership or select or create or describe from parent
+    define list_all: [user, role#assignee] or ownership or select or create or list_all from parent
```
c-thiel: There is no need to make this an assignable privilege, there are enough ways to obtain a way to list everything in a Namespace or Warehouse today. Those are:
* Direct assignment of `describe`, `select` or `modify` to an entity
* Direct assignment of any of these to a parent

`can_include_in_list` really only triggers if the permission "bubbles-up", that is, the permission is granted somewhere down the hierarchy, so that in parent folders, we are not allowed to show everything, but navigate the path to the entity we have permissions on.

Because of this we don't really need an additional assignable permission, instead defining:
`define can_list_everything: describe` in the Actions section should be enough.

I also prefer `can_list_everything` to `can_list_everything_in_namespace` to stay consistent with the naming `can_list_tables`, `can_list_views` that is already present.

In addition to adding this to Namespaces, we should also add it to the Warehouse, we can do this as a follow-up though.

authz/openfga/v3.3/schema.json
c-thiel, line 1: The `schema.json` is a json representation of all the `fga` files in the directory. The `json` is actually used by rust later. To update the `json`, we need to run `just update-openfga`. This will update the json in the folder with the highest version.

crates/lakekeeper/src/catalog/namespace.rs
```diff
@@ -72,15 +73,31 @@ impl<C: Catalog, A: Authorizer + Clone, S: SecretStore>
             .await?;
         let mut t = C::Transaction::begin_read(state.v1_state.catalog.clone()).await?;
 
+        let mut can_list_everything = false;
```
c-thiel, line 77: If we also add the Warehouse `can_list_everything` in OpenFGA, this is where we could use it to initialize.

crates/lakekeeper/src/catalog/namespace.rs
```diff
@@ -72,15 +73,31 @@ impl<C: Catalog, A: Authorizer + Clone, S: SecretStore>
...
         if let Some(parent) = parent {
             let namespace_id = C::namespace_to_id(warehouse_id, parent, t.transaction()).await; // Cannot fail before authz
+            let maybe_namespace_id = match namespace_id {
+                Ok(Some(id)) => Some(id),
+                _ => None,
+            };
+
```
c-thiel: This is not not needed and could be achieved more efficiently using `.ok.flatten`.
Instead, lets do the following:

```rust
            let namespace_id = authorizer
                .require_namespace_action(
                    &request_metadata,
                    namespace_id,
                    CatalogNamespaceAction::CanListNamespaces,
                )
                .await?;

            can_list_everything = authorizer
                .is_allowed_namespace_action(
                    &request_metadata,
                    namespace_id,
                    CatalogNamespaceAction::CanListEverythingInNamespace,
                )
                .await?;
```

crates/lakekeeper/src/catalog/tabular.rs
```diff
@@ -36,29 +36,51 @@ macro_rules! list_entities {
                     query,
                 )
                 .await?;
+                let namespace_id =
```
c-thiel: In `list_tables` and `list_views`, where this macro is called, we already have `_namespace_id` available. A Database read is expensive and should be avoided where possible. Lets use the information we already have instead of re-fetching it here.

crates/lakekeeper/src/catalog/tabular.rs
```diff
@@ -36,29 +36,51 @@ macro_rules! list_entities {
...
+                    C::namespace_to_id($warehouse_id, &namespace, trx.transaction()).await;
+                let mut can_list_everything = false;
+                if let Ok(Some(n)) = namespace_id {
```
c-thiel: Same as for namespaces - we already have a non-optional `namespace_id` available in the calling function that we can use.

crates/lakekeeper/src/catalog/tabular.rs
```diff
@@ -36,29 +36,51 @@ macro_rules! list_entities {
...
+                    can_list_everything = authorizer
+                        .is_allowed_namespace_action(
+                            &request_metadata,
+                            NamespaceId::from(*n),
+                            CatalogNamespaceAction::CanListEverythingInNamespace,
+                        )
+                        .await
+                        .unwrap_or(false);
```
c-thiel: Same as above we can't just unwrap any error to false. Maybe OpenFGA is down and this is actually a 500 error that we must popagate. So lets use `?` instead.

c-thiel:
> Tests generated by [`impl_pagination_tests`](https://github.com/lakekeeper/lakekeeper/blob/d3cab68cb5d3e6c6b7fd029831897335cf53dfe6/crates/lakekeeper/src/catalog/mod.rs#L343) pass for namespaces but some of them fail for tables and views. I think this is due to `HidingAuthorizer` not actually checking OpenFGA relation tuples for [`is_allowed_namespace_action`](https://github.com/lakekeeper/lakekeeper/blob/d3cab68cb5d3e6c6b7fd029831897335cf53dfe6/crates/lakekeeper/src/service/authz/mod.rs#L776). Instead it only checks its set of `hidden` objects.
> 
> For example: hiding some tables adds strings like `table:id1, table:id2, ...` to `HidingAuthorizer.hidden`. Then `HidingAuthorizer::is_allowed_namespace_action(_, "namespace:id", CanListEverythingInNamespace)` returns true not because `CanListEverythingInNamespace` was granted, but because `namespace:id` is not in the `hidden` set. As a result, the quick check path is executed and hidden tables are shown.
> 
> To fix that, I suppose `HidingAuthorizer` needs to be modified? But then I'm struggling to understand what exactly `hidden` means and how it interacts with OpenFGA, since [authz docs](https://docs.lakekeeper.io/docs/nightly/authorization/) don't mention `hidden` items.

The `HidingAuthorizer` is just used for tests - when it is used, there is no interaction with OpenFGA. Its is an alternative implementation of the `Authorizer` trait that we use to test paginated filtered lists.
It probably needs some adaptions now - yes. I can't say from the top of my head what changes exactly are needed, but as its only needed in tests, we have a free hand to make any change required.

authz/openfga/v3.3/schema.json
mooori: Yes, I'm running `update-openfga` after updating anything in the schema.

mooori:
I've addressed most comments in the latest commits. How about the following as next steps?

- Adapt `HidingAuthorizer` (#1152) in a separate PR
- Then come back to this PR, make tests pass and wrap it up
- Add `can_list_everything` to `warehouse`

c-thiel:
@mooori  sounds good to me!

c-thiel (approved):
LGTM!

Issue #1152, opened 2025-06-04
test(HidingAuthorizer): avoid false positive `can_list_everything`


mooori:
### Context

#1028, #1149

### Current behavior

`HidingAuthorizer` allows any action on an object if that object is not hidden. Leading to the following problem ([described here](https://github.com/lakekeeper/lakekeeper/pull/1149#issuecomment-2937205766)):

> [...] hiding some tables adds strings like `table:id1, table:id2, ...` to `HidingAuthorizer.hidden`. Then `HidingAuthorizer::is_allowed_namespace_action(_, "namespace:id", CanListEverythingInNamespace)` returns true not because `CanListEverythingInNamespace` was granted, but because `namespace:id` is not in the `hidden` set. As a result, the quick check path is executed and hidden tables are shown.

### Required behavior

`HidingAuthorizer` must avoid such false positive assignments of `can_list_everything`.

mooori:
### Proposal

For (some) namespace actions to be allowed, they must be explicitly assigned first. `HidingAuthorizer`'s state could be extended with something like:

```Rust
// If a catalog namespace action `a` is in the map, it is allowed for namespace `n`
// only if `n` is in the corresponding set.
allowed_namespace_actions: Arc<RwLock<HashMap<CatalogNamespacAction, HashSet<String>>>>
```

#### Usage in existing pagination tests

- Construct a `HidingAuthorizer` that requires `CatalogNamespaceAction::CanListEverything` to be explicitly allowed for a namespace.
- Current test failures in #1149 should be fixed as `can_list_everything` is not granted for any namespace. I.e. no more false positive `can_list_everything`.

#### Usage in new tests

I think with above proposal it should be possible to use `HidingAuthorizer` to construct tests that hit the new `can_list_everything` quick check path. By explicitly allowing `can_list_everything` for the namespace used in the test.

cc @c-thiel what do you think?

c-thiel:
@mooori thanks for the proposal, this LGTM!

mooori:
Confirmed locally that #1162 allows fixing test failures related to `can_list_everything`.

PR #1162, opened 2025-06-07
test(HidingAuthorizer): enable blocking actions

mooori:
Adresses #1152

Slightly deviates from [this proposal](https://github.com/lakekeeper/lakekeeper/issues/1152#issuecomment-2939838153) because after adding `can_list_everything` to `warehouse` I suppose `HidingAuthorizer` would need not only a new field `allowed_namespace_actions` but also `allowed_warehouse_actions`.

I think this can be simplified by instead adding `blocked_actions` to `HidingAuthorizer`, which can then be used to block `warehouse:can_list_everything`, `namespace:can_list_everything`, and potentially actions on other objects later on. This also feels like a more natural extension of `HidingAuhorizer`.

With this in place the problem described in #1152 should be fixed by:

```Rust
authz.block_action(format!("warehouse:{}", CatalogWarehouseAction::CanListEverything));
authz.block_action(format!("namespace:{}", CatalogNamespaceAction::CanListEverything));
```

### DIFF
--- a/crates/lakekeeper/src/service/authz/mod.rs
+++ b/crates/lakekeeper/src/service/authz/mod.rs
@@ -694,8 +694,7 @@ pub(crate) mod tests {
             self.blocked_actions.read().unwrap().contains(action)
         }
 
-        // `namespace:can_list_everything`
-        fn block_action(&mut self, object: &str) {
+        pub(crate) fn block_action(&self, object: &str) {
             self.blocked_actions
                 .write()
                 .unwrap()
@@ -939,7 +938,7 @@ pub(crate) mod tests {
             paste! {
                 #[tokio::test]
                 async fn [<test_block_ $entity _action>]() {
-                    let mut authz = HidingAuthorizer::new();
+                    let authz = HidingAuthorizer::new();
 
                     // Nothing is hidden, so the action is allowed.
                     assert!(authz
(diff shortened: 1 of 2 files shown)


crates/lakekeeper/src/service/authz/mod.rs
```diff
@@ -679,6 +689,18 @@ pub(crate) mod tests {
         pub(crate) fn hide(&self, object: &str) {
             self.hidden.write().unwrap().insert(object.to_string());
         }
+
+        fn action_is_blocked(&self, action: &str) -> bool {
+            self.blocked_actions.read().unwrap().contains(action)
+        }
+
+        // `namespace:can_list_everything`
```
c-thiel: We should probably remove the comment.
Should we make `block_action` `pub(crate)`?

c-thiel (dismissed):

c-thiel:
This LGTM, thanks @mooori!
Just one small comment that we should probably remove.

crates/lakekeeper/src/service/authz/mod.rs
mooori: Done.

crates/lakekeeper/src/service/authz/mod.rs
```diff
@@ -679,6 +689,18 @@ pub(crate) mod tests {
...
+        fn block_action(&mut self, object: &str) {
```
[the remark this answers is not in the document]
mooori: Done, to be consistent with `hide(&self, ...)` (no `mut`).

c-thiel (approved):

Issue #1173, opened 2025-06-13
Turn on the `external-account` feature in the `gcloud-auth` crate by default


gruuya:
**Is your feature request related to a problem? Please describe.**
I'd like to enable integration testing with Lakekeeper pointed against a real GCP project. For this I'm using the [OIDC setup](https://docs.github.com/en/actions/security-for-github-actions/security-hardening-your-deployments/configuring-openid-connect-in-google-cloud-platform), and the corresponding action.

The [preferred way](https://github.com/google-github-actions/auth?tab=readme-ov-file#direct-wif) to get this working is by using Workload Identity Federation, which avoids having to have long-lived credentials. This action outputs a `external_account` type key under `GOOGLE_APPLICATION_CREDENTIALS`.

When I then try to pick this up whilst creating a warehouse
```python
resp = requests.post(
        "http://localhost:8181/management/v1/warehouse",
        json={
            "warehouse-name": name,
            "storage-profile": {
                "type": "gcs",
                "bucket": GCS_BUCKET,
                "key-prefix": key_prefix,
            },
            "storage-credential": {
                "type": "gcs",
                "credential-type": "gcp-system-identity",
            },
        },
    )
```
I see `Failed to create gcp token source from system identity: UnsupportedAccountType(\"external_account\")`. The underlying crate that parses this [has support for this key type](https://github.com/yoshidan/google-cloud-rust/blob/97e02047af32672cf64554338b9cc06228d5aee5/foundation/auth/src/project.rs#L181-L209), but the feature is not turned on by default.

**Describe the solution you'd like**
Enable `external-account` feature for the `gcloud-auth` crate.

**Describe alternatives you've considered**
Create a permanent `service_account` type key and store it in the GitHub secrets, which is against security recommendations.

**Additional context**

PR #1174, opened 2025-06-13
feat: enable external-account feature in gcloud-auth crate

gruuya:
Closes https://github.com/lakekeeper/lakekeeper/issues/1173


Cargo.lock
gruuya, line 1: Wondering whether the addition of these transitive deps is ok.

PR #1176, opened 2025-06-13
chore: remove dependency on fork of rust-rdkafka

Grongrilla:


brgr-s:
@c-thiel with this PR, lakekeeper is referencing the main branch for rust-rdkafka on a commit, that basically bumps every relevant dependency to the current version at the time of the commit. This solves the `cmake` build issue and the requirement to `force-engine = true` for openssl.
Unfortunately, the referenced commit also breaks musl builds.... again. As we discussed, this PR also removes the musl builds for aarch64 and replaces it with a "standard" gnu build. The x86_64 musl build is also removed.
I did some basic testing if CloudEvents via Kafka are still working, and it seems fine.
I added a workflow that does some steps of the release workflow. Feel free to remove that workflow, I kept it in the PR because I thought it might be usefull in the future if more C libs are added.
A double check on the release build from your side would be good.
Some tests are failing randomly... I didn't touch them because I see no relation to this PR.

brgr-s:
closes #591 as we no longer need cross or musl builds

brgr-s:
In the meantime, rust-rdkafka 0.38.0 was released and we can drop the dependency on the main branch completely

brgr-s:
close in favor of #1515 and #1514

PR #1195, opened 2025-06-22
chore(deps): update azure_core and azure_identity to 0.25.0

themightychris:
Update azure_core and azure_identity to the latest.

#832 was failing because the feature name for `reqwest_rustls` dropped the `enable_` prefix in 0.22.0. The other two azure crates are at the latest still at 0.21.0

These updates appear necessary to properly support workload identities

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -19,9 +19,9 @@ dependencies = [
 
 [[package]]
 name = "adler2"
-version = "2.0.0"
+version = "2.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "512761e0bb2578dd7380c6baaa0f4ce03e84f95e960231d1dec8bf4d7d6e2627"
+checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "adler32"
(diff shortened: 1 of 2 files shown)


c-thiel:
It seems we need to change a bit more than just the version.

themightychris:
Yeah looks like it :-( Thanks for the quick eyes @c-thiel !

I'm at a fork in the road with an infrastructure I'm setting up for a client. Any sense if this is a quick thing that might get resolved in the next week or two? If not I'll have to abandon using workload identity for now and go a different direction. If so, I've never worked with Rust so throwing this PR over the fence to see if it passes tests is about as much as I can contribute directly, but I'm deep in the Azure workload identity weeds and could be highly engaged this week in testing a branch I can run `docker build` on and triaging any continued issues

c-thiel:
I just wanted to do the update, but noticed that Microsoft has not yet released a `azure_storage` and `azure_storage_blob` version compatible with the new `azure_core` version.
The updated blob crate is currently in beta: https://crates.io/crates/azure_storage_blob
Unfortunately some features we depend on, such as generating SAS tokens (`get_user_deligation_key`) are not impelmented yet. Hence we currently cannot update core either, as types are not compatible between the two versions.

themightychris:
@c-thiel I found a reference to `azure_storage` and `azure_storage_blobs` as "legacy" crates. It looks like `azure_storage_blob` (no s) might be the new way to access blob storage: https://docs.rs/azure_storage_blob/latest/azure_storage_blob/

*Edit*: nvm you found that already

themightychris:
Looks like Azure/azure-sdk-for-rust#2465 is tracking the missing functionality and going nowhere 😭

themightychris:
Would it make sense to indicate the "Azure System Identity" auth option in Lakekeeper's UI/docs as not working yet? As far as I can tell there's no way for it to work currently and maybe others can be saved from this rabbit hole

c-thiel:
@themightychris yes that makes sense I guess.
Can we narrow down which kind of managed identities are not working? Or is it all of them?

We are integration testing this by using the azure client to login and then expect Lakekeeper to pick up the credentials.
The code is here: https://github.com/lakekeeper/lakekeeper/blob/4725c4a81b69a0d36a6322d4482b1d1519ee5300/.github/workflows/unittests.yml#L171

So the old azure SDK seems to pick up some things, maybe just not proper managed identities that don't require any azure cli / login process?

themightychris:
@c-thiel oh I see, so maybe more of a nuanced note. I'm working with [Microsoft Entra Workload ID with Azure Kubernetes Service (AKS)](https://learn.microsoft.com/en-us/azure/aks/workload-identity-overview)

c-thiel:
https://github.com/Azure/azure-sdk-for-rust/issues/2465 was just closed - I think we can give this another shot!

brgr-s:
@themightychris we had a go at it with a fresh PR (https://github.com/lakekeeper/lakekeeper/pull/1687), but unfortunately the official Azure SDK for Rust is still missing a few features (https://github.com/lakekeeper/lakekeeper/issues/1689)

Issue #1197, opened 2025-06-23
Upgrade Lakekeeper v0.8.5 to v0.9.1 - InternalDatabaseError: Cleanup date missing for deleted tabular


nqvuong1998:
**Describe the bug**
When upgrade Lakekeeper v0.8.5 to v0.9.1 and try to access deleted tables on Lakekeeper UI, it returns error: `InternalDatabaseError: Cleanup date missing for deleted tabular`

**To Reproduce**
1. Change Lakekeeper server (enable soft-delete, exists some soft-deleted tables) and Lakekeeper migration from v0.8.5 to v0.9.1
2. Login Lakekeeper UI
3. Access warehouse -> namespace -> deleted and it will return a error

**Expected behavior**
No error occur

**Screenshots**

![Image](https://github.com/user-attachments/assets/c572e009-acac-4761-8ae1-aacb08784ad0)

**System Information:**
 - Deployment Type: Helm
 - Lakekeeper Version: 0.9.1
 - Query Engine: Spark 3.5.5 with Iceberg 1.9.1
 - Authentication Provider: Keycloak
 - Authorization Provider: OpenFGA
 - OS: Ubuntu
 - Browser: Chrome

nqvuong1998:
cc @c-thiel

c-thiel:
@nqvuong1998 thanks for the issue, I am looking into it!

c-thiel:
@nqvuong1998 I tried various different soft-deletion configs but can't reproduce the issue.
Could you share a Postgres Database dump excluding the "secret" table?
Do you by chance have a DB Backup from before the migration as well?

nqvuong1998:
Hi @c-thiel , here is my zip of pg_dump (excluded public.secret) before the migration.

[lakekeeper.sql.zip](https://github.com/user-attachments/files/20860588/lakekeeper.sql.zip)

c-thiel:
In which Warehouse / Namespace did you see the issues before?

nqvuong1998:
My warehouse is nonprod-sts-bucket
My namespace is risk_mon_out

c-thiel:
Found it, thanks!

c-thiel:
Ok, I found the issue: This problem was due to a race condition that was possible in Versions before `0.9.0` where undropping a table could "cancel" the task but not mark the table as "not deleted" as they did not share a transaction.

In `0.9.0` onwards this is no longer possible.

A fix for your DB state is coming in the referenced PR.

PR #1202, opened 2025-06-23
fix: Undropped tabulars without expiration tasks

c-thiel:
Previous to version 0.9 a race condition could lead to tabular expiration tasks
being cancelled, but the "tabular.deleted_at" field not being reset for the tabular.

This migration handles this in two ways:
1. Restore tabulars without conflicts by setting deleted_at to NULL
2. Create new expiration tasks (90 days from now) for tabulars that would conflict

Closes https://github.com/lakekeeper/lakekeeper/issues/1197

### DIFF
--- a/crates/lakekeeper/src/implementations/postgres/tabular/mod.rs
+++ b/crates/lakekeeper/src/implementations/postgres/tabular/mod.rs
@@ -711,13 +711,13 @@ pub(crate) async fn clear_tabular_deleted_at(
     .fetch_all(&mut **transaction)
     .await
     .map_err(|e| {
-        // tracing::warn!("Error marking tabular as undeleted: {e}");
+        tracing::warn!("Error marking tabular as undeleted: {e}");
         match &e {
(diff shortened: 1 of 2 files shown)


crates/lakekeeper/src/implementations/postgres/tabular/mod.rs
```diff
@@ -711,8 +711,22 @@ pub(crate) async fn clear_tabular_deleted_at(
     .fetch_all(&mut **transaction)
     .await
     .map_err(|e| {
-        tracing::warn!("Error marking tabular as undeleted: {}", e);
-        e.into_error_model("Error marking tabular as undeleted")
+        // tracing::warn!("Error marking tabular as undeleted: {e}");
```
mooori: Uncomment or delete?

c-thiel: uncommented :)

mooori (approved):

Issue #1215, opened 2025-07-02
Lakekeeper incorrectly returning paged response to client that doesn't support pagination


corleyma:
**Describe the bug**

HI think I may have found an unexpected behavior for lakekeeper's pagination (specifically for Iceberg REST spec apis, like list_tables).  The spec says that the REST catalog should return all of the pages if the client request omits next_page_token.  Lakekeeper appears to be returning a paginated response, however.  

**To Reproduce**

Using pyiceberg (patched requests to show the request + headers; pyiceberg does not support pagination currently so we would expect to get everything back):
```
In [5]: sum(1 for identifier in cat.list_tables("test_namespace")) # this namespace has 123 tables in it)

---REQUEST---
GET http://host.docker.internal:8181/catalog/v1/4ed5f7d6-4e1d-11f0-9ed5-83322dfdfc4c/namespaces/test_namespace/tables
Headers: {'User-Agent': 'PyIceberg/0.8.1', 'Accept-Encoding': 'gzip, deflate', 'Accept': '*/*', 'Connection': 'keep-alive', 'Content-type': 'application/json', 'X-Client-Version': '0.14.1', 'X-Iceberg-Access-Delegation': 'vended-credentials'}
---RESPONSE---
Status: 200
Headers: {'content-type': 'application/json', 'x-request-id': '0197c77d-458c-7301-a222-a231021cb3a1', 'vary': 'accept-encoding', 'content-encoding': 'gzip', 'transfer-encoding': 'chunked', 'date': 'Tue, 01 Jul 2025 19:36:05 GMT'}
Body: b'{"next-page-token":"MSYxNzUxMTcwNDkwMDQ0NDQ0JjAxOTdiOWU1LTFlYmUtNzYxMy1hNmM3LTA2OWVkNjczZTZhMA","identifiers":[...]}'

Out[5]: 100 # not 123
```
Note the next-page-token in the logged response headers.

**Expected behavior**
the list tables response should have included all of the tables in the namespace and an empty next_page_token since this client doesn't support pagination.

**Screenshots**

**System Information:**
 - Deployment Type: helm chart
 - Lakekeeper Version: 0.8.5
 - Query Engine: pyiceberg 0.8.1
 - Authentication Provider: None
 - Authorization Provider: None

mooori:
Thank you @corleyma for the report!

[...]

PR #1218, opened 2025-07-03
feat(config): add params for DEFAULT and MAX page size of paginated queries

mooori:
Previously `max` and `default` page sizes for paginated queries were hard coded.

This PR makes it possible to set these values in config. The default values remain unchanged.

The hard coded values could lead to behavior incompatible with specs as described in #1215. Spec compliant behavior can now be achieved by setting the default and max page sizes to `i64::MAX`.

### DIFF
--- a/crates/lakekeeper/src/config.rs
+++ b/crates/lakekeeper/src/config.rs
@@ -537,9 +537,10 @@ impl DynAppConfig {
     }
 
     /// Helper for common conversion of optional page size to `i64`.
-    pub fn page_size_or_pagination_max(&self, x: Option<i64>) -> i64 {
-        let size = x.unwrap_or(self.pagination_size_max.into());
-        std::cmp::min(size, self.pagination_size_max.into())
+    pub fn page_size_or_pagination_max(&self, page_size: Option<i64>) -> i64 {
+        page_size.map_or(self.pagination_size_max.into(), |i| {
+            i.clamp(1, self.pagination_size_max.into())
+        })
     }
 }
 


crates/lakekeeper/src/catalog/config.rs
```diff
@@ -70,9 +70,6 @@ impl<A: Authorizer + Clone, C: Catalog, S: SecretStore>
         config
             .defaults
             .insert("prefix".to_string(), CONFIG.warehouse_prefix(warehouse_id));
-        config
-            .defaults
-            .insert("rest-page-size".to_string(), DEFAULT_PAGE_SIZE.to_string());
```
c-thiel, line 75: Shouldn't we still return the default value here?

crates/lakekeeper/src/catalog/mod.rs
```diff
@@ -197,8 +189,8 @@ where
     // we're filling a auth-filtered page. Without a vec, that won't fly.
 {
     let page_size = page_size
-        .unwrap_or(DEFAULT_PAGE_SIZE)
-        .clamp(1, MAX_PAGE_SIZE);
+        .unwrap_or(CONFIG.pagination_size_default)
+        .clamp(1, CONFIG.pagination_size_max);
     let page_as_usize: usize = page_size.try_into().expect("1, 1000 is a valid usize");
```
c-thiel: This is a problem now with custom specified i64s. Should it be i64s in the config or better u32 or so?

crates/lakekeeper/src/catalog/config.rs
mooori, line 75: Is `rest-page-size` ever used? I searched but couldn't find anything.

mooori: I was going to leave a comment that I removed it because I couldn't find any usage of `rest-page-size`.

Perhaps I missed it or it's accessed externally?

crates/lakekeeper/src/catalog/mod.rs
mooori: Regarding the conversion to usize: The largest unsigned int with infallible conversion to usize is [`u16`](https://doc.rust-lang.org/std/primitive.usize.html#impl-From%3Cu16%3E-for-usize). Would that be enough?

~Otherwise we could make the config fields `usize`, but not sure if that's okay.~
**Edit** `usize` conversion to `i64` can fail, so using `usize` would cause issues elsewhere.

crates/lakekeeper/src/catalog/config.rs
mooori: Added it back in https://github.com/lakekeeper/lakekeeper/commit/301e0ee2ca54cb5f8847263882f2594112327932. This PR might not be the right place to remove this.

c-thiel: Yes, its used externally by clients to use this as the default pagination size if not explicitly overwritten by users.

crates/lakekeeper/src/catalog/mod.rs
c-thiel: Ok, we can keep it, but the expect is nevertheless wrong.

crates/lakekeeper/src/catalog/config.rs
c-thiel: Yes, externally.

PR #1227, opened 2025-07-08
GCP: Support vended credentials refresh

silentsokolov:
https://discord.com/channels/1266476748371918939/1266476749160714243/1390545380323627101

### DIFF
--- a/crates/lakekeeper/src/config.rs
+++ b/crates/lakekeeper/src/config.rs
@@ -124,6 +124,9 @@ pub struct DynAppConfig {
         serialize_with = "serialize_reserved_namespaces"
     )]
     pub reserved_namespaces: ReservedNamespaces,
+
+    /// Controls whether vended credentials should be refreshed or not
+    pub(crate) enable_refresh_credentials: bool,
     // ------------- STORAGE OPTIONS -------------
     /// If true, can create Warehouses with using System Identities.
     pub(crate) enable_aws_system_credentials: bool,
@@ -464,6 +467,7 @@ impl Default for DynAppConfig {
                 "system".to_string(),
                 "examples".to_string(),
             ])),
+            enable_refresh_credentials: false,
             pg_encryption_key: DEFAULT_ENCRYPTION_KEY.to_string(),
             pg_database_url_read: None,
             pg_database_url_write: None,
--- a/crates/lakekeeper/src/request_metadata.rs
+++ b/crates/lakekeeper/src/request_metadata.rs
@@ -6,6 +6,7 @@ use axum::{
     response::{IntoResponse, Response},
 };
 use http::{HeaderMap, Method};
+use iceberg::TableIdent;
 use iceberg_ext::catalog::rest::{ErrorModel, IcebergErrorResponse};
 use limes::Authentication;
 use uuid::Uuid;
@@ -207,6 +208,18 @@ impl RequestMetadata {
     #[must_use]
     pub fn base_uri_management(&self) -> String {
         format!("{}/management", self.base_url())
+    }
+
+    #[must_use]
+    pub fn refresh_credentials_endpoint(
+        &self,
+        warehouse_id: WarehouseId,
+        table: &TableIdent,
+    ) -> String {
+        format!(
+            "v1/{}/namespaces/{}/tables/{}/credentials",
+            warehouse_id, table.namespace, table.name,
+        )
     }
 }
 
--- a/crates/lakekeeper/src/service/storage/gcs/mod.rs
+++ b/crates/lakekeeper/src/service/storage/gcs/mod.rs
@@ -11,7 +11,10 @@ use google_cloud_auth::{
     token::DefaultTokenSourceProvider, token_source::TokenSource as GCloudAuthTokenSource,
 };
 use google_cloud_token::{TokenSource as GCloudTokenSource, TokenSourceProvider as _};
-use iceberg::io::{GCS_DISABLE_CONFIG_LOAD, GCS_DISABLE_VM_METADATA};
+use iceberg::{
+    io::{GCS_DISABLE_CONFIG_LOAD, GCS_DISABLE_VM_METADATA},
+    TableIdent,
+};
 use iceberg_ext::configs::{
     table::{gcs, TableProperties},
     Location,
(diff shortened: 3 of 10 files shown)


c-thiel:
Hey @silentsokolov, what exactly do you want to solve with this PR?
You are saying Iceberg 1.10 works?

The credential refresh endpoint also for GCP is already implemented in Lakekeeper.

Is this about pushing the config options `gcs.oauth2.refresh-credentials-endpoint` conditionally?

There is a lot wrong with the PR currently. If we add something like this, it needs to be configurable per Warehouse and not globally.

silentsokolov:
> Hey @silentsokolov, what exactly do you want to solve with this PR?

Support vended credentials refresh
Long-life tasks does not work, Spark / Custom Kafka jobs are failing due to an error

```
Caused by: com.google.cloud.storage.StorageException: java.lang.IllegalStateException: OAuth2Credentials instance does not support refreshing the access token. An instance with a new access token should be used, or a derived type that supports refreshing.
```

>  You are saying Iceberg 1.10 works?

Yes, Iceberg 1.10 works correctly
There is a bug in Iceberg versions <= 1.9 ([issue link](https://github.com/apache/iceberg/issues/13482)).
When `gcs.oauth2.refresh-credentials-enabled` is enabled, it causes a serialization error

> The credential refresh endpoint also for GCP is already implemented in Lakekeeper.

Do you mean that there's no need to add `gcs.oauth2.refresh-credentials-endpoint` because `gcs.oauth2.refresh-credentials-enabled` is sufficient?

Because [in the code](https://github.com/apache/iceberg/blob/main/gcp/src/main/java/org/apache/iceberg/gcp/gcs/PrefixedStorage.java#L81) there is a check .... and I did not find a default value.

> There is a lot wrong with the PR currently. If we add something like this, it needs to be configurable per Warehouse and not globally.

I am not sure this should be configured per Warehouse. It seems to be defined by the Iceberg spec itself. And should enable by default (in the future)

silentsokolov:
> There is a lot wrong with the PR currently

@c-thiel what exactly?

c-thiel:
I'll try to find some time this week to dig into this.
I don't like that this is a global option and also that we need all of these additional for vended credential generation:
```
        request_metadata: &RequestMetadata,
        warehouse_id: WarehouseId,
        table: &TableIdent,
```

crates/lakekeeper/src/config.rs
c-thiel, line 1: This should not be a global configuration, as we would force it upon all Warehouses.
We have two options: Either always enable it or make it configurable per Warehouse in the GCS Storage profile.

If this is not breaking clients, I would prefer to push the option always to clients and save the configuration option. Could you validate that this works correctly also for newer Iceberg versions?

crates/lakekeeper/src/request_metadata.rs
```diff
@@ -208,6 +209,18 @@ impl RequestMetadata {
...
+
+    #[must_use]
+    pub fn refresh_credentials_endpoint(
+        &self,
+        warehouse_id: WarehouseId,
+        table: &TableIdent,
+    ) -> String {
+        format!(
+            "v1/{}/namespaces/{}/tables/{}/credentials",
+            warehouse_id, table.namespace, table.name,
```
c-thiel: We can't use `Display` of the namespace. Instead, we should use `namespace.to_url_string()`.
Also the `table.name` might require URL encoding.

crates/lakekeeper/src/request_metadata.rs
```diff
@@ -208,6 +209,18 @@ impl RequestMetadata {
     pub fn base_uri_management(&self) -> String {
         format!("{}/management", self.base_url())
     }
+
+    #[must_use]
+    pub fn refresh_credentials_endpoint(
```
c-thiel: This shouldn't be part of the `request_metadata` - it doesn't use self.

crates/lakekeeper/src/service/storage/gcs/mod.rs
```diff
@@ -277,12 +280,16 @@ impl GcsProfile {
...
 
     /// Generate the table configuration for GCS.
+    #[allow(clippy::too_many_arguments)]
     pub(crate) async fn generate_table_config(
         &self,
         _: DataAccess,
         cred: &GcsCredential,
         table_location: &Location,
         storage_permissions: StoragePermissions,
+        request_metadata: &RequestMetadata,
```
c-thiel: Remove `RequestMetadata`. Its not required, see comment from above.

crates/lakekeeper/src/service/storage/gcs/mod.rs
```diff
@@ -277,12 +280,16 @@ impl GcsProfile {
...
+        warehouse_id: WarehouseId,
+        table: &TableIdent,
```
c-thiel: This says table, but you pass in Views as well and return the refresh properties pointing to a table endpoint for them. This is wrong. Must be optional.

docs/docs/configuration.md
c-thiel, line 1: Please try to not re-format existing docs, just add the line you need and reformat that one table

crates/lakekeeper/src/config.rs
silentsokolov: Yes, it works for every Iceberg version that support `refresh-credentials-endpoint`.
We test it on kafka connect / spark / pyiceberg / trino

However, I would prefer to save the configuration option for debug / incidents

c-thiel:
Fixes in Iceberg has been released, I don't think we need this anymore. If we do, please re-create an issue.

silentsokolov:
> Fixes in Iceberg has been released?

@c-thiel fixes?

corleyma:
@c-thiel I am not sure if this issue should actually be closed.  I have also been having spark jobs fail due to GCS vended credentials expiry (running spark 4/Iceberg 1.10/latest lakeeper).

I am not deeply familiar with how this was implemented in Iceberg, but from what I can see in e.g. [Polaris](https://github.com/apache/polaris/pull/2341/files), the catalog needs to return properties for vended credentials refresh (like the refresh endpoint) or the credentials won't be refreshed.  It seems like there are storage-provider specific versions of this property for S3 etc as well?

c-thiel:
@corleyma @silentsokolov

This is defininitely an underspecified area.
The information if credential refresh is supported or not is provided by the /config Endpoint. Servers CAN override the endpoint using the `x.refresh-credentials-endpoint` properties returned to the client. This should be optional. However some clients seem to depend on it.

I created https://github.com/lakekeeper/lakekeeper/pull/1546 as a follow-up to this PR that returns this redundant information for better client compatability. We are not returning `xxx.refresh-credentials-enabled` as the doc and code clearly sets the default to true.

Issue #1190, opened 2025-06-19
feat: Improve Migration Tests in CI


c-thiel:
Add additional tests in CI that test the migration to the current state from past releases of Lakekeeper.
These tests should include past instances with Warehouses that:
- Have hard deletion enabled
- Soft deletion enabled, but expiration runs immediately
- Soft deletion enabled with deletions still pending

PR #1229, opened 2025-07-09
test: lakekeeper migrations from previous versions to head

mooori:
Adds a test for migrating lakekeeper from the `latest-main` release to the head of the current branch. It is based on `docker-compose` and can be run locally with `just test_migration`. A CI workflow is added to run it automatically on PRs.

## Overview

- Brings up lakekeeper `latest-main`, creates tables and drops some of them
- Shuts down the previous lakekeeper instance
- Runs `migrate` and then `serve` with lakekeeper built off the head of the branch

Related to #1190


c-thiel (commented):
I think this is a good start!
A few things to add (apart from the different configs):
- Lets add another .py file that checks if the objects are still read & writable after the migration
- Lets also test that after the new DB-Schema has been applied, the old Lakekeeper instance can still be used for read operations of views and tables.

tests/migrations/spark.py
```diff
@@ -0,0 +1,132 @@
...
+
+        spark.sql(f"SELECT * FROM my_namespace.{table}").show()
+
+    # Use all `DROP` variants to delete some of the tables.
+    print("Deleting some of the tables")
+    for table in TABLES_TO_DROP:
+        spark.sql(f"DROP TABLE my_namespace.{table}")
+
+    # Sleep to let (short) soft-delete timeout expire.
+    # TODO(mooori): get this value from config
```
mooori: Planning to address this in the follow up PR that will enable running the migration test with different delete-profiles.

c-thiel (approved):

Issue #1228, opened 2025-07-08
feat: support for AWS's STS Session Tags (for ABAC authentication)


dolan-a:
**Is your feature request related to a problem? Please describe.**
To reduce the number of required AWS Roles in our AWS+Lakekeeper deployment, we want to use attribute-based access controls (ABAC) with AWS. To do this with Lakekeeper, I believe it would need to provide a Session Tag when assuming a Role via STS. The Tag can then be specified in ~a the Role's Trust Relationship~ (edit: I'm using it in a Policy, but using in the Trust might be an option, too).

**Describe the solution you'd like**
- On the Warehouse Storage Profile, provide a way to specify Session Tags. For example:
    ```json
    "storage-profile": {
        ...
        "sts-session-tags": {"tenant": "tenant-a", "env": "dev"},
        ...
    }
    ```
- In storage/s3.rs (~line 750), optionally pass tags. Here's my lazy version of a hard-coded tag, which I verified works as intended:
    ```rust
    use aws_sdk_sts::types::Tag;
    
    let tags = vec![
    Tag::builder()
        .key("tenant")
        .value("tenant-a")
        .build()
        .expect("Failed to build STS tag"),
    ];
    let assume_role_builder = assume_role_builder.set_tags(Some(tags));
    ```
- Update docs to mention optional tag use when defining permission policy
    ```json
    "Resource": "arn:aws:s3:::my-bucket/tenants/${aws:PrincipalTag/tenant}/catalog/*"
    ```
- ??

**Describe alternatives you've considered**
For our use case, the alternative is to create additional Roles in AWS

**Additional context**
1. I don't mind submitting a PR, but some guidance on how best to pass in the setting would be helpful. (**update:** We can have this discussion on [my recently opened PR](https://github.com/lakekeeper/lakekeeper/pull/1230))
2. I'm curious what your opinion is on using this approach for security
    - In my case, I have a pod using IRSA, and that allows the pod to assume a Role that has access to various tenants' paths, with the path being defined by the lakekeeper-provided tenant tag.
    - External-ID might not be usable with SessionTag. I had tried adding that as a condition on my Trust in AWS, but the UI complained that I couldn't have TagSession combined with it. In my case, I wasn't planning to use External-ID, but I know lakekeeper suggests it is used in certain cases, so wanted to call this out

c-thiel:
Hey @dolan-a, thanks for the Issue and the PR!
I left some comments there.

[...]

PR #1230, opened 2025-07-09
Adding optional STS session-tags for ABAC authorization

dolan-a:
This addresses https://github.com/lakekeeper/lakekeeper/issues/1228 by adding an optional `sts-session-tags` property to `S3Profile` to allow tags to be passed along when assuming an AWS Role.

**TODO:** 
- tests
- doc updating

### DIFF
--- a/crates/lakekeeper/src/service/storage/s3.rs
+++ b/crates/lakekeeper/src/service/storage/s3.rs
@@ -4,6 +4,7 @@ use std::{collections::HashMap, str::FromStr, sync::LazyLock};
 
 use aws_config::{identity::IdentityCache, BehaviorVersion, SdkConfig};
 use aws_sdk_sts::config::{ProvideCredentials as _, SharedIdentityCache};
+use aws_sdk_sts::types::Tag;
 use iceberg_ext::configs::{
     table::{client, custom, s3, TableProperties},
     ConfigProperty, Location,
@@ -88,6 +89,10 @@ pub struct S3Profile {
     #[builder(default = 3600)]
     #[serde(default = "fn_3600")]
     pub sts_token_validity_seconds: u64,
+    /// Optional session tags for STS assume role operations.
+    #[serde(default)]
+    #[builder(default, setter(strip_option))]
+    pub sts_session_tags: Option<HashMap<String, String>>,
     /// S3 flavor to use.
     /// Defaults to AWS
     #[serde(default)]
@@ -743,6 +748,26 @@ impl S3Profile {
 
         let assume_role_builder = if let Some(external_id) = external_id {
             assume_role_builder.external_id(external_id)
+        } else {
+            assume_role_builder
+        };
+
+        let assume_role_builder = if let Some(session_tags) = &self.sts_session_tags {
+            if !session_tags.is_empty() {
+                let tags: Vec<Tag> = session_tags
+                    .iter()
+                    .map(|(key, value)| {
+                        Tag::builder()
+                            .key(key)
+                            .value(value)
+                            .build()
+                            .expect("Failed to build STS tag")
+                    })
+                    .collect();
+                assume_role_builder.set_tags(Some(tags))
+            } else {
+                assume_role_builder
+            }
         } else {
             assume_role_builder
         };
@@ -1569,6 +1594,7 @@ pub(crate) mod test {
             path_style_access: Some(true),
             sts_role_arn: None,
             sts_enabled: false,
+            sts_session_tags: None,
             flavor: S3Flavor::Aws,
             allow_alternative_protocols: Some(false),
             remote_signing_url_style: S3UrlStyleDetectionMode::Auto,
@@ -1613,6 +1639,7 @@ pub(crate) mod test {
             path_style_access: Some(true),
             sts_role_arn: None,
             sts_enabled: false,
+            sts_session_tags: None,
             flavor: S3Flavor::Aws,
             allow_alternative_protocols: Some(false),
             remote_signing_url_style: S3UrlStyleDetectionMode::Auto,
@@ -2003,6 +2030,7 @@ mod is_overlapping_location_tests {
             path_style_access: None,
             sts_role_arn: None,
             sts_enabled: false,
+            sts_session_tags: None,
             flavor: S3Flavor::Aws,
             allow_alternative_protocols: None,
             remote_signing_url_style: S3UrlStyleDetectionMode::Auto,


crates/lakekeeper/src/service/storage/s3.rs
```diff
@@ -747,6 +752,26 @@ impl S3Profile {
...
+        let assume_role_builder = if let Some(session_tags) = &self.sts_session_tags {
+            if !session_tags.is_empty() {
+                let tags: Vec<Tag> = session_tags
+                    .iter()
+                    .map(|(key, value)| {
+                        Tag::builder()
+                            .key(key)
+                            .value(value)
+                            .build()
+                            .expect("Failed to build STS tag")
```
c-thiel: We should return with an error and not panic

crates/lakekeeper/src/service/storage/s3.rs
```diff
@@ -88,6 +89,10 @@ pub struct S3Profile {
     #[builder(default = 3600)]
     #[serde(default = "fn_3600")]
     pub sts_token_validity_seconds: u64,
+    /// Optional session tags for STS assume role operations.
+    #[serde(default)]
+    #[builder(default, setter(strip_option))]
+    pub sts_session_tags: Option<HashMap<String, String>>,
```
c-thiel: To make the structure simpler, I think stripping the Option is also OK. HashMap does implement Default. We don't need to distinguish between not set and set empty.
```suggestion
    pub sts_session_tags: HashMap<String, String>,
```

crates/lakekeeper/src/service/storage/s3.rs
```diff
@@ -747,6 +752,26 @@ impl S3Profile {
             assume_role_builder
         };
 
+        let assume_role_builder = if let Some(session_tags) = &self.sts_session_tags {
+            if !session_tags.is_empty() {
+                let tags: Vec<Tag> = session_tags
+                    .iter()
+                    .map(|(key, value)| {
+                        Tag::builder()
```
c-thiel: Tag actually has a infallible constructor: `Tag { key, value}`. The builder is fallible.
Using it would avoid error handling / panic

c-thiel:
@dolan-a  I fixed the format and expect. I think the code is good now.
If you add the docs we are good to merge.

I won't include this in todays release as it requires a small change in the UI as well.

docs/docs/storage.md
```diff
@@ -66,38 +67,38 @@ Secondly we need to create an AWS role that can access and delegate access to th
...
-			"Effect": "Allow",
-			"Action": [
-				"s3:*"
-			],
-			"Resource": [
-				"arn:aws:s3:::lakekeeper-aws-demo/*"
-			]
-		}
-	]
+    "Version": "2012-10-17",
```
dolan-a, line 70: This file had a mix of tabs and 4-space indentation on the JSON examples, so I adopted 4-space indentation throughout the file.

crates/lakekeeper/src/service/storage/s3.rs
```diff
@@ -747,6 +754,24 @@ impl S3Profile {
             assume_role_builder
         };
 
+        let assume_role_builder = if self.sts_session_tags.is_empty() {
```
dolan-a, line 681: I had to swap the ordering on this, but I verified the session tags are working with the latest code

c-thiel:
Hey @dolan-a, we are currently overhauling our FileIO and I want to add this change directly to the new FileIO.
In this PR we use the session tags for STS only, so for assumed roles that we hand back to clients as part of vended-credentials. Its not used for Lakekeepers FileIO itself.

Do you think this is desired or should the be used for all interactions with S3 in a Warehouse, also interactions of Lakekeeper itself?

dolan-antenucci-imprivata:
It seems that the AWS session tagging is specific to STS role assuming, so unless lakekeeper is making assume-role calls elsewhere, I don't think tags will be needed.

[Here's AWS's description](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_session-tags.html) of this tagging:

> Session tags are key-value pair attributes that you pass when you assume an IAM role or federate a user in AWS STS. You do this by making an AWS CLI or AWS API request through AWS STS or through your identity provider (IdP). When you use AWS STS to request temporary security credentials, you generate a session. Sessions expire and have credentials, such as an access key pair and a session token. When you use the session credentials to make a subsequent request, the request context includes the aws:PrincipalTag context key. You can use the aws:PrincipalTag key in the Condition element of your policies to allow or deny access based on those tags.

dolan-a:
(reposting reply w non-work account)

It seems that the AWS session tagging is specific to STS role assuming, so unless lakekeeper is making assume-role calls elsewhere, I don't think tags will be needed elsewhere.

[Here's AWS's description](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_session-tags.html) of this tagging:

> Session tags are key-value pair attributes that you pass when you assume an IAM role or federate a user in AWS STS. You do this by making an AWS CLI or AWS API request through AWS STS or through your identity provider (IdP). When you use AWS STS to request temporary security credentials, you generate a session. Sessions expire and have credentials, such as an access key pair and a session token. When you use the session credentials to make a subsequent request, the request context includes the aws:PrincipalTag context key. You can use the aws:PrincipalTag key in the Condition element of your policies to allow or deny access based on those tags.

dolan-a:
There were a few failed tests, but they were due to an issue with `snap` installing `just`. Clippy was one of the failed tests, so I confirmed it passes locally at least (after merging in the latest from main).

Let me know if there's anything else needed on this PR

c-thiel:
All good, thanks @dolan-a
I will include this in 0.10 together with a completly new FileIO. I will probly copy some of your changes over to a new PR, as this is easier than rebasing. Will let you know when its ready so you can test it!

dolan-a:
Sounds good. I was also planning a related PR to allow optional changing of the STS Session Name ([role_session_name in s3.rs](https://github.com/lakekeeper/lakekeeper/blob/62214831749df0d3f7076056dabf1adaa44059f4/crates/lakekeeper/src/service/storage/s3.rs#L726)). If you feel like adding, please do, otherwise I can submit a new PR later off the v0.10 release

c-thiel:
Thanks @dolan-a, this LGTM!

dolan-a:
@c-thiel - the openapi schema diff didn't like my manually added string length limits on the tags, so I removed. Re-review please :)

Issue #1120, opened 2025-05-22
URL context "/lakekeeper" not working


lefebsy:
**Describe the bug**
In kubernetes using helm chart with nginx ingress, the ingress path /whatever is not handled by lakekeeper. Same result with LAKEKEEPER__UI__LAKEKEEPER_URL with "http://exemple.com/whatever"

**To Reproduce**
With v0.8.4 or v0.8.5
1. Use helm chart
2. values to ingress path at /whatever and ingressClass at nginx

**Expected behavior**
Host/Whatever/ui in browser should expose lakekeeper service

**Screenshots**
Err 404

**System Information:**
 - Deployment Type: [kubernetes - vanilla yaml, kubernetes helm chart]
 - Lakekeeper Version: [0.8.5, 0.8.4]
 
 - Authentication Provider: [None]
 - Authorization Provider: [None]
 - OS: [Ubuntu]
 - Browser [any]

**Additional context**
Rust_log in debug do not write detailed informations about /context exposed.

martyngigg:
I was just about to open the same bug report.

I've reproduced it in a docker-compose example available [here](https://gist.github.com/martyngigg/651f16a8b2f1d7e3946ddda8a61527cb). My observations are that the management & catalog APIs respond as expected, e.g.

<details>

<summary>Output of GET http://localhost:8080/lakekeeper/management/v1/warehouse</summary>

```sh
$ curl -v -L http://localhost:8080/lakekeeper/management/v1/warehouse
* Host localhost:8080 was resolved.
* IPv6: ::1
* IPv4: 127.0.0.1
*   Trying [::1]:8080...
* connect to ::1 port 8080 from ::1 port 53302 failed: Connection refused
*   Trying 127.0.0.1:8080...
* Connected to localhost (127.0.0.1) port 8080
> GET /lakekeeper/management/v1/warehouse HTTP/1.1
> Host: localhost:8080
> User-Agent: curl/8.7.1
> Accept: */*
> 
* Request completely sent off
< HTTP/1.1 200 OK
< Server: nginx/1.27.4
< Date: Thu, 22 May 2025 08:22:05 GMT
< Content-Type: application/json
< Content-Length: 599
< Connection: keep-alive
< x-request-id: 0196f715-d5f2-72f3-afae-84002114ec89
< vary: accept-encoding
< 
* Connection #0 to host localhost left intact
[...]

PR #1248, opened 2025-07-17
fix: Skip port in determine_base_uri if host already contains it.

martyngigg:
See [comment in #1120](https://github.com/lakekeeper/lakekeeper/issues/1120#issuecomment-3084414525) for details.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -382,9 +382,9 @@ dependencies = [
 
 [[package]]
 name = "async-compression"
-version = "0.4.25"
+version = "0.4.27"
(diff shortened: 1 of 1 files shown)


c-thiel (dismissed):
@martyngigg good catch!
Thanks for the PR, this LGTM. I wasn't aware that the host could contain a port.

martyngigg:
> @martyngigg good catch! Thanks for the PR, this LGTM. I wasn't aware that the host could contain a port.

I wasn't either until yesterday! :)
Happy to help in a small way to contribute to such a useful project.

PR #1253, opened 2025-07-22
Add edge cases test for time conversion

dttung2905:


c-thiel:
Hey @dttung2905, these are purely chrono tests which we shouldn't have as part Lakekeeper.
You might want to donate them to the chrono project instead.

Issue #1255, opened 2025-07-23
Incorrect value in view_version.default_namespace_id field


derbysh:
**Describe the bug**
I have set up several projects and warehouses.
Each warehouse has a namespace with the same name, e.g. `bar`.

When a view is created in the `bar` namespace, the `default_namespace_id` is determined by fetching a namespace by name.
https://github.com/lakekeeper/lakekeeper/blob/5b0cd7e3040aea5c6af8e3228189990d8dbfd4e5/crates/lakekeeper/src/implementations/postgres/tabular/view/mod.rs#L321
This approach can result in the ID of a namespace from a completely different project being assigned, purely based on a name match. And that is what happens in my case.

As a result, referential integrity is compromised:

- The wrong namespace becomes linked to the view.

- That namespace can no longer be deleted by any means due to foreign key constraints.

- One project effectively leaks into another, creating cross-project data contamination.

I think it is a critical issue that needs to be addressed.

**To Reproduce**
Setup several projects/warehouses.
Create namespaces with the same name.
Create views in those namespaces.
See the results.

**Expected behavior**
The default_namespace_id calculation must take into account the current project/warehouse.

**System Information:**
 - Lakekeeper Version: 0.9.2

mooori:
We are currently working on adding `warehouse_id` to the primary keys of `view` and related `view_*` tables. This is part of a larger effort to extend many tables' primary key with `warehouse_id`. Work on it has been started in #1254.

As a result of this, I expect us to take the `warehouse_id` into account when determining the `default_namespace_id`, which should resolve the bug you reported above.

derbysh:
@mooori Thank you for the update. While you're working on this larger effort, do you think it might be possible to temporarily resolve the issue by passing the namespace_id from the calling code? I'm not entirely clear on the underlying logic, so just wanted to check.
https://github.com/lakekeeper/lakekeeper/blob/5b0cd7e3040aea5c6af8e3228189990d8dbfd4e5/crates/lakekeeper/src/implementations/postgres/tabular/view/mod.rs#L128

mooori:
>  temporarily resolve the issue by passing the namespace_id from the calling code

I've tried that locally and unfortunately it breaks some other behavior related to renaming views and object deletion. So I'm afraid it's not as quick of a fix as it initially seems.

Are there some other workarounds that you could apply until we address this issue by adding `warehouse_id` to the composite key?

derbysh:
[...]

PR #1259, opened 2025-07-28
fix: View Versions default Namespace should be limited to its own Warehouse

c-thiel:


c-thiel:
Fixes https://github.com/lakekeeper/lakekeeper/issues/1255

crates/lakekeeper/src/implementations/postgres/tabular/view/mod.rs
```diff
@@ -305,6 +305,7 @@ struct ViewVersionResponse {
 
 #[allow(clippy::too_many_lines)]
 async fn create_view_version(
+    namespace_id: NamespaceId, // Namespace the view is created in != default namespace
```
mooori: > Namespace the view is created in != default namespace

nit: I think it's worth turning that into a doc comment as it's quite subtle (it's also the reason why my attempt of fixing #1255 broke some behavior and tests).

c-thiel: Could you convert it and push to my branch @mooori? My internet isn't great this week

mooori: Sure, done!

Issue #306, opened 2024-08-28
Some error types don't match the specs


twuebi:
- [ ] `NamespaceNotFound` -> `NoSuchNamespaceException`

ndrluis:
Hello @twuebi, could you elaborate on what the problem is with Lakekeeper? I'm a bit confused about the issue.

twuebi:
Hi @ndrluis, it's been a while, I think I created this issue after running the rest compatibility suite of the Java iceberg repo against lakekeeper. A lot of their tests were failing but not due to functional problems but due to the error code not matching what Java would expect. 

Lakekeeper could try to align itself with what the test suite expects. In some cases it wouldn't make a difference in others it makes a difference in what exception a client will see which can or cannot be misleading.

I think in the example above, trying to read a table from a non-existing namespace will produce a table does not exist error instead of a name space does not exist error due to error code not matching the Java expectation.

ndrluis:
In some tests, I got this kind of error that I don't know if it makes sense to fix, like these. @twuebi @c-thiel @mooori what do you think about these tests?

```
Iceberg REST Compatibility Kit > JUnit Jupiter > RESTCompatibilityKitViewCatalogTests > renameTableTargetAlreadyExistsAsView() FAILED
    java.lang.AssertionError:
    Expecting throwable message:
      "Error renaming Table"
    to contain:
      "Cannot rename ns.table to ns.view. View already exists"
````
```
Iceberg REST Compatibility Kit > JUnit Jupiter > RESTCompatibilityKitViewCatalogTests > updateViewLocationConflict() FAILED
    java.lang.AssertionError:
    Expecting throwable message:
      "View not found or action can_get_metadata forbidden for Principal(oidc~9410d0bf-4487-4177-a34f-af364cac0a59)"
    to contain:
      "View does not exist: ns.view"
```

I got some errors that returned a NotAuthorized exception, and this seems really strange to me. I will list all the cases that returned "Not authorized" and try to implement the tests on the Lakekeeper side to reproduce them.

```
Iceberg REST Compatibility Kit > JUnit Jupiter > RESTCompatibilityKitCatalogTests > testConcurrentReplaceTransactionSortOrderConflict() FAILED
    org.apache.iceberg.exceptions.NotAuthorizedException: Not authorized: Failed to authenticate
        at app//org.apache.iceberg.rest.ErrorHandlers$DefaultErrorHandler.accept(ErrorHandlers.java:234)
[...]

PR #1278, opened 2025-08-08
fix: table rename authorization and error handling

ndrluis:
- Fix namespace authorization bug in rename_table to check destination namespace instead of source
- Remove error mapping that converted 403 to 404 to prevent information disclosure
- Implement missing require_namespace_action and require_table_action in AllowAllAuthorizer
- Add comprehensive tests for rename table authorization scenarios

## Summary by CodeRabbit

* **Bug Fixes**
  * Rename now checks drop permission on the source, requires create permission in the destination namespace, and uses destination existence for gating; not-found vs. forbidden responses are now propagated more accurately.
  * Authorization handling adjusted to return appropriate NotFound responses for missing namespaces/tables.

* **Tests**
  * Expanded rename-table test coverage for permission and missing-resource scenarios.
  * Updated Python test to expect NoSuchTable when dropping a non-existent table.

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -1,4 +1,24 @@
+[profile.default]
+default-filter = """
+    not (package(lakekeeper-io) and binary(integration_tests))
+    and not test(::aws_integration_tests::)
+    and not test(::aws_kms_integration_tests::)
+    and not test(::azure_integration_tests::)
+    and not test(::azure_system_credentials_integration_tests::)
+    and not test(::gcs_integration_tests::)
+    and not test(::gcs_hns_integration_tests::)
+    and not test(::gcp_system_credentials_integration_tests::)
+    and not test(::kv2_integration_tests::)
+    and not test(::minio_integration_tests::)
+    and not test(::openfga_integration_tests::)
+    and not test(::r2_integration_tests::)
+"""
+
+[profile.debug]
+default-filter = "test(::gcp_system_credentials_integration_tests::)"
+
 [profile.ci]
+default-filter = "all()"
 failure-output = "immediate-final"
 fail-fast = false
 retries = 0
--- a/.github/workflows/unittests.yml
+++ b/.github/workflows/unittests.yml
@@ -78,23 +78,11 @@ jobs:
           toolchain: ${{ env.RUST_TOOLCHAIN }}
         env:
           RUST_CACHE_KEY_OS: rust-${{ env.RUST_TOOLCHAIN }}-cache-ubuntu-24.04
-          TEST_OPENFGA: 1
-          TEST_AZURE: 1
-          TEST_AWS: 1
-          TEST_GCS: 1
-          TEST_KV2: 1
-          TEST_MINIO: 1
 
       - name: Cargo clippy
         run: just check-clippy
         env:
           RUST_CACHE_KEY_OS: rust-${{ env.RUST_TOOLCHAIN }}-cache-ubuntu-24.04
-          TEST_OPENFGA: 1
-          TEST_AZURE: 1
-          TEST_AWS: 1
-          TEST_GCS: 1
-          TEST_KV2: 1
-          TEST_MINIO: 1
 
   check-generated-openapi-matches:
     runs-on: ubuntu-24.04
@@ -176,7 +164,7 @@ jobs:
 
       - name: Write gcp credential to file
         if: ${{ env.GCS_CREDENTIAL != '' }}
-        run: echo "${{ secrets.GCS_CREDENTIAL }}" > /tmp/gcs.json
+        run: echo '${{ secrets.GCS_CREDENTIAL }}' > /tmp/gcs.json
         shell: bash
 
       - name: Setup Rust toolchain and cache
@@ -223,13 +211,11 @@ jobs:
         run: cargo nextest run --profile ci --all-targets --all-features --workspace
         env:
           DATABASE_URL: postgres://postgres:postgres@localhost:5432/postgres
-          TEST_KV2: 1
           LAKEKEEPER_TEST__KV2__URL: http://localhost:8200
           LAKEKEEPER_TEST__KV2__USER: test
           LAKEKEEPER_TEST__KV2__PASSWORD: test
           LAKEKEEPER_TEST__KV2__SECRET_MOUNT: secret
           # minio test envs
-          TEST_MINIO: 1
           LAKEKEEPER_TEST__S3_BUCKET: tests
           LAKEKEEPER_TEST__S3_REGION: local
           LAKEKEEPER_TEST__S3_ACCESS_KEY: minio-root-user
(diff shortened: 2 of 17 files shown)


ndrluis:
I was diving deeper into this issue, and I believe that since we have AuthZ, we will never be fully compliant with the REST specification, as the specification doesn't account for AuthZ's existence.

Since AuthZ is not a hard requirement for using Lakekeeper, we need to be compliant with both scenarios. For the scenario without AuthZ, commit [bb6f14f](https://github.com/lakekeeper/lakekeeper/pull/1278/commits/bb6f14feb7f30a1f95aeba63834361dbed9e7c5c) addresses this, but I will add tests and implementation for the other cases.

For the AuthZ scenario in the table rename process, we need to verify:

| action | type | source |
|--------|------|--------|
| can_rename | table | origin table |
| can_create_table | namespace | destination namespace |

c-thiel:
@ndrluis  thanks for the PR!
While reviewing it, I actually found an AuthZ bug: The namespace authz check checked the source and not the destination. The source however is already validated by `authorized_table_ident_to_id`.

I think what you are mostly looking for is the 404 status code? Or is the error type also important for you?
What do you think about this fix?
https://github.com/lakekeeper/lakekeeper/pull/1283

Could definitly use your test though. Happy to change the error type as well.

ndrluis:
The original problem I'm trying to solve is the compatibility with the RCK. Currently, we are returning RenameTableIdOrNamespaceNotFound, which is an invalid error type for the clients. The client expects two kinds of errors when this endpoint returns 404: `NoSuchNamespaceException` or `NoSuchTableException`.

I'm explaining more about the issue in this issue: https://github.com/lakekeeper/lakekeeper/issues/306

Currently, I'm adjusting my dev environment because I have to use my personal computer instead of the company machine. However, I have planned to add a test suite for both cases: one without AuthZ to run with RCK and be compliant with the API, and an AuthZ version that tests all the cases we have. I will push the changes to this PR for you to review the test implementations. 

What do you think?

ndrluis:
Another question I have is what kind of tests you think would be better to implement. I'm concerned about having a lot of duplicated tests based on the clients (Trino/PyIceberg/Spark) and integration tests at the Rust level. What are your thoughts about this, @c-thiel? Should we have opinionated documentation that explains how we should implement tests in Lakekeeper?

c-thiel:
Ok, then lets also change the error types in the Authorizer.

For me the most important tests are rust tests. Clients should expect to recieve other error types as well - there are many things that can go wrong. However I agree that for these standard errors we should definitly return them.

We want to tackle better RCK compatability a bit more holisticly in 0.11. It should also run in CI.

ndrluis:
@c-thiel I added the tests and the fix, but I found another case: when we are moving a table between namespaces, we are "deleting" the table from the source namespace, so I'm not sure if we should verify whether the user can_delete/can_rename the original table and can_create a table in the target namespace.

If we wanted to be more restrictive, I believe we should not allow moving tables between namespaces when a user has different permission sets for each namespace, because if they're different, then we're creating an opening for privilege escalation.

However, I believe it would be better for us to improve the implementation of HidingAuthorizer or run these types of tests against OpenFGA, because we're lacking support for specifying resource identifiers.

c-thiel:
You are right, there is a gap regarding privilege escalation. The basic permissions are:
- Delete table in the source namespace and
- Create table in the target namespace

Those are enough for non-managed namespaces, as by default owners can GRANT permissions on owned objects. However, there is one edge case: If managed access is enabled, then owners should not be able to introduce new GRANTS into the managed area - which would stay attached to the moved table. So we should also check managed_access. This isn't easy unfortunately, as the concept of Managed access is currently local to the OpenFGA Authorizer. We might want to introduce a new Catalog Permission: "can_create_moved_objects" or so.

ndrluis:
@c-thiel ready for review

crates/lakekeeper/src/catalog/tables.rs
```diff
@@ -886,6 +886,18 @@ impl<C: Catalog, A: Authorizer + Clone, S: SecretStore>
...
+        authorized_table_ident_to_id::<C, _>(
+            authorizer.clone(),
+            &request_metadata,
+            warehouse_id,
+            source,
+            list_flags,
+            CatalogTableAction::CanDrop,
+            t.transaction(),
+        )
+        .await?;
```
c-thiel, line 899: We don't need this here.
`CanDrop` is less specific than `CanRename`, which is checked below. `CanRename` might be implemented as `CanDrop` AND Something else, depending on the Authz system. In our case, for the OpenFGA Authorizer, they are even identical. In any case, the AuthZ system must make sure that `Rename` is as least as restrictive as drop.

crates/lakekeeper/src/catalog/tables.rs
```diff
@@ -1470,7 +1482,6 @@ pub(crate) async fn authorized_table_ident_to_id<C: Catalog, A: Authorizer>(
     authorizer
         .require_table_action(metadata, table_id, action)
         .await
-        .map_err(set_not_found_status_code)
```
c-thiel, line 1473: We probably shouldn't remove this for two reasons:
1. Many query engines cannot process a 403 correctly. Not for rename table - it might be OK there, but for example table_exists with a 403 / 401 code is not widely supported. Some versions of trino will even abort "SHOW TABLES FROM ..." if they encounter a 403 / 401. So we must retain the 404 error code. This method is used not only for rename, but other methods as well.
2. It leaks information about tables that the client is not authorized to know. Lets assume a client queries a table "hr.employees_to_fire", but has no access to it. Then, if we would return a different error code based on the fact if the table exists or not, we would leak the information if the table exists or not. Thus we should always return 404 with TableNotFound. This is conclusive with the UI, where we won't show these tables. So there is no reason to return a 401/403 - instead, the table should appear non-existant.

crates/lakekeeper/src/service/authz/implementations/allow_all.rs
```diff
@@ -240,4 +241,45 @@ impl Authorizer for AllowAllAuthorizer {
     async fn delete_view(&self, _view_id: ViewId) -> Result<()> {
         Ok(())
     }
+
+    async fn require_namespace_action(
```
c-thiel, line 245: We should not implement these two methods (require_namespace_action & require_table_action) here - it only leeds to errors.
These methods are already implemented for `AllowAllAuthorizer` via the implementations in the `Authorizer` trait. The behaviour implemented here has no effect for the `OpenFGAAuthorizer` - only for `AllowAllAuthorizer`. Diverting behaviour is an anti-pattern and should only be applied if we have very good reasons for it - which I believe we don't.

c-thiel (changes requested):

crates/lakekeeper/src/catalog/tables.rs
ndrluis: Agreed. Just to confirm we're aligned on the error handling

If a user doesn't have permission to rename a table, raise a 404 with NoSuchTableException
If a user doesn't have permission to create a table in the destination namespace, raise a 404 with NoSuchNamespaceException
If a user tries to rename a table that doesn't exist, raise a 404 with NoSuchTableException
If a user tries to rename a table to a namespace that doesn't exist, raise a 404 with NoSuchNamespaceException

WDYT?

c-thiel: Perfect :)

c-thiel:
Closed in favor of: https://github.com/lakekeeper/lakekeeper/pull/1346

PR #1283, opened 2025-08-13
fix: Proper error when destination NS of rename OP doesn't exist

c-thiel:
## Summary by CodeRabbit

* **Bug Fixes**
  * Renaming a table now validates permissions against the destination namespace, ensuring correct authorization.
  * Improved error handling during table rename to return a clearer “Not Found” response when the destination namespace is unavailable or invalid.


c-thiel:
Closed in favor of https://github.com/lakekeeper/lakekeeper/pull/1346

PR #1338, opened 2025-09-16
fix: case insensitivity of catalog endpoints

### LINKED ISSUE #1319: Ensure case insensitivity of Catalog endpoints

mooori:
Closes #1319 

It seems like there are no other getters where collation could cause a similar issue, or did I miss something? In particular namespaces should not be affected as there is no corresponding `namespace_idents_to_ids` function. Besides tabulars and namespaces, I think collation is applied nowhere else.

Regarding the modification of the db query: indices generated on the Rust side would need to be plumbed into a new, non-static CTE at the beginning of the query. Going instead with the jsonb encoding of `Vec<Vec<String>>` appeared to be the simpler approach. As a positive side effect, the whole query is static now.

### DIFF
--- a/crates/lakekeeper/migrations/20250916082614_jsonb_to_textarr.sql
+++ b/crates/lakekeeper/migrations/20250916082614_jsonb_to_textarr.sql
@@ -1,0 +1,3 @@
+CREATE FUNCTION jsonb_to_textarr(jsonb) RETURNS text[] AS $$
+    SELECT array_agg(txt) FROM jsonb_array_elements_text($1) txt;
+$$ LANGUAGE SQL;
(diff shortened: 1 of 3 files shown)


mooori:
@coderabbitai full review

mooori:
@coderabbitai full review

mooori:
@coderabbitai incremental

mooori:
@c-thiel this is now on top of the postgres migration.

c-thiel:
@mooori I added a test in [1742e2a](https://github.com/lakekeeper/lakekeeper/pull/1338/commits/1742e2a873fa497702bc7c2c289392fd71791971) which seems to fail with the same error.

c-thiel (changes requested):
Integration test doesn't seem to work.

c-thiel (approved):

c-thiel (approved):

crates/lakekeeper/migrations/20250916082614_jsonb_to_textarr.sql
c-thiel, line 1: I am unsure if this warrants a migration.
Migrations are here forever - do you think having it inline would be large disadvantage?

mooori: The function is now inlined and the migration removed. The motivation for having a function was readability, though I think after inlining it's still manageable.

PR #1340, opened 2025-09-16
feat!: Require warehouse_id when checking permissions on views or tables via ID

mooori:
`warehouse_id` is added as preparation for v0.11, which will then use it.

## Breaking change

Adds a field to `TabularIdentOrUuid` which is used on the `/check` endpoint of the management API.

## Summary by CodeRabbit

* **New Features**
  * Added warehouse scoping to table/view identification. When using the table-id path, warehouse-id is now required alongside table-id.
  * For the namespace-based path, only namespace is required; table and warehouse-id are now optional.

* **Documentation**
  * Updated OpenAPI schema to reflect the new warehouse-id requirement for table-id paths and relaxed requirements for the namespace-based path.


mooori:
@coderabbitai full review

c-thiel (approved):

PR #1460, opened 2025-10-24
fix: Debug assertion table identifier mismatch for signer

c-thiel:
## Summary by CodeRabbit

* **Bug Fixes**
  * S3 signing now enforces location- and table-based permission checks using real lookup results instead of placeholder data.
  * Requests to sign URLs for non-existent table locations now return a 400 Bad Request with clear error handling.
  * Removed fallback placeholder behavior, improving correctness and reducing unexpected authorization results.


c-thiel:
@coderabbit full review

PR #1509, opened 2025-11-24
feat: Add ETag to responses and evaluate If-None-Match Header

Midor:
## Summary by CodeRabbit

* **New Features**
  * HTTP ETag support for caching; responses may include ETag headers.
  * Table API accepts If-None-Match values and can return 304 Not Modified when metadata is unchanged.
  * Conditional load-table responses now surface a NotModified variant when applicable.

* **Tests**
  * Added unit and end-to-end tests for ETag generation, parsing, and conditional load-table behavior.

### DIFF
--- a/crates/lakekeeper/src/server/tables/load_table.rs
+++ b/crates/lakekeeper/src/server/tables/load_table.rs
@@ -32,11 +32,8 @@ fn get_etag(table_info: &TabularInfo<TableId>) -> Option<ETag> {
         .map(create_etag)
 }
 
-fn etag_already_present(etags: &[ETag], etag: Option<&ETag>) -> bool {
-    match etag {
-        Some(etag) => etags.iter().any(|e| e == etag || e == &ETag::from("*")),
-        None => false,
-    }
+fn etag_already_present(etags: &[ETag], etag: &ETag) -> bool {
+    etags.iter().any(|e| e == etag || e == &ETag::from("*"))
 }
 
 /// Load a table from the catalog
@@ -78,8 +75,11 @@ pub(super) async fn load_table<C: CatalogStore, A: Authorizer + Clone, S: Secret
 
     // ------------------- ETAG CHECK -------------------
     let etag = get_etag(&table_info);
-    if let Some(etag_value) = etag.as_ref().map(|e| e.as_str().trim_matches('"'))
-        && etag_already_present(&etags, Some(&etag_value.into()))
+    if let Some(etag_value) = etag
+        .as_ref()
+        .map(|e| e.as_str().trim_matches('"'))
+        .map(ETag::from)
+        && etag_already_present(&etags, &etag_value)
     {
         return Ok(LoadTableResultOrNotModified::NotModifiedResponse(
             etag.unwrap(),


crates/iceberg-ext/src/catalog/rest/table.rs
```diff
@@ -99,11 +105,104 @@ pub struct CommitTransactionRequest {
...
+    let hash = xxh3_64(text.as_bytes());
+    format!("\"{hash:x}\"")
+}
+
 #[cfg(feature = "axum")]
-impl_into_response!(LoadTableResult);
+impl IntoResponse for LoadTableResult {
+    fn into_response(self) -> axum::http::Response<axum::body::Body> {
+        // self.metadata_location
+        let metadata_location = self.metadata_location.clone().unwrap_or_default();
```
c-thiel: I don't think we should return an ETag for staged tables (tables without metadata location), as fetching them would yield 404 anyway.

crates/iceberg-ext/src/catalog/rest/table.rs
```diff
@@ -99,11 +105,104 @@ pub struct CommitTransactionRequest {
...
+        let etag = create_etag(&metadata_location);
+
+        let mut header = HeaderMap::new();
+        header.insert(header::ETAG, etag.parse().unwrap());
```
c-thiel: We should handle the error gracefully and not panic. Lets use `tracing::error` to log the error and not return anything.

Cargo.toml
```diff
@@ -26,14 +25,14 @@ async-nats = "0.44.0"
 async-stream = "0.3.6"
 async-trait = "0.1.89"
 aws-config = { version = "1.8.10", features = ["behavior-version-latest"] }
+aws-credential-types = { version = "1.2" }
+aws-sdk-s3 = { version = "1.65" }
 aws-sdk-sts = "1.82.0"
+aws-sigv4 = { version = "1.2" }
```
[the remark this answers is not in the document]
c-thiel: formatted correctly now

crates/iceberg-ext/src/catalog/rest/table.rs
```diff
@@ -99,11 +105,104 @@ pub struct CommitTransactionRequest {
...
+        (header, axum::Json(self)).into_response()
+    }
+}
```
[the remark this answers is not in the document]
Midor: 1. The edge case for a staged table with an empty metadata_location is now implemented. As well as tests added for this edge case.
2. The logic for handling the possible panic while `etag.parse()` was added, although I was not able to synthetically create a test case for this, as `create_etag` doesn't allow the creation of invalid etags.
Additionally: Another test for correct handling of the JSON body was added.

crates/iceberg-ext/src/catalog/rest/table.rs
```diff
@@ -99,11 +105,104 @@ pub struct CommitTransactionRequest {
...
+        };
+
+        let response = load_table_result.into_response();
+        let headers = response.headers();
+
+        assert_eq!(
+            headers.get(header::ETAG).unwrap(),
+            &create_etag("s3://bucket/table/metadata.json")
+        );
+    }
```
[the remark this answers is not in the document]
Midor: Test was added. See https://github.com/lakekeeper/lakekeeper/pull/1509#discussion_r2556465445

crates/iceberg-ext/src/catalog/rest/table.rs
Midor: Logic is now implemented. See https://github.com/lakekeeper/lakekeeper/pull/1509#discussion_r2556465445

crates/iceberg-ext/src/catalog/rest/table.rs
Midor: This is also added in the current implementation. It will still return an empty `HeaderMap` and the JSON body, but it will log this error with the suggested method.

Midor:
@coderabbitai Added the handling of multiple separate If-None-Match Headers, concating all into a single vector.

Midor:
@coderabbitai review

crates/lakekeeper/src/api/iceberg/v1/tables.rs
```diff
@@ -70,6 +73,27 @@ impl From<ListTablesQuery> for PaginationQuery {
...
+                let mut header = HeaderMap::new();
+                header.insert(header::ETAG, etag.parse().unwrap());
+                (StatusCode::NOT_MODIFIED, header).into_response()
+            }
+            LoadTableResultOrNotModified::LoadTableResult(load_table_result) => {
+                load_table_result.into_response()
+            }
+        }
+    }
+}
```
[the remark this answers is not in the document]
Midor: This should now be handled the same as with LoadTableResult::into_response. Additionally this time I was able to create a test case for this problem.

crates/iceberg-ext/src/catalog/rest/table.rs
```diff
@@ -99,11 +105,153 @@ pub struct CommitTransactionRequest {
...
+
+        let Some(ref metadata_location) = self.metadata_location else {
+            return (headers, body).into_response();
+        };
+        let etag = create_etag(metadata_location);
+
+        let Ok(header_value) = etag.parse::<HeaderValue>() else {
+            tracing::error!(
+                "Failed to create valid ETAG header from etag string: {}",
+                etag
```
c-thiel: Instead of let ... lese, lets use match, so that we can include the Err variant in the error trace.

crates/lakekeeper/src/api/iceberg/v1/tables.rs
```diff
@@ -70,6 +73,32 @@ impl From<ListTablesQuery> for PaginationQuery {
     }
 }
 
+#[derive(Debug, Clone, PartialEq)]
+pub enum LoadTableResultOrNotModified {
+    LoadTableResult(LoadTableResult),
+    NotModifiedResponse(String),
```
c-thiel: Just a nit: It feels a bit strange to have the two enum variants handle etags differently. The first variant contains the information to compute it, while the second variant is the etag itself.
Also when looking at this public type, I don't like much to use `String` as the inner type, as it doesn't give any information about the contained data (the `NotModifiedResponse` doesn't either).

I think I would prefer one of the following:

1. Use `TableInfo` as the inner type
2. Introduce a new leightweight Etag type that is pretty much transparent to string

WDYT?

crates/lakekeeper/src/server/tables/load_table.rs
```diff
@@ -59,6 +79,18 @@ pub(super) async fn load_table<C: CatalogStore, A: Authorizer + Clone, S: Secret
     )
     .await?;
 
+    let etag = get_etag(&table_info);
+    if etag_already_present(
+        &etags,
+        etag.clone()
+            .map(|e| e.trim_matches('"').to_string())
+            .as_ref(),
```
c-thiel: Nit: We have two useless clones in here: `.clone()` and `to_string()`.

```suggestion
        etag.as_ref().map(|e| e.trim_matches('"'))
```

This won't compile because `etag_already_present` requires `&String` instead of `&str` for its second argument. We should almost always prefer `&str` to `&String`.

crates/lakekeeper/src/server/tables/load_table.rs
```diff
@@ -16,13 +17,31 @@ use crate::{
...
 
+fn get_etag(table_info: &TabularInfo<TableId>) -> Option<String> {
+    table_info
+        .metadata_location
+        .as_ref()
+        .map(lakekeeper_io::Location::as_str)
+        .map(create_etag)
+}
+
+fn etag_already_present(etags: &[String], etag: Option<&String>) -> bool {
```
c-thiel:
```suggestion
fn etag_already_present(etags: &[String], etag: Option<&str>) -> bool {
```

Then below:
`etags.iter().any(|s| s == etag)`

crates/iceberg-ext/src/catalog/rest/table.rs
Midor: Should be fixed now.

crates/lakekeeper/src/server/tables/load_table.rs
Midor: Is fixed, as well as the underlaying data structure.

crates/lakekeeper/src/api/iceberg/v1/tables.rs
Midor: At first I tended to Option 1, but this would mean to create the etag once for checking against the If-None-Match Header, and the second time to return it in the NotModified Response. So I created the ETag Tuple Struct https://github.com/lakekeeper/lakekeeper/blob/269a017fad73f5243d54cbbf20a88e4de41c659e/crates/iceberg-ext/src/catalog/rest/table.rs#L108-L128 and used it everywhere, where normally would be returned or used an ETag.

crates/lakekeeper/src/server/tables/load_table.rs
Midor: The functions are now using the ETag Tuple Struct (see https://github.com/lakekeeper/lakekeeper/pull/1509#discussion_r2559945204). But the improvement with `etags.iter().any` is implemented, as well as a simple improvement to put the check for wildcard (`*`) into the call to `etags.iter().any`

Midor:
@coderabbitai review

Midor:
# Comparison between this implementation and [Apache Iceberg](https://github.com/apache/iceberg)

## ETag Creation
The [RFC 723 - Hypertext Transfer Protocol (HTTP/1.1): Conditional Requests - 2.3. ETag](https://datatracker.ietf.org/doc/html/rfc7232#section-2.3) mentions, that ETags need to be put in quotes, so strong ETags can be distinguished from weak ETags (containing a `W/` as Prefix), and also from a wildcard `*`.

In the implemention from [Apache Iceberg - ETagProvider](https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/rest/ETagProvider.java#L31-L36) the ETags are generated without those quotes. Also in the [Unit-Tests](https://github.com/apache/iceberg/blob/main/core/src/test/java/org/apache/iceberg/rest/TestETagProvider.java) They showed exactly this usage. And in some Implementations, which I guess are [Test Fixtures](https://github.com/apache/iceberg/blob/main/core/src/test/java/org/apache/iceberg/rest/RESTCatalogAdapter.java#L292-L301) they showed that it is used as-is and not quoted afterwards.

## "If-None-Match" Header Comparison
To handle cases, where the "If-None-Match" is sent with an ETag missing the quotes, we extract the ETags and remove the surrounding quotes and so compare only the content. This ensures, that even if clients strip the quotes or have there own logic to rebuild the ETag, it should still work.

Even without this handling this shouldn't be a problem at all, as the implementation from the [Test Fixtures](https://github.com/apache/iceberg/blob/main/core/src/test/java/org/apache/iceberg/rest/RESTCatalogAdapter.java#L292-L299) (it is also the same for other cases in the same file) implies to take the incoming "If-None-Match" header and compare it directly against the previously generated ETag.
So the important point is to have the same generation logic for ETags during their creation and comparison.

**Additional Note**: Another change we have in comparison to [Apache Iceberg](https://github.com/apache/iceberg) is the following.

We also handle the case, that the "If-None-Match" header can be added **multiple** times to a request, whereas in the [Test Fixtures](https://github.com/apache/iceberg/blob/main/core/src/test/java/org/apache/iceberg/rest/RESTCatalogAdapter.java#L292-L293) only the **first** "If-None-Match" header is used, in which case some cache misses could happen.

Typically a sender is advised to send a single header with comma separated values instead of multiple headers, so this change should not be a big problem, except for the use-case of a single header with multiple comma separated values, which are also handled as-is in [Apache Iceberg](https://github.com/apache/iceberg).

But it is also not forbidden, and so a recipient should be able to handle multiple headers the same way as comma separated values. See [RFC 7230 - Hypertext Transfer Protocol (HTTP/1.1): Message Syntax and Routing - Field Order](https://datatracker.ietf.org/doc/html/rfc7230#section-3.2.2).

PR #1511, opened 2025-11-24
feat: Add warehouse-level STS and remote signing control

### LINKED ISSUE #1498: feat: Add Warehouse option to disable STS & Remote Signing on Warehouse level
Currently even if STS is disable for a Warehouse, Remote signing is always enabled.
With this issue we want to introduce a flag that disables all forms of vended-credentials for Warehouses. This is most important for S3, but if we can add support ADLS & GCP without much additional work as well, we should.

AndreaBozzo:
Leaving this here so CI starts running

## Summary
Implements warehouse-level flags to control STS (vended credentials) and remote signing functionality.

Will close #1498

### Database
- Added migration to create `sts_enabled` and `remote_signing_enabled` columns in the warehouse table
- Defaults: `true` for backward compatibility

### Core Implementation
- Added `sts_enabled` and `remote_signing_enabled` fields to `ResolvedWarehouse`
- Updated all 9 Postgres warehouse queries to include new fields
- Modified storage profile generation logic:
  - **S3**: Checks warehouse flags before generating STS credentials or enabling remote signing. Falls back to remote signing if STS disabled but remote signing enabled
  - **ADLS**: Checks warehouse flag before generating SAS tokens
  - **GCS**: Checks warehouse flag before generating downscoped tokens
- Added new error types:
  - `VendedCredentialsDisabled` (HTTP 403 Forbidden)
  - `RemoteSigningDisabled` (HTTP 403 Forbidden)
- Protected S3 remote signing handler to verify `warehouse.remote_signing_enabled` before processing requests

### Files Modified (15)
- `migrations/20251124000000_warehouse_credential_vending_flags.sql` (NEW)
- `src/service/catalog_store/warehouse.rs`
- `src/implementations/postgres/warehouse.rs`
- `src/service/storage/{s3.rs,az.rs,gcs/mod.rs,mod.rs,error.rs}`
- `src/server/s3_signer/sign.rs`
- `src/server/tables{.rs,/create_table.rs,/load_table.rs}`
- `src/server/views/{commit.rs,create.rs,load.rs}`

## Behavior
- When `sts_enabled=false`: No STS/SAS/downscoped tokens generated
- When `remote_signing_enabled=false`: Remote signing blocked
- **S3 special case**: Can fallback from STS to remote signing if the latter is enabled
- Both disabled results in HTTP 403 Forbidden error

## Testing Notes
- Code compiles successfully (type checks pass)
- All 8 call sites updated to pass warehouse flags
- Test helper functions updated with default values (`true`)

## Backward Compatibility
Fully backward compatible - existing warehouses default to both flags enabled.

## Optional Next Steps
- Add integration tests for disabled credential scenarios
- Add management API endpoints to modify flags

## Release Notes

* **New Features**
  * Added warehouse-level flags to control credential vending and remote signing capabilities.
  * Added enforcement checks that prevent vended credentials and remote signing when warehouse flags are disabled.


AndreaBozzo:
@c-thiel marking this Open as workflows are successful, optional notes can be taken now or in specific sub-issues if needed

crates/lakekeeper/migrations/20251124000000_warehouse_credential_vending_flags.sql
```diff
@@ -0,0 +1,12 @@
+-- Add warehouse-level flags to control STS and Remote Signing
+-- Issue #1498: Add Warehouse option to disable STS & Remote Signing on Warehouse level
+
+-- Add sts_enabled flag (defaults to true for backward compatibility)
+ALTER TABLE "warehouse" ADD COLUMN "sts_enabled" BOOLEAN NOT NULL DEFAULT true;
+
+-- Add remote_signing_enabled flag (defaults to true for backward compatibility)
+ALTER TABLE "warehouse" ADD COLUMN "remote_signing_enabled" BOOLEAN NOT NULL DEFAULT true;
```
c-thiel, line 8: I would prefer to add these flags to the corresponding `StorageProfiles` for the following reasons:
1. `remote_signing` is something very specific to S3, it doesn't make sense for other Storages
2. The term STS is not used by Azure. Future storage services, like Hadoop, don't support this
3. In the future we might want to extend Warehouses to support multiple storage profiles. This setting is very much scoped to individual profiles then, and not to all profiles that are associated with a Warehouse.

WDYT?

AndreaBozzo:
Hi @c-thiel, thanks for the feedback! You're right about moving these flags to the storage profile level. 

But I need your input on a few design decisions before proceeding as this is an impactful change :

Design Questions:
1. Naming Convention:

Option A: Rename sts_enabled to vended_credentials_enabled

Provider-agnostic terminology
Aligns with existing VendedCredentialsDisabled error
Covers S3 STS, Azure SAS, GCS downscoped tokens
Breaking change for S3Profile (requires gradual migration)

Option B: Keep sts_enabled for all profiles
Minimal change
AWS-specific terminology becomes "standard"

Which do you prefer? I'm guessing A but as it is an important naming decision, i'll leave this up

2. Handling S3Profile Duplication ( this is important as i might need to close & rebase this pr )
S3Profile already has sts_enabled (line 91), creating duplication: Option A: Keep existing sts_enabled + add only remote_signing_enabled
Option B: Add vended_credentials_enabled + support both with OR logic temporarily
Option C: Direct replacement (immediate breaking change) 

How should we handle this?

3. Handling S3Profile Duplication

Since S3Profile already has sts_enabled: 
Option A: Keep existing sts_enabled + add only remote_signing_enabled 
Option B: Add vended_credentials_enabled + support both with OR logic temporarily 
Option C: Direct replacement (immediate breaking change) 

How should we handle this?

4.  Default Values

Option A: Defaults true
Backward compatible
Matches this pr behavior

Option B: Defaults false
More secure (explicit opt-in)
Breaking change

Confirm defaults true for backward compatibility?

5.Migration Strategy

Option A: Clean Slate

Remove warehouse columns completely
JSONB + serde defaults handle everything
Zero data migration needed

Option B: Temporary Fallback

Keep warehouse columns as fallback
Read from profile if present, else from warehouse
Remove in future major version

Option C: ???

Which approach do you prefer?

6. API Management

Currently missing API endpoints to modify these flags post-creation. Should the refactoring include dedicated API endpoints, or should users modify via the existing update storage profile endpoint?

Side personal Note

The good news: Since storage profiles are already JSONB, we need zero complex database migrations. Serde defaults will handle backward compatibility automatically. Let me know your preferences on these decisions and i'll proceed with the modifications!

c-thiel:
@AndreaBozzo  some very good points - let me go through them:

1. Naming: We can't really re-use S3 `sts_enabled` unfortunately, because there are some on-prem S3 storages out there that don't support `STS`. In this case, users mostly want to to use remote-signing. So we need an option to say: Disable STS but use Remote signing. And if we rename the existing option, we cannot do this anymore. My current preference would be to introduce `remote_signing_disabled` for the S3Profile, use `sts_enabled` for GCS (as GCP also calls this STS) and use either `sts_enabled` or maybe even more specific `sas_enabled` for Azure. Beeing specific also leaves room for other options in the future, that can then be enabled / disabled individually.
2. I think this might be clarified in my point 1?
4. Keep it backward compatible, so the default should enable STS / Remote signing.
5. A :)
6. As we now add it to the storage profile nothing more needs to be done. Updating via the existing endpoint is OK as long as the request object is backward compatible (has a default)

AndreaBozzo:
Thanks christian, i'll try to re-submit tonight or tomorrow

PR #1534, opened 2025-12-05
feat: Extend housekeeping tasks by adding project_id to task tables and make warehouse_id optional to enable project-wide tasks.

Midor:
## Summary by CodeRabbit

* **New Features**
  * Project-level task queue config endpoints added.
  * Project-scoped task management: list, get details, and control project tasks.
  * Task scheduling now includes project context for clearer multi-tenant behavior.

* **API Changes**
  * Task/list/detail responses enhanced with project-aware metadata, entity identifiers, and richer execution details.
  * New filtering options for warehouse- and project-level tasks (entity and time-range support).

* **Chores / Migration**
  * DB migration to enable project-scoped tasks and configs.


Midor:
@coderabbitai review

Midor:
@coderabbitai review

crates/lakekeeper/migrations/20251128101923_enable_tasks_on_project_level.sql
```diff
@@ -0,0 +1,84 @@
...
+	add primary key (task_config_id);
+
+-- 5. Modify task constraints and make warehouse_id optional
+alter table task 
+	drop constraint if exists task_unique_warehouse_id_entity_type_entity_id_queue_name,
+	alter column warehouse_id drop not null,
+	alter column entity_id drop not null,
+	alter column entity_name drop not null,
+	add unique nulls not distinct (project_id, warehouse_id, entity_type, entity_id, queue_name),
+	add check ("entity_type" not in ('table', 'view') or warehouse_id is not null);
```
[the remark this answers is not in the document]
Midor: I tested this by setting up a fresh PostgreSQL database and run all migrations.
The result was, that the old key was not kept and only the new key was existing in the end.

Midor:
@coderabbitai review

crates/lakekeeper/src/api/management/v1/openapi.rs
```diff
@@ -317,6 +334,17 @@ pub fn api_doc<A: Authorizer>(queue_api_configs: &[&QueueApiConfig]) -> utoipa::
...
+    let paths = &mut doc.paths.paths;
+    if paths
+        .remove(ManagementV1Endpoint::SetProjectTaskQueueConfig.path())
+        .is_none()
+    {
+        tracing::warn!(
+            "No path found for SetProjectTaskQueueConfig, not patching queue configs in."
+        );
+        // return;
+    }
```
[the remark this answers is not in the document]
Midor: Will be implemented later

crates/lakekeeper/migrations/20251128101923_enable_tasks_on_project_level.sql
```diff
@@ -0,0 +1,91 @@
...
+	add value if not exists 'project';
+alter type entity_type
+	add value if not exists 'warehouse';
+
+-- 9. Add new values for api_endpoints enum
+ALTER TYPE api_endpoints ADD VALUE 'management-v1-set-project-task-queue-config';
+ALTER TYPE api_endpoints ADD VALUE 'management-v1-get-project-task-queue-config';
+ALTER TYPE api_endpoints ADD VALUE 'management-v1-control-project-tasks';
+ALTER TYPE api_endpoints ADD VALUE 'management-v1-get-project-task-details';
+ALTER TYPE api_endpoints ADD VALUE 'management-v1-list-project-tasks';
```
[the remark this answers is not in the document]
Midor: Should be fixed. Please check again.

crates/lakekeeper/src/api/management/mod.rs
```diff
@@ -1741,6 +1745,128 @@ pub mod v1 {
...
+        )
+    ))]
+    async fn control_project_tasks<C: CatalogStore, A: Authorizer + Clone, S: SecretStore>(
+        Extension(metadata): Extension<RequestMetadata>,
+        AxumState(api_context): AxumState<ApiContext<State<A, C, S>>>,
+        Json(request): Json<ControlTasksRequest>,
+    ) -> Result<StatusCode> {
+        ApiServer::<C, A, S>::control_project_tasks(request, api_context, metadata).await?;
+        Ok(StatusCode::NO_CONTENT)
+    }
```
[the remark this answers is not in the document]
Midor: Should be fixed. Please check again.

crates/lakekeeper/src/api/management/v1/tasks.rs
```diff
@@ -77,13 +81,170 @@ pub struct Task {
...
+        Ok(ProjectListTask {
+            task_id: value.task_id(),
+            project_id: value.project_id().clone(),
+            queue_name: value.queue_name().clone(),
+            status,
+            scheduled_for: value.scheduled_for().ok_or(ErrorModel::internal(
+                "WarehouseListTask needs to be scheduled",
+                "ScheduledForMissing",
+                None,
+            ))?,
```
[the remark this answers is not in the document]
Midor: Should be fixed. Please check again.

crates/lakekeeper/src/implementations/postgres/tasks/get_task_details.rs
```diff
@@ -84,44 +92,65 @@ fn parse_task_details(
...
+        ),
+        EntityType::Project => EntityId::Project,
+        EntityType::Warehouse => {
+            return Err(ErrorModel::internal(
+                "Listing tasks for Project or Warehouse entity types is not yet supported.",
+                "InternalError",
+                None,
+            )
+            .into());
+        }
```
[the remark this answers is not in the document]
Midor: Should be fixed. Please check again.

crates/lakekeeper/src/implementations/postgres/tasks/list_tasks.rs
```diff
@@ -39,35 +44,53 @@ struct TaskRow {
...
+        EntityType::Project => EntityId::Project,
+        EntityType::Warehouse => {
+            return Err(ErrorModel::internal(
+                "Listing tasks for Warehouse entity types is not yet supported.",
+                "InternalError",
+                None,
+            )
+            .into());
+        }
+    };
```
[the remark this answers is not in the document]
Midor: Should be fixed. Please check again.

docs/docs/api/management-open-api.yaml
```diff
@@ -1478,6 +1478,90 @@ paths:
...
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ListTasksResponse'
+        4XX:
+          description: ''
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/IcebergErrorResponse'
   /management/v1/project/{project_id}:
```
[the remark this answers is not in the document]
Midor: Should be fixed. Please check again.

crates/authz-openfga/src/relations.rs
c-thiel, line 1: If we add permissions here, we need to add them to the OpenFGA Model as well, otherwise we will get internal errors on query.

crates/lakekeeper/src/api/management/v1/tasks.rs
```diff
@@ -77,13 +81,170 @@ pub struct Task {
...
+impl TryFrom<TaskInfo> for TaskInfoAPI {
+    type Error = ErrorModel;
+
+    fn try_from(value: TaskInfo) -> Result<Self, Self::Error> {
+        let status = value
+            .status()
+            .copied()
+            .map(TaskStatus::from)
+            .or(value.outcome().copied().map(TaskStatus::from))
+            .ok_or(ErrorModel::internal(
```
c-thiel: nit: for errors it is typically better to use `ok_or_else`. The `internal` constructor of `ErrorModel` allocates storage for this error. This allocation is not needed in the happy case. If we use `ok_or_else` with a closure instead the error path would be slightly slower (one more dispatch due to the closure), while the happy path does not need any extra allocation.

docs/docs/api/management-open-api.yaml
```diff
@@ -6914,6 +7127,86 @@ components:
             view-id:
               type: string
               format: uuid
+    TaskInfoAPI:
```
c-thiel: It is not good that "API" leaks into the type in the API.

docs/docs/api/management-open-api.yaml
```diff
@@ -5775,6 +5999,71 @@ components:
                   enum:
                     - modify
           title: ProjectAssignmentModify
+    ProjectListTask:
```
c-thiel: This type has the same semantic as what is currently called `TaskInfoAPI` on Warehouse level - we should thus give them similar names.

crates/lakekeeper/src/api/management/v1/tasks.rs
```diff
@@ -77,13 +81,170 @@ pub struct Task {
...
+                None,
+            ))?;
+        Ok(TaskInfoAPI {
+            task_id: value.task_id(),
+            project_id: value.project_id().clone(),
+            warehouse_id: value.warehouse_id().ok_or(ErrorModel::internal(
+                "WarehouseListTask needs WarehouseId",
+                "WarehouseIdMissing",
+                None,
+            ))?,
```
c-thiel: This struck me as odd here as the name of type `TaskInfoAPI` doesn't hint me to beeing warehouse specific.

crates/lakekeeper/src/api/management/v1/tasks.rs
```diff
@@ -77,13 +81,170 @@ pub struct Task {
...
+            .ok_or(ErrorModel::internal(
+                "Task must have either status or outcome",
+                "StatusOrOutcomeMissing",
+                None,
+            ))?;
+        Ok(TaskInfoAPI {
+            task_id: value.task_id(),
+            project_id: value.project_id().clone(),
+            warehouse_id: value.warehouse_id().ok_or(ErrorModel::internal(
+                "WarehouseListTask needs WarehouseId",
```
c-thiel: The error message is so specific - are we sure `WarehouseListTask` is the only place taht needs this conversion?

c-thiel: The same conversion is used in:

```rust
impl TryFrom<TaskDetails> for GetProjectTaskDetailsResponse {
    type Error = ErrorModel;

    fn try_from(value: TaskDetails) -> Result<Self, Self::Error> {
        Ok(Self {
            task: ProjectTaskInfoAPI::try_from(value.task)?,
            task_data: value.data,
            execution_details: value.execution_details,
            attempts: value.attempts,
        })
    }
}
```
So the error message should not hint to the Endpoint.

crates/lakekeeper/src/api/management/v1/tasks.rs
```diff
@@ -77,13 +81,170 @@ pub struct Task {
...
+                "WarehouseListTask needs entity_name",
+                "EntityNameMissing",
+                None,
+            ))?,
+            status,
+            scheduled_for: value.scheduled_for().ok_or(ErrorModel::internal(
+                "WarehouseListTask needs to be scheduled",
+                "ScheduledForMissing",
+                None,
+            ))?,
```
c-thiel: This is not correct. `scheduled_for` / `attempt_scheduled_for` are both required fields (in the DB and semantically), so this conversion here must be obsolete.

crates/lakekeeper/migrations/20251128101923_enable_tasks_on_project_level.sql
```diff
@@ -0,0 +1,96 @@
...
+	add primary key (task_config_id);
+
+-- 5. Modify task constraints and make warehouse_id optional
+alter table task 
+	drop constraint if exists task_unique_warehouse_id_entity_type_entity_id_queue_name,
+	alter column warehouse_id drop not null,
+	alter column entity_id drop not null,
+	alter column entity_name drop not null,
+	add unique nulls not distinct (project_id, warehouse_id, entity_type, entity_id, queue_name),
+	add check ("entity_type" not in ('table', 'view') or warehouse_id is not null);
```
c-thiel: We prepare this only for `entity_type`: `table` and `views`. Yet just a few lines below we introduce warehouse tasks, which would also require a warehouse id. We should include `warehouse` here too.

crates/lakekeeper/migrations/20251128101923_enable_tasks_on_project_level.sql
```diff
@@ -0,0 +1,96 @@
...
+
+-- 6. Modify task_config constraints and make warehouse_id optional
+alter table task_config
+	alter column warehouse_id drop not null,
+	add unique nulls not distinct (project_id, warehouse_id, queue_name);
```
c-thiel: Can't we just merge this with the previous alter table statement on `task_config`?

crates/lakekeeper/migrations/20251128101923_enable_tasks_on_project_level.sql
```diff
@@ -0,0 +1,96 @@
...
+	add column if not exists task_config_id uuid default gen_random_uuid(),
+	add primary key (task_config_id);
+
+-- 5. Modify task constraints and make warehouse_id optional
+alter table task 
+	drop constraint if exists task_unique_warehouse_id_entity_type_entity_id_queue_name,
+	alter column warehouse_id drop not null,
+	alter column entity_id drop not null,
+	alter column entity_name drop not null,
+	add unique nulls not distinct (project_id, warehouse_id, entity_type, entity_id, queue_name),
```
c-thiel: For more complex constraint we should specify names. This has two advantages:
- Executing it twice fails (currently it does not, it would just get a new name)
- We don't rely on consisten postgres name generation if we ever want to change this constraint.

crates/lakekeeper/migrations/20251128101923_enable_tasks_on_project_level.sql
```diff
@@ -0,0 +1,96 @@
...
+	alter column warehouse_id drop not null,
+	add unique nulls not distinct (project_id, warehouse_id, queue_name);
+
+-- 7. Modify task_log constraints and make warehouse_id optional
+alter table task_log
+	alter column warehouse_id drop not null,
+	alter column entity_id drop not null,
+	alter column entity_name drop not null,
+	add check ("entity_type" not in ('table', 'view') or warehouse_id is not null);
+
```
c-thiel: We are missing all the new indexes (or extensions of existing indexes) to make polling on project tasks feasable.

crates/lakekeeper/migrations/20251128101923_enable_tasks_on_project_level.sql
```diff
@@ -0,0 +1,96 @@
...
+alter table task_config
+	alter column warehouse_id drop not null,
+	add unique nulls not distinct (project_id, warehouse_id, queue_name);
+
+-- 7. Modify task_log constraints and make warehouse_id optional
+alter table task_log
+	alter column warehouse_id drop not null,
+	alter column entity_id drop not null,
+	alter column entity_name drop not null,
+	add check ("entity_type" not in ('table', 'view') or warehouse_id is not null);
```
c-thiel: This constraint would just be called `task_check` by default, which has a very high likelihood of collision with future checks. We should give it an explicit name.

crates/lakekeeper/migrations/20251128101923_enable_tasks_on_project_level.sql
c-thiel: Shouldn't we add this constraint to task_log as well?

crates/lakekeeper/src/service/tasks/mod.rs
```diff
@@ -208,10 +273,11 @@ pub struct TaskInput {
 /// Metadata stored for each task in the database backend.
 /// This is separate from the task payload, which is specific to each task type.
 pub struct TaskMetadata {
-    pub warehouse_id: WarehouseId,
+    pub warehouse_id: Option<WarehouseId>,
+    pub project_id: ProjectId,
     pub parent_task_id: Option<TaskId>,
     pub entity_id: EntityId,
-    pub entity_name: Vec<String>,
+    pub entity_name: Option<Vec<String>>,
```
c-thiel: As all entities in Lakekeeper are named, I don't think this should be optional. Think about the user: We return this as part of a list endpoint. Without an entity_name, the user just sees IDs. This is unfriendly and requires the user or the UI to run many more queries just to refresh all the machine readable IDs.

crates/lakekeeper/src/service/tasks/mod.rs
```diff
@@ -96,14 +100,18 @@ pub enum TaskEntity {
 pub enum TaskEntityNamed {
     Table(TableNamed),
     View(ViewNamed),
+    Project(ProjectId),
+    Warehouse(WarehouseId),
```
c-thiel: This is called `Named` yet we don't have names for Project & Warehouse. This is quite unexpected. Lets try to find a better name.

crates/lakekeeper/src/api/management/v1/tasks.rs
```diff
@@ -77,13 +81,170 @@ pub struct Task {
...
+                EntityId::View(view_id) => TaskEntity::View { view_id: *view_id },
+                _ => {
+                    return Err(ErrorModel::internal(
+                        "Only Table and View entities are supported in WarehouseListTask",
+                        "UnsupportedEntityType",
+                        None,
+                    ));
+                }
+            },
+            entity_name: value.entity_name().cloned().ok_or(ErrorModel::internal(
```
c-thiel: nit: We have a few unnecessary clones here. This method takes ownership of the original struct. This allows us to just move the ownership of the original object, which is more efficient than creating duplicates and right after dropping the original version.

crates/lakekeeper/src/api/management/v1/tasks.rs
```diff
@@ -77,13 +81,170 @@ pub struct Task {
...
+        let status = value
+            .status()
+            .copied()
+            .map(TaskStatus::from)
+            .or(value.outcome().copied().map(TaskStatus::from))
+            .ok_or(ErrorModel::internal(
+                "Task must have either status or outcome",
+                "StatusOrOutcomeMissing",
+                None,
+            ))?;
```
c-thiel: We should make this typesafe. It is the responsibility of the DB to return either one or the other. Currently this information is implicit "in the docstring", which leaves room for human error to return neither or both. Because of the current data structure, this case would be detected at runtime and exposed to the user as an internal error. Instead, lets make it typesafe so that this error is detected at compile time.

crates/lakekeeper/src/service/tasks/mod.rs
```diff
@@ -277,6 +352,102 @@ pub struct Task {
...
+        &self.queue_name
+    }
+
+    #[must_use]
+    pub fn attempt(&self) -> i32 {
+        self.id.attempt
+    }
+
+    #[must_use]
+    pub fn status(&self) -> Option<&TaskStatus> {
```
c-thiel: Types that implement `Copy` should be returned owned.

crates/lakekeeper/migrations/20251128101923_enable_tasks_on_project_level.sql
c-thiel: This migration leaves room for data corruption due to missing checks.
We need this:


entity_type | warehouse_id | entity_id | entity_name
-- | -- | -- | --
project | NULL | NULL | NULL
warehouse | NOT NULL | NULL | NULL
table | NOT NULL | NOT NULL | NOT NULL
view | NOT NULL | NOT NULL | NOT NULL

<br class="Apple-interchange-newline">

With the current approach also this would be possible:

entity_type | warehouse_id | entity_id | entity_name
-- | -- | -- | --
project | NULL | NULL | NULL
project | NULL | NOT NULL | NULL
project | NULL | NOT NULL | NOT NULL
project | NULL | NULL | NOT NULL

So the DB schema leaves amiguity to represent a project task, enabling the creation of duplicates.
The DB schema should be strict and never allow us to insert corrupt data.

crates/lakekeeper/src/api/management/v1/tasks.rs
```diff
@@ -727,16 +1163,18 @@ async fn authorize_control_tasks<A: Authorizer, C: CatalogStore>(
 ) -> Result<()> {
     let required_tabular_ids = tasks
         .iter()
-        .map(|t| match &t.entity {
-            TaskEntityNamed::Table(tabular) => TabularId::Table(tabular.table_id),
-            TaskEntityNamed::View(tabular) => TabularId::View(tabular.view_id),
+        .filter_map(|t| match &t.entity {
+            TaskEntityNamed::Table(tabular) => Some(TabularId::Table(tabular.table_id)),
+            TaskEntityNamed::View(tabular) => Some(TabularId::View(tabular.view_id)),
+            _ => None,
```
c-thiel: This introduces a security issue for Warehouse-level tasks: We already know at this point that the user cannot control all tasks.
Yet, we just filter out all warehouse level tasks here, never performing any AuthZ check for them.

crates/lakekeeper/src/service/tasks/tabular_expiration_queue.rs
```diff
@@ -89,17 +89,31 @@ pub(crate) async fn tabular_expiration_worker<C: CatalogStore, A: Authorizer>(
...
+        } else {
+            tracing::debug_span!(
+                QN_STR,
+                entity_type = %entity_type,
+                entity_id = %entity_id_uuid,
+                deletion_kind = ?task.data.deletion_kind,
+                attempt = %task.attempt(),
+                task_id = %task.task_id(),
+            )
+        };
```
c-thiel: Is it worth creating a span for this invalid case?

crates/lakekeeper/src/service/tasks/tabular_expiration_queue.rs
```diff
@@ -115,15 +129,17 @@ async fn instrumented_expire<C: CatalogStore, A: Authorizer>(
     let entity_id = task.task_metadata.entity_id;
     match handle_table::<C, A>(catalog_state.clone(), authorizer, task).await {
         Ok(()) => {
-            tracing::debug!("Task of `{QN_STR}` worker exited successfully. {entity_id} deleted.");
+            tracing::debug!(
+                "Task of `{QN_STR}` worker exited successfully. {entity_id:?} deleted."
```
c-thiel: Don't use debug prints in logs / traces eccept if you know there is a good representation (i.e. Vec is OK).
This option struct is not readable.

crates/lakekeeper/src/service/tasks/mod.rs
```diff
@@ -277,6 +352,102 @@ pub struct Task {
     pub(crate) data: serde_json::Value,
 }
 
+#[derive(Debug, Clone, PartialEq)]
+pub struct TaskInfo {
+    pub task_metadata: TaskMetadata,
+    pub queue_name: TaskQueueName,
+    pub id: TaskAttemptId,
+    pub status: Option<TaskStatus>,
+    pub outcome: Option<TaskOutcome>,
```
c-thiel: This introduces unneccessary ambiguity. It allows for a struct with no field set or both. Semantically, they are mutually exclusive and required, so we should encode this in a type.

crates/lakekeeper/src/implementations/postgres/tasks/list_tasks.rs
```diff
@@ -39,35 +44,46 @@ struct TaskRow {
...
+        EntityType::Warehouse => EntityId::Warehouse,
+    };
+    Ok(TaskInfo {
+        id: TaskAttemptId {
+            task_id: row.task_id.into(),
+            attempt: row.attempt,
+        },
+        queue_name: row.queue_name.into(),
+        task_metadata: TaskMetadata {
+            project_id: ProjectId::from(row.project_id.clone()),
```
c-thiel: I don't think clone is needed here.

crates/lakekeeper/src/api/management/v1/tasks.rs
```diff
@@ -77,13 +81,170 @@ pub struct Task {
...
+        Ok(ProjectListTask {
+            task_id: value.task_id(),
+            project_id: value.project_id().clone(),
+            queue_name: value.queue_name().clone(),
+            status,
+            scheduled_for: value.scheduled_for().ok_or(ErrorModel::internal(
+                "ProjectListTask needs to be scheduled",
+                "ScheduledForMissing",
+                None,
+            ))?,
```
c-thiel: Same as for `TaskInfo`. This should be required.

crates/lakekeeper/src/api/management/v1/tasks.rs
```diff
@@ -77,13 +81,170 @@ pub struct Task {
     pub updated_at: Option<chrono::DateTime<chrono::Utc>>,
 }
 
+impl TryFrom<TaskInfo> for TaskInfoAPI {
+    type Error = ErrorModel;
+
+    fn try_from(value: TaskInfo) -> Result<Self, Self::Error> {
```
c-thiel, line 88: This conversion feels very hacky. We already have a valid internal `TaskInfo` object and yet we manage to fail in 4 different ways on the last mile when converting to an API Object. I would expect a single error: Either the internal task has the correct scope / is a warehouse task or not.

crates/lakekeeper/src/service/authz/mod.rs
```diff
@@ -214,6 +214,10 @@ pub enum CatalogProjectAction {
     ListRoles,
     SearchRoles,
     GetEndpointStatistics,
+    ModifyTaskQueueConfig,
+    GetTaskQueueConfig,
+    GetAllTasks,
+    ControlAllTasks,
```
c-thiel: We should make these actions more specific (GetProjectTask) as otherwise we would have a breaking change when we extend the endpoint to return more then just project tasks:
1. If we would just return the additional tasks, we would have a Semantic breaking change as different entities are returned for the same call
2. If we would not return the additional tasks, we would need to introduce an new permission, one for "only project", and one for all.

In general its a good idea to start with as fine-grained permission as possible.

crates/lakekeeper/src/service/tasks/mod.rs
c-thiel: This doesn't feel right. If entity_id is known, then also an entity_name should be known. In Lakekeeper all entities are named, so we should know either both or none.

crates/lakekeeper/src/service/tasks/mod.rs
```diff
@@ -220,19 +286,25 @@ pub struct TaskMetadata {
...
 }
 
-#[derive(Debug, Copy, Clone, PartialEq, derive_more::From)]
+#[derive(Debug, Clone, Copy, PartialEq, derive_more::From)]
 pub enum EntityId {
     Table(TableId),
     View(ViewId),
+    Project,
+    Warehouse,
 }
```
c-thiel: Maybe we should call this `TaskEntity` or so? Not all variants have IDs

c-thiel:
@coderabbit full review.
Resolve discussions if issue is fixed

crates/lakekeeper/src/implementations/postgres/tasks/list_tasks.rs
```diff
@@ -109,6 +110,22 @@ pub(crate) async fn list_tasks(
...
+        TaskFilter::ProjectId {
+            project_id,
+            include_sub_tasks,
+        } => (None, project_id, *include_sub_tasks),
+        TaskFilter::TaskIds(_) => Err(ErrorModel::internal(
+            "TaskFilter for TaskIds not implemented for list_tasks.",
+            "InternalError",
+            None,
+        ))?,
+    };
```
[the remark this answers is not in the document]
c-thiel: unreachable for now, impl later

crates/lakekeeper/src/service/identifier/project.rs
```diff
@@ -24,6 +24,12 @@ impl From<ProjectId> for String {
     }
 }
 
+impl From<String> for ProjectId {
+    fn from(s: String) -> Self {
+        ProjectId::from_db_unchecked(s)
+    }
+}
+
```
c-thiel: We need to remove this as it bypasses validation. `from_db_unchecked` is much clearer.
This is an unsafe constructor that must not be used accidentially.

c-thiel:
@coderabbit review & resolve

c-thiel:
@coderabbit review

Issue #1566, opened 2026-01-12
when enabled openfga, `show all tables` is very slow(use a low-privilege account)


yoogoc:
when enabled openfga, `show all tables` is very slow(use a low-privilege account), seem `BulkCheck` in `list_entities` is very very high latency, I think can use `ListObjects` to alternative
example: https://github.com/yoogoc/lakekeeper/tree/slow-list-entity/examples/access-control-simple

c-thiel:
@yoogoc this might also already help:
https://github.com/lakekeeper/lakekeeper/pull/1617

c-thiel:
Hey @yoogoc, we where missing indices before which where required for listing. #1617  added those.
We see significant improvement now in listing performance. I am closing this issue for now, please re-open if the issue persists for you!

yoogoc:
@c-thiel Thanks! I'll give it a try and get back to you.

yoogoc:
> Hey [@yoogoc](https://github.com/yoogoc), we where missing indices before which where required for listing. [#1617](https://github.com/lakekeeper/lakekeeper/pull/1617) added those. We see significant improvement now in listing performance. I am closing this issue for now, please re-open if the issue persists for you!

<img width="1173" height="353" alt="Image" src="https://github.com/user-attachments/assets/16beba86-eaff-4ac1-89b2-e509c1b2ff7d" />

@c-thiel Thanks for the update! However, I noticed that the PR seems to be focused on roles — based on my test cases, the performance issue doesn't appear to have improved on my end. Could you take another look?

PR #1572, opened 2026-01-19
refactor list_entities and list_namespaces

yoogoc:
https://github.com/lakekeeper/lakekeeper/issues/1566

## Summary by CodeRabbit

* **New Features**
  * Added ability to query allowed tables, views, and namespaces for the current user, enabling better access control visibility.
  * Implemented optimized permission checking for listing operations using fast-path evaluation and result masking.

* **Documentation**
  * Updated example notebooks and configurations with new access control scenarios and DuckDB workflow demonstrations.

* **Chores**
  * Enhanced OpenFGA configuration with timeout and cache settings in example environments.

### DIFF
--- a/crates/authz-openfga/src/authorizer.rs
+++ b/crates/authz-openfga/src/authorizer.rs
@@ -15,8 +15,8 @@ use lakekeeper::{
         TableId, UserId, ViewId,
         authz::{
             AuthorizationBackendUnavailable, Authorizer, CannotInspectPermissions,
-            CatalogProjectAction, CatalogUserAction, IsAllowedActionError, ListProjectsResponse,
-            NamespaceParent, UserOrRole,
+            CatalogProjectAction, CatalogUserAction, IsAllowedActionError, ListAllowedEntitiesResponse,
+            ListProjectsResponse, NamespaceParent, UserOrRole,
         },
         health::Health,
     },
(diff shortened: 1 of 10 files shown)


c-thiel:
@coderabbit full review

c-thiel:
Thanks for the PR @yoogoc. I agree that for OpenFGA it would be faster to just use it for listing, especially if many items are filtered out.
Of the options that are available for OpenFGA (https://openfga.dev/docs/interacting/search-with-permissions#possible-options), we currently implement `Option 1: Search, then check`, while this PR implements `Option 3: Build a list of IDs, then search`.

My main concern is that not all Authorizers implement these kinds of List endpoints. To stay widely compatible with our `Authorizer` interface, we should make the use of Authorizer list endpoints optional and fall-back to what we have implemented currently. One example of such an Authorizer would be Cedar / Amazon Verified Permissions that does not support this query but folks are using today with Lakekeeper.

yoogoc:
@c-thiel hi, I added a fallback mechanism: If list_allowed_[namespace/table/views]s is not implemented by the Authorizer, the legacy flow will be executed.

c-thiel:
Improved indexes for the current approach are here:
https://github.com/lakekeeper/lakekeeper/pull/1617

c-thiel:
Closing this as I believe https://github.com/lakekeeper/lakekeeper/pull/1617 should be sufficient for now.
If not please re-open an issue and we can re-evaluate adding this separate codepath.

PR #1605, opened 2026-02-10
feat: Add FileInfo to `LakekeeperStorage::list`

Midor:
## Summary by CodeRabbit

* **New Features**
  * Listings now return richer file entries that include last_modified timestamps alongside locations across all storage backends.
* **Bug Fixes / Behavior**
  * Read/write/delete semantics preserved; batch deletions and listing pagination now operate on the richer file entries without changing user-facing behavior.
* **Tests**
  * Integration tests updated to use and validate listing results that include timestamps.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3218,7 +3218,7 @@ dependencies = [
 [[package]]
 name = "iceberg"
 version = "0.8.0"
-source = "git+https://github.com/lakekeeper/iceberg-rust.git?rev=acda05d67eb0bb304ba2f86e135c9ecb6caf793e#acda05d67eb0bb304ba2f86e135c9ecb6caf793e"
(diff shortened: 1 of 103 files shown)


Midor:
@coderabbitai review. Resolve if solved

c-thiel:
@coderabbitai review

crates/io/src/adls/adls_storage.rs
```diff
@@ -459,6 +446,30 @@ impl LakekeeperStorage for AdlsStorage {
...
+        };
+        let full_path = format!(
+            "{}://{}/{}",
+            base_location.scheme(),
+            base_location.authority_with_host(),
+            path_name
+        );
+        let location = Location::from_str(&full_path).ok()?;
+        let last_modified = path.last_modified;
+        let last_modified = DateTime::from_timestamp(last_modified.unix_timestamp(), 0)?;
```
c-thiel: This should not return none if we have a valid Path but no last_modified. Instead we should make last_modified Optional in FileInfo. Also for the other cloud providers - is it really worth to always fail just if the modified time cannot be extracted?

Midor:
@coderabbitai review and resolve if fixed

PR #1615, opened 2026-02-17
feat: Customization Option for Storage Layout

Midor:
## Summary by CodeRabbit

* **New Features**
  * Configurable storage layout for S3/GCS/ADLS with templated namespace/table paths (flat, parent, full) using {uuid} and {name} (percent-encoded).

* **Improvements**
  * Hierarchical default namespace/table location resolution and layout-aware path rendering.
  * Improved S3 signing and URL/path handling.

* **Documentation**
  * OpenAPI and user docs updated with storage-layout schemas and examples.

* **Tests**
  * Added/updated tests including special-character name handling and updated fixtures.

### DIFF
--- a/docs/docs/storage.md
+++ b/docs/docs/storage.md
@@ -62,7 +62,22 @@ With the default `{uuid}`-only templates this looks like:
 s3://my-bucket/warehouse/<uuid-of-production-namespace>/<uuid-of-table>/
 ```
 
-To use this layout explicitly with custom templates:
+To use the default layout explicitly:
+
+```json
+{
+  "storage-profile": {
+    "type": "s3",
+    "storage-layout": {
+      "type": "parent-namespace-and-table",
+      "namespace": "{uuid}",
+      "table": "{uuid}"
+    }
+  }
+}
+```
+
+or for specifying custom templates:
 
 ```json
 {


crates/lakekeeper/src/service/storage/gcs/mod.rs
```diff
@@ -67,6 +68,9 @@ pub struct GcsProfile {
     /// Defaults to true.
     #[serde(default = "default_true")]
     pub sts_enabled: bool,
+    /// The layout to use for namespaces and tables stored in this storage profile. If not set, the default layout is `StorageLayout::Flat(TableNameRenderer("{uuid}".to_string()))`, which does not include the namespace in the path.
```
c-thiel: This docstring shows up in OpenAPI, it should not contain rust types. I think we should just paste the equivalent JSON representation here.

crates/lakekeeper/src/service/storage/storage_layout.rs
```diff
@@ -0,0 +1,523 @@
...
+}
+
+#[derive(Debug, Clone, Hash, Eq, PartialEq, Serialize, Deserialize)]
+#[cfg_attr(feature = "open-api", derive(utoipa::ToSchema))]
+#[serde(rename_all = "kebab-case")]
+pub enum StorageLayout {
+    Flat(TableNameRenderer),
+    Parent(ParentLayout),
+    Full(FullLayout),
+}
```
c-thiel, line 60: I think the JSON structure would be cleaner if we would use `serde(tag = "type", rename_all ....)`.

This also introduces five new types to the OpenAPI spec. They currently don't have speeking names even though they are in the global OpenAPI namespace.
I think we should give them all the same prefix and make them more speaking:
* `StorageLayout`
* `StorageLayoutTableTemplate`
* `StorageLayoutNamespaceTemplate`
* `StorageLayoutFlat` (doesn't exist yet)
* `StorageLayoutFullHierarchy`
* `StorageLayoutTableAndParentNamespace`
(or similar - my point is just that names should be self-explanatory and include the context if they only make sense in this one context).

I think we should serialize the Flat layout also as a struct, so that the table template is always in a `table` field, making it more consistent across variants.

Renderer is quite technical term, in the OpenAPI I think template would be a better name.

crates/lakekeeper/src/service/storage/storage_layout.rs
c-thiel, line 1: Nit: After finalizing the serialization of the StorageLayout, I think we should add a few deserialization tests to ensure that a unittest triggers if we accidentally break deserialization in a non-backward compatible way.

c-thiel (commented):
Just a few nits, LGTM otherwise!

crates/lakekeeper/src/service/storage/s3.rs
```diff
@@ -51,6 +51,7 @@ use crate::{
                 CredentialsError, IcebergFileIoError, InvalidProfileError, TableConfigError,
                 UpdateError, ValidationError,
             },
+            storage_layout::StorageLayout,
         },
```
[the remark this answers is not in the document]
c-thiel: @coderabbitai we want a way to reset to default.
@Midor how about adding a third variant to the enum that is the current default? I haven't thought it through completly.

c-thiel:
Another thought: I think we might want to clean identifiers. We probably don't want issues with spaces or other special characters. I am not sure if maybe urlencoding solves this problem, but we should certainly test it.
I think especially for this complex case we should also add an integration test.

We already have the test: test_special_characters_in_names in pytest. Maybe we should just default to the full hierarchy template with idents & ids when running it?

crates/lakekeeper/src/service/storage/s3.rs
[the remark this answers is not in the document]
Midor: I'd also lean towards **Option 1**. I take a look into it.

Midor:
@coderabbitai review and resolve only if fixed

docs/docs/api/management-open-api.yaml
```diff
@@ -3574,6 +3574,11 @@ components:
             - string
             - 'null'
           description: Subpath in the filesystem to use.
+        layout:
```
c-thiel: I think we want to call this here and in the other profiles too "storage_layout" to match the type and be more expressive. Layout could otherwise also reference default partitioning layout or so.

c-thiel:
@coderabbitai full review
Escaping paths is no longer neccessary. Proper encoding is used everywhere, shown by integration tests.

c-thiel:
@coderabbitai approve if all issues resolved

c-thiel (dismissed):

crates/lakekeeper/src/service/storage/storage_layout.rs
```diff
@@ -0,0 +1,918 @@
...
+pub struct StorageLayoutFullHierarchy {
+    pub namespace: StorageLayoutNamespaceTemplate,
+    pub table: StorageLayoutTableTemplate,
+}
+
+/// No namespace directories; all tables are placed directly under the base location.
+///
+/// For a table `my_table` (uuid `…002`) the path is: `<base>/<table-segment>`.
+/// The table template must contain `{uuid}` to avoid collisions between tables with the same name.
+#[derive(Debug, Clone, Hash, Eq, PartialEq, Serialize, Deserialize)]
```
c-thiel: Deserialization avoids constructor.

crates/lakekeeper/src/service/storage/storage_layout.rs
```diff
@@ -0,0 +1,918 @@
...
+
+impl StorageLayout {
+    pub fn try_new_flat(table_template: String) -> Result<Self, StorageLayoutError> {
+        if !table_template.contains("{uuid}") {
+            return Err(StorageLayoutError::InvalidTemplate(format!(
+                "For the 'Flat' layout, the table template '{table_template}' must contain the {{uuid}} placeholder to prevent path collisions."
+            )));
+        }
+        Ok(Self::Flat(StorageLayoutFlat {
+            table: StorageLayoutTableTemplate(table_template),
```
c-thiel: Failure logic should be on sub-type

crates/lakekeeper/src/service/storage/storage_layout.rs
```diff
@@ -0,0 +1,918 @@
...
+    pub fn try_new_parent(
+        namespace_template: String,
+        table_template: String,
+    ) -> Result<Self, StorageLayoutError> {
+        Ok(Self::Parent(StorageLayoutParentNamespaceAndTable {
+            namespace: StorageLayoutNamespaceTemplate(namespace_template),
+            table: StorageLayoutTableTemplate(table_template),
+        }))
+    }
+
```
c-thiel, line 154: Why fallible?

crates/lakekeeper/src/service/storage/storage_layout.rs
```diff
@@ -0,0 +1,918 @@
...
+            namespace_path_rendered,
+            vec![format!("{}", parent_namespace.uuid),]
+        );
+
+        let table_name_rendered = layout.render_table_segment(&table);
+        assert_eq!(table_name_rendered, format!("{}", table.name));
+    }
+
+    #[test]
+    fn test_storage_layout_deserialization_of_flat_layout() {
```
c-thiel, line 753: This test is wrong and must fail. Bypasses constructor.

c-thiel:
@coderabbitai full review

c-thiel:
@coderabbitai review

crates/lakekeeper/src/server/tabular.rs
```diff
@@ -21,23 +26,35 @@ pub(super) fn determine_tabular_location(
...
+                .default_namespace_location(&namespace_path)
                 .map_err(|e| {
                     ErrorModel::internal(
                         "Failed to generate default namespace location",
                         "InvalidDefaultNamespaceLocation",
                         Some(Box::new(e)),
                     )
-                })?,
+                })?
+        };
```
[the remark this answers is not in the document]
c-thiel: docs added

docs/docs/storage.md
```diff
@@ -34,6 +33,130 @@ For S3 / AWS and Azure / ADLS Warehouses, Lakekeeper optionally supports additio
...
+{
+  "storage-profile": {
+    "type": "s3",
+    "storage-layout": {
+      "type": "parent-namespace-and-table",
+      "namespace": "{name}-{uuid}",
+      "table": "{name}-{uuid}"
+    }
+  }
+}
```
Midor, line 92: Shouldn't this be only with `{uuid}` instead of `{name}-{uuid}` as it should represent the default layout?

PR #1673, opened 2026-03-22
feat: Generic Table API with Lance

fresh-borzoni:
This adds Generic Table API  - a way to register non-Iceberg table formats like Lance or Delta in Lakekeeper and get STS credential vending without faking Iceberg metadata.                                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                     
Right now if you want Lance tables governed by Lakekeeper, you have to create dummy Iceberg tables with fake schemas and stash the real location in properties. This works but it's ugly. The Generic Table API gives these formats first-class registration: you POST a name, format, and base-location, and GET back vended storage credentials just like Iceberg tables do. 

<!-- This is an auto-generated comment: release notes by coderabbit.ai -->
## Summary by CodeRabbit

* **New Features**
  * Generic Table API: create/list/load/rename/drop generic (non-Iceberg) tables with credential vending, soft-delete/undrop, protection, and force-delete.
  * Management endpoints and OpenAPI for generic-table actions & protection.
  * Event publishing for generic-table lifecycle and CLI OpenAPI generator.

* **Authorization**
  * Full permission model and management operations for generic tables (OpenFGA integration).

* **Tests**
  * End-to-end and integration tests covering create/load/list/credentials/rename/drop and task/expiration flows.

* **Documentation**
  * User docs and OpenAPI spec added for Generic Tables.

<!-- review_stack_entry_start -->

[![Review Change Stack](https://storage.googleapis.com/coderabbit_public_assets/review-stack-in-coderabbit-ui.svg)](https://app.coderabbit.ai/change-stack/lakekeeper/lakekeeper/pull/1673?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

<!-- review_stack_entry_end -->
<!-- end of auto-generated comment: release notes by coderabbit.ai -->


crates/lakekeeper/migrations/20260220000000_add_generic_table.sql
c-thiel, line 1: Just my first thoughts:

Should we try to re-use the `tabular` table - similar as we do it for `table` and `view`?
This would help us in a few ways:
- Ensure that we have a globally unique ID that doesn't overlap
- Ensure that the name doesn't collide with table or view names
- Re-Use indices
- Re-Use search

If we would use the `tabular` table we might also be able to re-use the existing Properties table, which would allow us to easily find tables by properties.

For updated_at and also for potential future caches we might want to consider using:
`SELECT trigger_updated_at_and_version_if_distinct('"generic_table"');` instead of plain `updated_at`. This also adds a trigger to update those fields.

I am also wondering if we might want to make format an enum in rust - this could have a single variant to start with: Unknown(String). But it would be an extension point for well-known formats.

c-thiel (commented):
Thanks for the PR @fresh-borzoni!
Just my first thoughts down below. I'll try to find some time next week to give it a bit more in-depth thoughts :)

crates/lakekeeper/migrations/20260220000000_add_generic_table.sql
fresh-borzoni: @c-thiel  TY for the thoughts.

Agreed. Tabular integration is planned (there's a TODO for TabularId::GenericTable), I kept it out for the PoC intentionally, but it's mechanical to add if the direction is good. So, we may experiment with this, define the shape and then create umbrella issue to iterate on this.

c-thiel:
@fresh-borzoni are there currently open points from your side?
Planning to give it a more thorough look early next week. If you have points you want to address I don't have to focus on those.

fresh-borzoni:
@c-thiel Thanks for the heads-up. 
I’m planning to add the authz wiring and test everything end-to-end, and I should have it finished before next week

c-thiel:
@fresh-borzoni I see you are still actively developing - let me know when this is ready for review!

fresh-borzoni:
@coderabbitai review

fresh-borzoni:
@coderabbitai review

fresh-borzoni:
@coderabbitai review

fresh-borzoni:
@coderabbitai review

fresh-borzoni:
@c-thiel I think you can review now

authz/openfga/v4.3/store.fga.yaml
c-thiel, line 1: We can't modify existing OpenFGA Model versions - we must create a new Version v4.5 which then has these new components.
Just copy the v4.4 folder into v4.5 and then modify `authz-openfga/src/migration.rs` to use this new version.

crates/lakekeeper/migrations/20260220000001_generic_table_check_constraint.sql
c-thiel, line 1: All these migrations belong to one feature. Each extra file adds a _sqlx_migrations row and a checksum verified on every startup — strictly additive. sqlx migrations are immutable once shipped, so the count we land with is permanent. Keep the footprint small and put everything in a single migration.

Lakekeeper applies all migrations inside one outer transaction (each file is forced to no_tx), so we can't reference newly-added enum values as typed literals anyway — splitting buys us nothing on that front. The single transaction is intentional: it prevents partial migrations that would otherwise leave the DB in a corrupt state.

crates/lakekeeper/migrations/20260220000000_add_generic_table.sql
```diff
@@ -0,0 +1,32 @@
...
+    version          BIGINT NOT NULL DEFAULT 0,
+    created_at       TIMESTAMPTZ NOT NULL DEFAULT now(),
+    updated_at       TIMESTAMPTZ,
+    PRIMARY KEY (warehouse_id, generic_table_id),
+    FOREIGN KEY (warehouse_id, generic_table_id)
+        REFERENCES tabular(warehouse_id, tabular_id) ON DELETE CASCADE
+);
+SELECT trigger_updated_at_and_version_if_distinct('"generic_table"');
+
+CREATE TABLE generic_table_properties (
```
c-thiel: I think we should add the created & updated triggers for this table too - similar to the `table_properties` table. WDYT?

crates/lakekeeper/migrations/20260220000000_add_generic_table.sql
```diff
@@ -0,0 +1,32 @@
+ALTER TYPE tabular_type ADD VALUE IF NOT EXISTS 'generic-table';
```
c-thiel: When a Postgres enum is used as a typed literal inside the same transaction it's mutated in (CHECK constraints, equality comparisons, etc.), ALTER TYPE ... ADD VALUE doesn't work — PG forbids referencing the new value until the enclosing tx commits. Since Lakekeeper applies all migrations in a single shared transaction, this restriction spans the whole batch, not just the current file.

The pattern is to recreate the type rather than alter it: RENAME the existing type to <name>_old, CREATE TYPE <name> with the full new variant set, ALTER TABLE ... ALTER COLUMN ... TYPE <name> USING col::text::<name> for every column that uses it, then DROP TYPE <name>_old. The new type is a fresh object in this transaction, so its values are usable as literals immediately.

See 20251228101923_enable_tasks_on_project_level.sql for prior art (added project, warehouse, etc. to entity_type via this exact dance, then used them in fresh CHECK constraints in the same migration).

For generic-table, the same recreate-the-type approach lets us keep the whitelist form of the CHECK constraints — no ::text casts, no split files, no weakened typ <> 'view' blacklist.

crates/lakekeeper/migrations/20260220000000_add_generic_table.sql
```diff
@@ -0,0 +1,32 @@
...
+
+CREATE TABLE generic_table (
+    warehouse_id     UUID NOT NULL,
+    generic_table_id UUID NOT NULL,
+    format           TEXT NOT NULL,
+    doc              TEXT,
+    schema_info      JSONB,
+    statistics       JSONB,
```
c-thiel: Should we bound those to prevent killing postgres with errornous blobs?

crates/lakekeeper/src/implementations/postgres/tabular/mod.rs
```diff
@@ -897,6 +987,7 @@ pub(crate) async fn search_tabular<'e, 'c: 'e, E: sqlx::Executor<'c, Database =
                     AND w.status = 'active'
                     AND t.deleted_at IS NULL
                     AND t.metadata_location IS NOT NULL
+                    AND t.typ != 'generic-table'
```
c-thiel: Why not include generic tables in the search?

crates/lakekeeper/src/implementations/postgres/tabular/generic_table.rs
```diff
@@ -0,0 +1,605 @@
...
+        entries.push(GenericTableListEntry {
+            generic_table_id: row.generic_table_id.into(),
+            warehouse_id: row.warehouse_id.into(),
+            namespace_id: row.namespace_id.into(),
+            name: row.name.clone(),
+            tabular_ident,
+            format: GenericTableFormat::from(row.format.clone()),
+            namespace_ident: namespace_ident.clone(),
+            protected: row.protected,
+            properties: HashMap::new(),
```
c-thiel: If we don't want to include properties in lists we should remove the field. This suggests it is present when its actually always empty.

crates/lakekeeper/src/api/router.rs
```diff
@@ -127,6 +132,7 @@ pub async fn new_full_router<
     let mut router = Router::new()
         .nest("/catalog/v1", v1_routes)
         .nest("/management/v1", management_routes)
+        .nest("/v1", generic_table_routes)
```
c-thiel: I think the /v1 prefix is too general - I think we should use `/lakekeeper/v1/` as this is a lakekeeper specific data API.

crates/lakekeeper/src/server/generic_tables/load.rs
```diff
@@ -0,0 +1,137 @@
...
+    let GenericTableParameters {
+        prefix,
+        namespace,
+        table_name,
+    } = parameters;
+    let warehouse_id = require_warehouse_id(prefix.as_ref())?;
+    let authorizer = &state.v1_state.authz;
+
+    // ------------------- AUTHZ: GetMetadata -------------------
+    let event_ctx = APIEventContext::for_generic_table(
```
c-thiel, line 41: NOTE: Unlike `load_generic_table_credentials`, this path does not support DEFINER-chained authorization (`referenced-by`). It runs two separate authz round-trips (GetMetadata, then [ReadData, WriteData]) against the calling user only. If a generic table is reached through a view's DEFINER chain, the second authz call evaluates the wrong
principal and returns `storage_permissions = None`, so the response silently drops `config` and `storage_credentials`. Callers that need DEFINER-chained credential vending must use `/credentials`.

Either fold the three actions into the batched authorize_load_generic_table` flow used by credentials.rs, or keep this divergence and document it in the OpenAPI spec (which we still need to add)

crates/lakekeeper/migrations/20260220000000_add_generic_table.sql
```diff
@@ -0,0 +1,32 @@
+ALTER TYPE tabular_type ADD VALUE IF NOT EXISTS 'generic-table';
+
+CREATE TABLE generic_table (
+    warehouse_id     UUID NOT NULL,
+    generic_table_id UUID NOT NULL,
+    format           TEXT NOT NULL,
```
c-thiel: I know other catalogs don't do this but I would be in favor of restricting it:
`CHECK (format ~ '^[a-z][a-z0-9_-]{0,63}$')`

this would reject obvious garbage. WDYT?

lance-lakekeeper-test/test_lance_lakekeeper.py
c-thiel, line 1: Do you see a reason to keep Patterns A–D here? These kinds of workarounds are exactly what we want to retire with this PR.

c-thiel (commented):
Thanks for putting this together! I did a high level pass today and left a few comments. I think most are straight-forward - let me know if not. Two things that are missing currently:
- OpenAPI for the new generic-table endpoints. I see the management OpenAPI was updated because we added new actions, but we need a new OpenAPI spec for these new endpoints as they are neither just management (they are data plane) nor iceberg catalog
- We should run the `lance-lakekeeper-test` in CI. I think we should also nest this folder in a `integration-test` folder so that we can add tests for other formats into the same place in the future.

lance-lakekeeper-test/test_lance_lakekeeper.py
fresh-borzoni: no specific need, it was just the starting point with my investigations with Lance, I never thought of it as a proper integration test, but we can rework into it 👍

crates/lakekeeper/src/implementations/postgres/tabular/mod.rs
fresh-borzoni: I don't remember tbh, fixed

crates/lakekeeper/migrations/20260220000000_add_generic_table.sql
fresh-borzoni: added, and added app validation as well

crates/lakekeeper/migrations/20260220000000_add_generic_table.sql
fresh-borzoni: makes sense, thank you for explaining, I missed this path

authz/openfga/v4.3/store.fga.yaml
fresh-borzoni: makes sense

tests/migrations/justfile
```diff
@@ -38,6 +38,8 @@ test_migration $LAKEKEEPER_INITIAL_IMAGE $CREATE_WAREHOUSE_REQ:
     # Start lakekeeper with locally built binary. Triggers migration with that binary.
     # Flag --wait ensures the service is up and healthy before the next command executes.
     docker compose up --detach --wait lakekeeper_2
+    # Refresh connections so the statement cache picks up recreated enum OIDs.
+    docker compose up --detach --wait --force-recreate lakekeeper_initial
```
fresh-borzoni: there was an issue with stale OIDs for enums, so this fixes CI/CD, but this might be an issue on production tbh, as until statement cache is cleared, queries will try to use OIDS for previous version enums

fresh-borzoni:
@c-thiel Thank you for the review, addressed comments, PTAL 🙏

crates/lakekeeper/src/server/generic_tables/create.rs
```diff
@@ -0,0 +1,286 @@
...
+        format: request.format.clone(),
+        location,
+        doc: request.doc.clone(),
+        schema: request.schema.clone(),
+        statistics: request.statistics.clone(),
+        properties: request.properties.clone(),
+    };
+
+    let mut t = C::Transaction::begin_write(state.v1_state.catalog.clone()).await?;
+    let info = C::create_generic_table(creation, t.transaction()).await?;
```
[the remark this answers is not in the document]
c-thiel: https://github.com/lakekeeper/lakekeeper/issues/1774

c-thiel (commented):
Instead of a review I've put my comments in commits - please review them from your side and let me know what you think!
I have a generated commit msg in the commit so that you have a summary of the most important changes:
https://github.com/lakekeeper/lakekeeper/pull/1673/commits/473748b4370bbdc35efdb5db5dd862655fdaed2f

c-thiel:
@coderabbitai resolve
Summarize what's still open

c-thiel:
@coderabbitai full review

crates/lakekeeper/src/implementations/postgres/tabular/generic_table.rs
```diff
@@ -0,0 +1,602 @@
...
+        INNER JOIN warehouse w ON w.warehouse_id = t.warehouse_id AND w.status = 'active'
+        INNER JOIN namespace n ON n.namespace_id = t.namespace_id AND n.warehouse_id = t.warehouse_id
+        LEFT JOIN (
+            SELECT generic_table_id,
+                   ARRAY_AGG(key) as keys,
+                   ARRAY_AGG(value) as values
+            FROM generic_table_properties
+            WHERE warehouse_id = $1
+            GROUP BY generic_table_id
+        ) gtp ON t.tabular_id = gtp.generic_table_id
```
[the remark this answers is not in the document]
c-thiel: Tom Lane during the PG16 parallel-array_agg review (2018): "nobody does [write ORDER BY on both] in production... if we commit this [breaking co-alignment], something like 0.01% of users will be pleased, another 20% unaffected because they wrote ORDER BY which prevents parallel aggregation, and the remaining 80% will scream because we broke their queries."

This is OK

crates/lakekeeper/src/api/data/v1/generic_tables/mod.rs
```diff
@@ -0,0 +1,496 @@
...
+    Extension(metadata): Extension<RequestMetadata>,
+) -> Result<LoadGenericTableCredentialsResponse> {
+    let load_credentials_query = raw_query
+        .as_deref()
+        .and_then(|q| {
+            use serde::de::{IntoDeserializer, value::StrDeserializer};
+            let deserializer: StrDeserializer<'_, serde::de::value::Error> = q.into_deserializer();
+            LoadGenericTableCredentialsQuery::deserialize(deserializer).ok()
+        })
+        .unwrap_or_default();
```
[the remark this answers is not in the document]
c-thiel: logged warning - matching table behaviour

docs/docs/api/generic-table-open-api.yaml
```diff
@@ -0,0 +1,428 @@
...
+    CreateGenericTableRequest:
+      type: object
+      required:
+        - name
+        - format
+      properties:
+        base-location:
+          type:
+            - string
+            - 'null'
```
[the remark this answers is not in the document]
c-thiel: CreateGenericTableRequest.base_location is intentionally Option<String> because the server auto-derives a location from the warehouse storage profile when omitted. determine_tabular_location in [create.rs:251-257](vscode-webview://1mmrcnqfkhb3tld6onumj8uktjd950t36r0aiq82q68mcqhranul/crates/lakekeeper/src/server/generic_tables/create.rs#L251) accepts None and resolves it.

GenericTableData (the response) IS required to carry base-location because by then the location is resolved. The asymmetry — optional in request, required in response — is correct: making the request require it would force every caller to pre-compute a location, defeating the auto-derivation.

No change in this PR.

docs/docs/api/management-open-api.yaml
```diff
@@ -3791,6 +3953,19 @@ components:
...
+          properties:
+            generic-table:
+              allOf:
+                - $ref: '#/components/schemas/TabularIdentOrUuid'
+                - type: object
+                  required:
+                    - action
+                  properties:
+                    action:
+                      $ref: '#/components/schemas/LakekeeperGenericTableAction'
```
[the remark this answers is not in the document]
c-thiel: Resolved the underlying ergonomic concern (clients shouldn't have to send
`table-id` for a generic-table check) by following the existing convention
instead: `TabularIdentOrUuid` is already shared between Table and View, with
`view_id` accepted as a deserialization alias for `table_id`. Extended that
same pattern to generic tables — added `generic_table_id` and `generic_table`
as aliases.

A dedicated schema would have been asymmetric: View shares the type and
just adds an alias; generic-table doing different would have been the
inconsistency, not the original code. Chose consistency over a parallel type.

The wire format the server emits stays `table-id` / `table` across all three
kinds; clients can send any of the aliases.

tests/migrations/justfile
```diff
@@ -38,6 +38,10 @@ test_migration $LAKEKEEPER_INITIAL_IMAGE $CREATE_WAREHOUSE_REQ:
     # Start lakekeeper with locally built binary. Triggers migration with that binary.
     # Flag --wait ensures the service is up and healthy before the next command executes.
     docker compose up --detach --wait lakekeeper_2
+    # Cycle so the pool re-prepares statements against recreated enum OIDs.
+    # --no-deps skips re-running the old-image migrate (which would now fail).
+    docker compose restart lakekeeper_initial
+    docker compose up --detach --wait --no-deps lakekeeper_initial
```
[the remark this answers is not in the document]
c-thiel: Out of scope.

crates/authz-openfga/src/check.rs
```diff
@@ -505,6 +521,101 @@ async fn authorize_check_view<C: CatalogStore, S: SecretStore>(
...
+    action: AllGenericTableRelations,
+) -> Result<String, AuthZError> {
+    let authorizer = api_context.v1_state.authz.clone();
+    let (warehouse, gt_info) = tokio::join!(
+        C::get_active_warehouse_by_id(warehouse_id, api_context.v1_state.catalog.clone()),
+        C::get_generic_table_info(
+            warehouse_id,
+            generic_table.clone(),
+            TabularListFlags::active(),
+            api_context.v1_state.catalog.clone(),
```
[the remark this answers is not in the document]
c-thiel: Out of Scope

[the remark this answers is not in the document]
c-thiel: https://github.com/lakekeeper/lakekeeper/issues/1778

c-thiel:
https://github.com/lakekeeper/lakekeeper/pull/1673/commits/90a53477831204b5c1f52192d92bdf997e0b8d4f Closes https://github.com/lakekeeper/lakekeeper/issues/1778

authz/openfga/v4.5/components/lakekeeper_generic_table.fga
```diff
@@ -0,0 +1,37 @@
...
+
+    # ------------------ Actions ------------------
+    define can_drop: modify
+    define can_undrop: modify
+    define can_write_data: modify
+    define can_read_data: select
+    define can_get_metadata: describe
+    define can_rename: modify
+    define can_include_in_list: can_get_metadata
+    define can_set_protection: modify
```
fresh-borzoni, line 26: Protection for generic tables looks half-wired, as we added `can_set_protection` to the FGA model, but there's no endpoint, no `protected` field on the load response, and `APIGenericTableAction` still excludes `SetProtection` with a   "doesn't expose" comment.

Is it intentional? and if so - docs seems to contradict.

fresh-borzoni (commented):
@c-thiel Thank you for helping with this, appreciate you closing the gaps here.
Looked through, found one more gap

c-thiel:
@fresh-borzoni another point I just noticed: In the `api` module I think we should move generic tables to their own folder. We don't have them as part of the management API anymore, I think we should reflect this in rust.

authz/openfga/v4.5/components/lakekeeper_generic_table.fga
c-thiel: Addressed in https://github.com/lakekeeper/lakekeeper/pull/1673/commits/77b9dd279550ea0fd0f91ef1776be1ec9a20eb72

PR #1709, opened 2026-04-13
docs: document case sensitivity design and fix case-only rename conflict

c-thiel:
BEGIN_COMMIT_OVERRIDE
fix: return caller's case from namespace lookups, canonical from id lookups

fix: allow renaming tables to a different case of their own name
docs: document case sensitivity design and fix case-only rename conflict
END_COMMIT_OVERRIDE

## Summary by CodeRabbit

* **New Features**
  * Lookups are case-insensitive but preserve and echo caller casing; name-based queries return caller-cased identifiers while id-based queries return canonical stored casing. Cache supports both canonical and caller-case keys and overlays caller casing on hierarchy results.

* **Bug Fixes**
  * Renaming a table no longer falsely conflicts with itself during conflict checks.

* **Documentation**
  * Added comprehensive "Identifier Case Sensitivity" guidance and a gotcha note.

* **Tests**
  * Expanded backend, Python, and Spark tests covering lookup, casing preservation, cache behavior, rename and view scenarios.


c-thiel:
@coderabbitai review

PR #1728, opened 2026-04-22
chore: Make LakekeeperStorage dyn-Compatible

brgr-s:
## Summary by CodeRabbit

* **Refactor**
  * Storage trait methods converted to async with unified parameter shapes and new delegating wrappers for backends.
* **Bug Fixes**
  * Improved recursive and batch delete flows; NotFound treated as success and batch errors mapped consistently.
* **Tests**
  * Unit and integration tests updated to match revised storage API.
* **Chores**
  * Added workspace-managed deps for HTTP middleware, retry, and tempfile; examples/install updated to stable Strimzi API and fixed operator version.

### DIFF
--- a/crates/io/src/adls/adls_storage.rs
+++ b/crates/io/src/adls/adls_storage.rs
@@ -123,7 +123,7 @@ impl LakekeeperStorage for AdlsStorage {
         Ok(())
     }
 
-    async fn delete_batch(&self, paths: Vec<String>) -> Result<(), DeleteBatchError> {
+    async fn delete_batch(&self, paths: &[String]) -> Result<(), DeleteBatchError> {
         // Group paths by account and filesystem
         let grouped_paths = group_paths_by_container(paths)?;
 
@@ -422,7 +422,7 @@ impl LakekeeperStorage for AdlsStorage {
 
     /// Native ADLS Gen2 recursive delete via `DELETE` + `recursive=true`.
     ///
-    /// The trailing slash is stripped before parsing: the ADLS DataLake API
+    /// The trailing slash is stripped before parsing: the ADLS API
     /// accepts either form for a path, and `recursive=true` is what signals
     /// directory semantics. For file paths the `recursive` flag is ignored
     /// server-side, so the same call safely handles both files and directories.
@@ -430,7 +430,7 @@ impl LakekeeperStorage for AdlsStorage {
     /// `NotFound` responses are treated as success, matching the idempotent
     /// semantics of `delete` and `delete_batch` on this backend — removing an
     /// already-absent prefix is a no-op, not an error.
-    async fn remove_all(&self, path: &str) -> Result<(), DeleteError> {
+    async fn remove_all(&self, path: &str, _: Option<usize>) -> Result<(), DeleteError> {
         let path = path.trim_end_matches('/');
         let adls_location = AdlsLocation::try_from_str(path, true)?;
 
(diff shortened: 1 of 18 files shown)


crates/io/src/adls/adls_storage.rs
```diff
@@ -123,10 +123,7 @@ impl LakekeeperStorage for AdlsStorage {
         Ok(())
     }
 
-    async fn delete_batch(
-        &self,
-        paths: impl IntoIterator<Item = impl AsRef<str>>,
-    ) -> Result<(), DeleteBatchError> {
+    async fn delete_batch(&self, paths: Vec<String>) -> Result<(), DeleteBatchError> {
```
c-thiel: Avoid `Vec<String>` in arguments as it needs allocations. We should use `&[String]` or `&[&str]`.
Did you consider using BoxStream like iceberg?

Would also help with orphan file removal.

crates/io/src/error.rs
```diff
@@ -166,6 +166,15 @@ impl From<DeleteError> for DeleteBatchError {
...
 
+impl From<DeleteBatchError> for DeleteError {
+    fn from(err: DeleteBatchError) -> Self {
+        match err {
+            DeleteBatchError::InvalidLocation(e) => DeleteError::InvalidLocation(e),
+            DeleteBatchError::IOError(e) => DeleteError::IOError(e),
+        }
+    }
+}
+
```
c-thiel, line 177: We might want to add context here to preserver the information that this is coming from a batch deletion.

crates/io/src/lib.rs
```diff
@@ -178,388 +177,270 @@ impl FileInfo {
...
+                    .collect();
+                self.delete_batch(locations).await.map_err(Into::into)
+            })
+            .await
     }
 }
 
+/// Maximum number of concurrent `delete_batch` calls issued by the default
+/// [`LakekeeperStorage::remove_all`] implementation.
+pub const REMOVE_ALL_CONCURRENCY: usize = 10;
```
c-thiel: This should be configurable. We can use it as a default, but clients must be able to tweak this. I think we should keep a constant and make it the default but the storage should allow to overwrite it.

Is 10 a common value for S3 and GCS?

crates/io/src/adls/adls_storage.rs
brgr-s: Replaces with `&[String]` as the option with more ups than downs.

crates/io/src/error.rs
brgr-s: Added context on call site, instead of creating own error type. `LakekeeperStorage` methods are used in placed where they get mapped to `IOErrorExt`, which has only `From<Read/Write/Delete_Error>`. Seems pointless to extend this.

crates/io/src/lib.rs
brgr-s: `remove_all` has `Option`al parameter now.

10 was taken from previous implementation.

crates/io/src/lib.rs
```diff
@@ -178,388 +177,307 @@ impl FileInfo {
...
-                None => Ok(()),
-            }
-        }
+                    DeleteBatchError::IOError(ioerror) => DeleteError::IOError(
+                        ioerror.with_context("Remove all operation failed when deleting files"),
+                    ),
+                })
+            })
+            .await
     }
```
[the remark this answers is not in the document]
brgr-s: I would expect `Some(1)` to be interpreted as "sequential", and `Some(0)`/`None` as unlimited, which is consistent with [for_each_concurrent](https://docs.rs/futures/latest/futures/stream/trait.StreamExt.html#method.for_each_concurrent)

c-thiel (approved):

PR #1734, opened 2026-04-30
chore: Bridge lakekeeper-io to Iceberg Storage trait

brgr-s:
## Summary by CodeRabbit

* **New Features**
  * Partial byte-range reads (parallel chunking for large objects)
  * File metadata with size and unified exists API
  * Streaming writers with auto-promotion to multipart/resumable uploads
  * Iceberg storage integration layer

* **Improvements**
  * Stronger integrity and last-modified validation for reads/writes
  * Safer batch delete behavior and tighter error mapping
  * Directory listings now include per-object sizes

* **Tests**
  * Expanded cross-backend tests for range reads, streaming writers, metadata, and existence checks


crates/io/src/iceberg_bridge.rs
```diff
@@ -0,0 +1,221 @@
...
+
+impl From<DeleteBatchError> for iceberg::Error {
+    fn from(value: DeleteBatchError) -> Self {
+        iceberg::Error::new(
+            iceberg::ErrorKind::Unexpected,
+            format!("Delete stream error: {value}"),
+        )
+        .with_source(value)
+    }
+}
```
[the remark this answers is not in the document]
brgr-s: `iceberg` has no `ErrorKind::NotFound`, precondition is checked via `.exists()`.

A reasonable mapping could be `ErrorKind::ConditionNotMatch => iceberg::ErrorKind::PreconditionFailed` and _maybe_ `InvalidLocation => DataInvalid`. All other `ErrorKind`s do not really map to `iceberg::ErrorKind`s. The rest would be mapped to `Unexcpected`. This seems to be a lot of boilerplate for one correct mapping and "kinda sorta" correcht mapping.

crates/io/src/adls/adls_storage.rs
```diff
@@ -512,3 +557,322 @@ fn group_paths_by_container(
...
+}
+
+/// Best-effort variant of [`delete_file`] that swallows the delete error after
+/// logging it.
+///
+/// Use on cleanup paths where the caller propagates a different (original)
+/// error and the delete failure is unactionable: `close`, bulk write, `Drop`.
+/// The `context` field is included in the warn log to disambiguate which
+/// cleanup site triggered the deletion.
+async fn try_delete_file(client: &FileClient, path: &str, context: &str) {
```
c-thiel: Should we rename to `delete_partial_file_logged` - to make the purpose of this function in contrast to `delete_file`?

crates/io/src/adls/adls_storage.rs
```diff
@@ -512,3 +557,322 @@ fn group_paths_by_container(
...
+            let chunk = self.buffer.split_to(DEFAULT_BYTES_PER_REQUEST).freeze();
+            let chunk_len =
+                safe_usize_to_i64(chunk.len(), self.path.clone()).map_err(WriteError::IOError)?;
+            if let Err(append_error) =
+                append_chunk(&self.client, self.offset, chunk, &self.path).await
+            {
+                if let Err(delete_error) = delete_file(&self.client, &self.path).await {
+                    self.state = AdlsWriterState::AbortFailed;
+                    return Err(delete_error);
+                }
```
c-thiel: When append_chunk fails and the cleanup delete_file also fails, the writer returns the cleanup error and drops the original failure. The actionable signal is the upload error

crates/io/src/s3/s3_storage.rs
```diff
@@ -434,19 +379,559 @@ fn try_parse_file_info(
...
+                    .await
+                    {
+                        Ok(part) => part,
+                        Err(upload_error) => {
+                            if let Err(abort_error) =
+                                abort_multipart(&self.client, &self.location, &upload_id).await
+                            {
+                                let _ =
+                                    std::mem::replace(&mut self.state, S3WriterState::AbortFailed);
+                                return Err(abort_error);
```
c-thiel: In S3FileWrite::close, when the final part upload fails and the subsequent abort_multipart also fails, the code returns abort_error and silently drops upload_error. Compare with the complete_multipart failure branch immediately below it, which uses try_abort_multipart and returns the original error — that pattern is correct and should be applied here too.

crates/io/src/adls/adls_storage.rs
```diff
@@ -285,92 +306,116 @@ impl LakekeeperStorage for AdlsStorage {
...
+    async fn metadata(&self, path: &str) -> Result<FileInfo, ReadError> {
+        let adls_location = AdlsLocation::try_from_str(path, true)?;
+        require_key(&adls_location)?;
+        let client = self.get_file_client(&adls_location)?;
+        let head_response = head(&client, &adls_location).await?;
+        let size = head_response
+            .content_length
+            .and_then(|cl| crate::size_to_u64(cl, adls_location.location().as_str()));
+        let last_modified =
+            DateTime::from_timestamp(head_response.last_modified.unix_timestamp(), 0);
```
c-thiel: metadata truncates sub-second last_modified. DateTime::from_timestamp(ts, 0) hardcodes the nanosecond field to 0. This is inconsistent with the GCS parse_timestamp helper which passes t.subsec_nanos().
Do we have a reason for this? If so, lets add a comment.

crates/io/src/iceberg_bridge.rs
```diff
@@ -0,0 +1,229 @@
...
+    }
+
+    async fn metadata(&self, path: &str) -> iceberg::Result<FileMetadata> {
+        let info = self.lakekeeper_io.metadata(path).await?;
+        let size = info.size().ok_or_else(|| {
+            iceberg::Error::new(
+                iceberg::ErrorKind::Unexpected,
+                format!("Backend did not report size for {path}"),
+            )
+        })?;
```
c-thiel, line 98: The trait `FileInfo` allows `size` to be none. The bridge is narrower here and fails.
Shouldn't we widen it?

crates/io/src/adls/adls_storage.rs
```diff
@@ -197,86 +194,110 @@ impl LakekeeperStorage for AdlsStorage {
...
+                        WriteError::IOError(IOError::new(
+                            ErrorKind::ConditionNotMatch,
+                            format!(
+                                "Calculated offset for write exceeds i64 limit: {raw_offset} > {}",
+                                i64::MAX
+                            ),
+                            path.to_string(),
+                        ))
+                    })?;
+                    let chunk = Bytes::copy_from_slice(chunk);
```
c-thiel: bytes.chunks(chunk_size) returns &[u8], then Bytes::copy_from_slice(chunk) copies into a fresh allocation. Instead bytes.slice(offset..offset+len) produces a zero-copy view.

crates/io/src/iceberg_bridge.rs
```diff
@@ -0,0 +1,229 @@
...
+                .delete_batch(&chunk)
+                .await
+                .map_err(Into::<iceberg::Error>::into)?;
+        }
+        Ok(())
+    }
+
+    fn new_input(&self, path: &str) -> iceberg::Result<iceberg::io::InputFile> {
+        Ok(iceberg::io::InputFile::new(
+            Arc::new(self.clone()),
```
c-thiel, line 148: every new_input / new_output allocates a fresh Arc<Self>. Iceberg's InputFile requires Arc<dyn Storage>. The bridge already holds an inner Arc<dyn LakekeeperStorage>, so cloning self into a new outer Arc is wasteful per-file. Wrap IcebergStorageBridge in Arc once at construction and store it as Arc<Self>, or change the public constructor to return Arc<Self>.

crates/io/src/s3/s3_storage.rs
```diff
@@ -80,242 +82,170 @@ impl LakekeeperStorage for S3Storage {
...
 
-        // Use the shared utility function to assemble chunks
-        let combined_data =
-            crate::assemble_chunks(flattened_results, file_size, DEFAULT_BYTES_PER_REQUEST).await?;
+        if range_size <= MAX_BYTES_PER_REQUEST {
+            return fetch_range(&self.client, &s3_location, range, None).await;
+        }
 
-        Ok(combined_data)
+        let head_response = head(&self.client, &s3_location).await?;
```
c-thiel, line 247: nit: We could piggyback on chunk 0's ETag instead of a separate call.

crates/io/src/iceberg_bridge.rs
```diff
@@ -0,0 +1,229 @@
...
+
+#[typetag::serde]
+impl StorageFactory for IcebergStorageBridgeFactory {
+    fn build(&self, _config: &StorageConfig) -> iceberg::Result<Arc<dyn Storage>> {
+        // we ignore `config` because the inner `bridge` is already configured.
+        Ok(self.bridge.clone())
+    }
+}
+
+/// Intentional hard fail for Ser/Deser because `brdige` cannot be ser/deser,
```
c-thiel:
```suggestion
/// Intentional hard fail for Ser/Deser because `bridge` cannot be ser/deser,
```

c-thiel (commented):
The integration tests cover the writer state machine on the happy path (basic, multi-chunk, large-streaming, double-close, write-after-close). They do not cover the writer's Drop-triggered abort path on S3/GCS — i.e., dropping a Multipart-state writer without calling close and verifying the multipart upload is actually aborted on the cloud side.

brgr-s:
we can try, but if the test fails it might be because the detached task is still running or waiting on a blocked runtime. it has to be timeout based, which felt a bit flacky to me.

crates/io/src/adls/adls_storage.rs
brgr-s: good catch, I think other paths do it correctly

crates/io/src/adls/adls_storage.rs
brgr-s: S3 and ADLS return only seconds precision. however setting to `t.nanoseconds()` doen't hurt, I just forgot to do it. I opt for setting `t.nanoseconds()` _and_ adding a comment.

crates/io/src/iceberg_bridge.rs
brgr-s: `iceberg::FileMetadata` has a required `size` field, so the bridge fails here.

crates/io/src/adls/adls_storage.rs
brgr-s: not the only place :unamused:

crates/io/src/iceberg_bridge.rs
brgr-s: possible to hold `Weak<Self>` in `LakekeeperStorageBridge` and `upgrade` on construction of `InputFile` or `OutputFile`, but constructed would need to be `-> Arc<Self>` and other shenanigans.

Issue #1571, opened 2026-01-16
Support multiple OpenID Connect providers for cross-cluster Kubernetes authentication


mithilajoshi:
### Use Case

We run Lakekeeper in a management cluster but need to authenticate Spark jobs running in separate compute clusters. These compute clusters use EKS Pod Identity / IRSA, so jobs have Kubernetes ServiceAccount tokens issued by their respective EKS OIDC providers.

### Current Behavior

- `LAKEKEEPER__OPENID_PROVIDER_URI` accepts a single OIDC provider (e.g., Okta)
- `LAKEKEEPER__OPENID_ADDITIONAL_ISSUERS` adds issuers to the validation list but **does not fetch JWKS from those issuers**
- `LAKEKEEPER__ENABLE_KUBERNETES_AUTHENTICATION` uses TokenReview API, which only works within the same cluster

When an EKS token arrives:
1. `can_handle_token()` returns `true` (EKS issuer is in `additional_issuers`)
2. `authenticate()` fails with "Key id not found in JWKS" (because JWKS was only fetched from primary provider)

### Proposed Solution

Support multiple OIDC provider URIs, each creating its own `JWKSWebAuthenticator` in the chain:

`LAKEKEEPER__OPENID_PROVIDER_URIS="https://okta.example.com,https://oidc.eks.us-east-1.amazonaws.com/id/ABC123"`

Or alternatively:

`LAKEKEEPER__OPENID_PROVIDER_URI="https://okta.example.com"LAKEKEEPER__OPENID_PROVIDER_URI_2="https://oidc.eks.us-east-1.amazonaws.com/id/ABC123"`

Each URI would:

- Create a separate JWKSWebAuthenticator
- Fetch JWKS from its own .well-known/openid-configuration
- Be added to the AuthenticatorChain

The existing chain architecture in limes already supports this - the change is in Lakekeeper's configuration layer.

#### Why additional_issuers doesn't solve this
The `add_additional_issuers()` method in limes is designed for cases where the same IdP is reachable under multiple URLs (e.g., Azure's login.microsoftonline.com vs sts.windows.net). It validates the issuer claim but doesn't fetch JWKS from those additional issuers.
From limes/src/jwks.rs:
`pub fn add_additional_issuers(mut self, additional_issuers: Vec<String>) -> Self {    // Only adds to self.issuers list - no new JwksClient created    self.issuers.extend(additional_issuers);    self}`

### Environment
Lakekeeper Version: 0.11.1
[...]

PR #1760, opened 2026-05-19
feat(authn): support multiple OIDC providers via LAKEKEEPER__OPENID_PROVIDERS

mithilajoshi:
## Summary

This is that implementation.

In issue #1571 (now closed), @c-thiel commented:

> "if someone would build a more generic configuration for OSS that is backwards compatible, we would gladly include that too"

This PR adds `LAKEKEEPER__OPENID_PROVIDERS` — a config-driven, backwards-compatible way to authenticate tokens from multiple OIDC providers simultaneously (e.g. Okta for users + EKS OIDC for Kubernetes service accounts). The existing `OPENID_PROVIDER_URI` and `OPENID_ADDITIONAL_ISSUERS` can't do this: `additional_issuers` only validates the issuer claim but doesn't fetch JWKS from those URIs, leading to "Key id not found in JWKS" failures.

## Configuration

```bash
LAKEKEEPER__OPENID_PROVIDERS='[
  {
    "uri": "https://company.okta.com",
    "idp_id": "okta",
    "audience": "https://company.okta.com",
    "subject_claims": "sub"
  },
  {
    "uri": "https://oidc.eks.us-east-1.amazonaws.com/id/CLUSTER_ID",
    "idp_id": "eks-prod",
    "audience": "sts.amazonaws.com"
  }
]'
```

Per-provider fields:

| Field | Required | Description |
|-------|----------|-------------|
| `uri` | yes | OIDC provider URI (must expose `.well-known/openid-configuration`) |
| `idp_id` | yes | Prefix used in user IDs: `{idp_id}~{subject}` |
| `audience` | no | Comma-separated accepted audiences |
| `additional_issuers` | no | Comma-separated additional trusted issuers |
| `scope` | no | Required scope claim |
| `subject_claims` | no | Comma-separated claims to use as subject, in preference order. Defaults to `oid,sub`. |

## Design

- **Backwards compatible**: `LAKEKEEPER__OPENID_PROVIDER_URI` continues to work unchanged. Multi-provider mode only activates when `LAKEKEEPER__OPENID_PROVIDERS` is set.
- Failed providers are logged and skipped — one misconfigured provider doesn't block the others.
- Multi-provider mode uses `{idp_id}~{subject}` (e.g. `okta~alice@co.com`). Single-provider mode is unchanged.

## Files changed

- `crates/lakekeeper/src/config.rs` — `OidcProviderConfig` struct, `openid_providers` field, serde helpers, `authn_enabled()` update, 6 new unit tests
- `crates/lakekeeper/src/service/authn.rs` — `build_multi_oidc_authenticators()`, `build_single_oidc_authenticator()` extracted, dynamic chain builder
- `crates/lakekeeper-bin/src/serve.rs` — minor import fix
- `docs/docs/authentication.md` — "Multiple OIDC Providers" section
- `docs/docs/configuration.md` — `LAKEKEEPER__OPENID_PROVIDERS` reference and example

## Testing

- 6 new unit tests in `config.rs`
- Manually tested against Okta + EKS OIDC

## Breaking changes

None — purely additive.

Relates to #1571

## Summary by CodeRabbit

* **New Features**
  * Configure multiple OIDC identity providers with per-provider settings, deterministic ordering, and provider-scoped user IDs
  * Mark secondary providers optional at startup (failures skip) while required providers still block startup
  * UI login enablement follows consolidated authentication-enabled logic

* **Documentation**
  * Added guides and examples for multi-provider configuration, startup resilience, and identity-continuity guidance

* **Tests**
  * Expanded tests for parsing, ordering, startup behavior, and authn-enabled cases

[Review Change Stack](https://app.coderabbit.ai/change-stack/lakekeeper/lakekeeper/pull/1760?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/crates/lakekeeper/src/config.rs
+++ b/crates/lakekeeper/src/config.rs
@@ -25,7 +25,7 @@ use crate::{
     WarehouseId,
     service::{
         ArcProjectId, UserId,
-        authn::{IDP_SEPARATOR, K8S_IDP_ID},
+        authn::{IDP_SEPARATOR, K8S_IDP_ID, OIDC_IDP_ID},
     },
 };
 
@@ -167,6 +167,10 @@ fn validate_openid_provider_ids(config: &DynAppConfig) {
         assert!(
             !idp_id.eq_ignore_ascii_case(K8S_IDP_ID),
             "Invalid OIDC provider '{idp_id}': IdP ID '{K8S_IDP_ID}' is reserved"
+        );
+        assert!(
+            !idp_id.eq_ignore_ascii_case(OIDC_IDP_ID),
+            "Invalid OIDC provider '{idp_id}': IdP ID '{OIDC_IDP_ID}' is reserved"
         );
     }
 }
@@ -461,8 +465,8 @@ pub struct DynAppConfig {
     /// Supports nested claims using dot notation, e.g., `resource_access.account.roles`
     pub openid_roles_claim: Option<String>,
     /// Multiple OIDC providers keyed by identity provider ID.
-    /// When set, each provider gets its own JWKS authenticator and this takes
-    /// precedence over the single-provider configuration (`openid_provider_uri`).
+    /// When set, each provider gets its own JWKS authenticator and is added
+    /// in addition to the single-provider configuration (`openid_provider_uri`).
     #[serde(default)]
     pub openid_providers: HashMap<String, OidcProviderConfig>,
 
@@ -1148,8 +1152,16 @@ impl DynAppConfig {
         self.default_tabular_expiration_delay_seconds
     }
 
+    /// Is any authentication active? Used by /info to reject anonymous.
     pub fn authn_enabled(&self) -> bool {
-        self.openid_provider_uri.is_some() || !self.openid_providers.is_empty()
+        self.openid_provider_uri.is_some()
+            || !self.openid_providers.is_empty()
+            || self.enable_kubernetes_authentication
+    }
+
+    /// Does the UI have an SSO target? Used by the UI config.
+    pub fn ui_login_enabled(&self) -> bool {
+        self.openid_provider_uri.is_some()
     }
 
     /// Helper for common conversion of optional page size to `i64`.
@@ -2014,6 +2026,19 @@ mod test {
     }
 
     #[test]
+    #[should_panic(expected = "IdP ID 'oidc' is reserved")]
+    fn test_openid_provider_id_rejects_oidc() {
+        figment::Jail::expect_with(|jail| {
+            jail.set_env(
+                "LAKEKEEPER_TEST__OPENID_PROVIDERS__OIDC__URI",
+                "https://company.okta.com",
+            );
+            let _config = get_config();
+            Ok(())
+        });
+    }
+
+    #[test]
     fn test_user_assignments_cache() {
         figment::Jail::expect_with(|_jail| {
             let config = get_config();
@@ -2168,7 +2193,7 @@ mod test {
                 "trino.run-as-owner",
             );
             jail.set_env(
-                "LAKEKEEPER_TEST__TRUSTED_ENGINES__TRINO__IDENTITIES__OIDC__AUDIENCES",
+                "LAKEKEEPER_TEST__TRUSTED_ENGINES__TRINO__IDENTITIES__prod_AUDIENCES",
                 "[trino_dev, trino_prod]",
             );
             jail.set_env(
@@ -2625,6 +2650,16 @@ mod test {
                 "LAKEKEEPER_TEST__OPENID_PROVIDERS__OKTA__URI",
                 "https://okta.example.com",
             );
+            let config = get_config();
+            assert!(config.authn_enabled());
+            Ok(())
+        });
+    }
+
+    #[test]
+    fn test_authn_enabled_with_kubernetes_only() {
+        figment::Jail::expect_with(|jail| {
+            jail.set_env("LAKEKEEPER_TEST__ENABLE_KUBERNETES_AUTHENTICATION", "true");
             let config = get_config();
             assert!(config.authn_enabled());
             Ok(())
--- a/crates/lakekeeper/src/config.rs
+++ b/crates/lakekeeper/src/config.rs
@@ -73,6 +73,22 @@ fn get_config() -> DynAppConfig {
         .extract::<DynAppConfig>()
         .expect("Valid Configuration");
 
+    if let Some(providers) = &config.openid_providers {
+        let mut seen_idp_ids = HashSet::new();
+        for provider in providers {
+            assert!(
+                !provider.idp_id.trim().is_empty(),
+                "Invalid OIDC provider '{}': idp_id must not be empty",
+                provider.uri
+            );
+            assert!(
+                seen_idp_ids.insert(provider.idp_id.clone()),
+                "Invalid OIDC providers: duplicate idp_id '{}'",
+                provider.idp_id
+            );
+        }
+    }
+
     // Ensure base_uri has a trailing slash
     if let Some(base_uri) = config.base_uri.as_mut() {
         let base_uri_path = base_uri.path().to_string();
@@ -438,6 +454,16 @@ pub struct DynAppConfig {
     /// The field should contain an array of strings or a single string.
     /// Supports nested claims using dot notation, e.g., `resource_access.account.roles`
     pub openid_roles_claim: Option<String>,
+    /// Multiple OIDC providers configuration.
+    /// When set, each provider gets its own JWKS authenticator.
+    /// Takes precedence over the single-provider configuration (`openid_provider_uri`).
+    /// Set as a JSON string: `'[{"uri":"https://...", "idp_id":"okta", ...}]'`
+    #[serde(
+        default,
+        deserialize_with = "deserialize_openid_providers",
+        serialize_with = "serialize_openid_providers"
+    )]
+    pub openid_providers: Option<Vec<OidcProviderConfig>>,
 
     // ------------- AUTHORIZATION - OPENFGA -------------
     #[serde(default)]
@@ -618,6 +644,63 @@ where
         .as_deref()
         .map(|value| value.join(","))
         .serialize(serializer)
+}
+
+/// Deserialize OIDC providers from either:
+/// - A JSON string: `'[{"uri":"https://...", "idp_id":"okta", ...}]'`
+/// - A JSON array (when parsing config files)
+fn deserialize_openid_providers<'de, D>(
+    deserializer: D,
+) -> Result<Option<Vec<OidcProviderConfig>>, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    let value = Option::<serde_json::Value>::deserialize(deserializer)?;
+    match value {
+        None => Ok(None),
+        Some(serde_json::Value::String(s)) => {
+            // Parse JSON string
+            let providers: Vec<OidcProviderConfig> =
+                serde_json::from_str(&s).map_err(serde::de::Error::custom)?;
+            if providers.is_empty() {
+                Ok(None)
+            } else {
+                Ok(Some(providers))
+            }
+        }
+        Some(serde_json::Value::Array(arr)) => {
+            // Parse JSON array directly
+            let providers: Vec<OidcProviderConfig> =
+                serde_json::from_value(serde_json::Value::Array(arr))
+                    .map_err(serde::de::Error::custom)?;
+            if providers.is_empty() {
+                Ok(None)
+            } else {
+                Ok(Some(providers))
+            }
+        }
+        Some(other) => Err(serde::de::Error::custom(format!(
+            "Expected JSON string or array for openid_providers, got: {}",
+            other
+        ))),
+    }
+}
+
+fn serialize_openid_providers<S>(
+    value: &Option<Vec<OidcProviderConfig>>,
+    serializer: S,
+) -> Result<S::Ok, S::Error>
+where
+    S: serde::Serializer,
+{
+    match value {
+        None => serializer.serialize_none(),
+        Some(providers) => {
+            let json_str =
+                serde_json::to_string(providers).map_err(serde::ser::Error::custom)?;
+            serializer.serialize_str(&json_str)
+        }
+    }
 }
 
 fn deserialize_origin<'de, D>(deserializer: D) -> Result<Option<Vec<HeaderValue>>, D::Error>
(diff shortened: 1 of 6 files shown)


c-thiel (commented):
Thanks for the PR @mithilajoshi!
A few high level comments first - I'll do a more thorough review in the next days:
- This PR drops support for `openid_roles_claim` - we still need this as part of the IdP configuration so that Authorizers have the role information from tokens available. OpenFGA does not use this yet in our default implementation, but we must keep support for custom implementations
- `idp_id` must be validated to not contain `~` (`IDP_SEPARATOR`)
- The `OidcProviderConfig` are wrong - they show index config.rs:1066-1071 advertises figment-indexed syntax (LAKEKEEPER__OPENID_PROVIDERS__0__URI=...), but the custom deserialize_openid_providers at config.rs:649-688] only accepts Value::String (JSON) or Value::Array
- Every other multi-field config in this file uses structured per-field env vars; this commit is the first to require a JSON-string blob env var. I believe we should align the configuration and use structs instead (make it a `HashMap` and have the idp id in the env var as a placeholder). This would also make uniqueness checks obsolete, provides faster accessors and naturally encodes the uniqueness semantic in the type
- In `authn.rs:286` we always skip providers for which we can't load JWKS configurations. Some configurations might want to stick to the current strict behaviour. We might want to add a flag `...__REQUIRE_CONNECTED_ON_STARTUP` with default being `true` that users can disable.
- Route the legacy single-provider path through a one-element `Vec<OidcProviderConfig>` (with `idp_id = "oidc"` to match the existing `OIDC_IDP_ID` constant at `authn.rs:69` and preserve existing user IDs; the corresponding `K8S_IDP_ID = "kubernetes"` at `authn.rs:70` would also need to be reserved against user-configured `idp_id` values) so both paths share one builder — removes ~60 lines of duplication in `authn.rs` and makes the `roles_claim` regression structurally impossible.

c-thiel:
You can ingnore the vault tests failing - its an update by Hashicorp. Fix is landing on main soon.

crates/lakekeeper/src/service/authn.rs
```diff
@@ -125,79 +129,160 @@ pub async fn get_default_authenticator_from_config() -> anyhow::Result<Option<Bu
...
+            }
+            Ok(Some(chain_builder.build().into()))
+        }
     }
 }
 
+fn oidc_provider_configs_from_config(
+    config: &crate::config::DynAppConfig,
+) -> Vec<(String, OidcProviderConfig)> {
+    if !config.openid_providers.is_empty() {
```
c-thiel, line 289: I think we should switch this from replace to extend semantics: `OPENID_PROVIDER_URI` stays as the primary OIDC provider (idp_id = oidc) and `OPENID_PROVIDERS` adds additional ones. That mirrors how `ENABLE_KUBERNETES_AUTHENTICATION `is additive today, and gives us two things for free:

1. Existing deployments keep their `oidc~<sub>` user IDs in Postgres/OpenFGA — without this, an operator who points `OPENID_PROVIDERS__OKTA__URI` at the same IdP silently orphans every existing user/role binding.
2. The UI keeps working unchanged

crates/lakekeeper/src/config.rs
```diff
@@ -149,6 +154,23 @@ fn get_config() -> DynAppConfig {
     config
 }
 
+fn validate_openid_provider_ids(config: &DynAppConfig) {
```
c-thiel, line 166: Following https://github.com/lakekeeper/lakekeeper/pull/1760/changes#r3281889513
I think we should also reserve oidc (case-insensitive, same as kubernetes), since it's the IdP ID of the primary provider. That gives us a clean collision check between OPENID_PROVIDER_URI and any OPENID_PROVIDERS__OIDC__*.

crates/lakekeeper/src/config.rs
```diff
@@ -969,6 +1000,58 @@ impl std::default::Default for Tokio {
...
+/// (e.g., Okta for users + EKS OIDC for Kubernetes service accounts).
+///
+/// # Example Environment Variables
+/// ```bash
+/// LAKEKEEPER__OPENID_PROVIDERS__OKTA__URI=https://company.okta.com
+/// LAKEKEEPER__OPENID_PROVIDERS__OKTA__AUDIENCE=https://company.okta.com
+/// LAKEKEEPER__OPENID_PROVIDERS__OKTA__SUBJECT_CLAIMS=sub
+/// ```
+#[derive(Clone, Deserialize, Serialize, Debug, PartialEq)]
+pub struct OidcProviderConfig {
```
c-thiel: Maybe we call this `AdditionalOidcProviderConfig` following the argument above

crates/lakekeeper/src/config.rs
```diff
@@ -969,6 +1000,58 @@ impl std::default::Default for Tokio {
...
+    )]
+    pub subject_claims: Option<Vec<String>>,
+    /// Claim to use in provided JWT tokens to extract roles.
+    /// The field should contain an array of strings or a single string.
+    /// Supports nested claims using dot notation, e.g., `resource_access.account.roles`
+    #[serde(default)]
+    pub roles_claim: Option<String>,
+    /// If true, fail startup when this provider's OIDC/JWKS configuration cannot be loaded.
+    #[serde(default = "default_true")]
+    pub require_connected_on_startup: bool,
```
c-thiel: Minor: we should document (.md) that the primary provider always requires connect on startup.

crates/lakekeeper/src/service/authn.rs
```diff
@@ -125,79 +129,160 @@ pub async fn get_default_authenticator_from_config() -> anyhow::Result<Option<Bu
...
+    // Collect all authenticators into a chain: OIDC first (priority), then any additional
+    let mut all_authenticators: Vec<AuthenticatorEnum> = authn_oidc_list;
+    if let Some(authn) = authn_k8s_audience.map(AuthenticatorEnum::from) {
+        all_authenticators.push(authn);
+    }
+    if let Some(authn) = authn_k8s_legacy.map(AuthenticatorEnum::from) {
+        all_authenticators.push(authn);
+    }
+
+    match all_authenticators.len() {
```
c-thiel, line 253: The config-parsing tests are great. I think we should add one test that calls `get_default_authenticator_from_config and asserts` (a) chain order with primary + additional + K8s, (b) `require_connected_on_startup=false` actually skips a failing additional provider, (c) failure when `true`. That's the part most likely to regress.

docs/docs/authentication.md
```diff
@@ -452,6 +452,52 @@ We are now ready to deploy Lakekeeper and login via the UI. Set the following en
...
 
+## Multiple OIDC Providers
+
+For scenarios where you need to authenticate tokens from multiple identity providers simultaneously—such as Okta for human users and EKS OIDC for Kubernetes service accounts—Lakekeeper supports configuring multiple OIDC providers.
+
+When multiple providers are configured, each provider fetches its own JWKS keys independently. Incoming tokens are checked against each provider in order until one successfully validates the token.
+
+### Configuration
+
+Configure each provider under `LAKEKEEPER__OPENID_PROVIDERS__<IDP_ID>__`. This configuration takes precedence over the single-provider `LAKEKEEPER__OPENID_PROVIDER_URI` if both are set.
```
c-thiel: I think we should reword this — OPENID_PROVIDERS adds to, not replaces, OPENID_PROVIDER_URI. The example should also show the typical shape: keep your primary provider on OPENID_PROVIDER_URI (drives the UI), and put EKS/secondary providers under OPENID_PROVIDERS__*.

docs/docs/authentication.md
```diff
@@ -452,6 +452,52 @@ We are now ready to deploy Lakekeeper and login via the UI. Set the following en
...
+
+### When to Use Multiple Providers
+
+Common use cases include:
+
+- **Okta/Entra + EKS OIDC**: Human users authenticate via corporate IdP, while Kubernetes workloads use EKS OIDC tokens
+- **Multi-cluster Kubernetes**: Different EKS/GKE clusters each have their own OIDC provider
+- **Migration scenarios**: Gradually migrating from one IdP to another while both remain active
+
+See the [Configuration Reference](./configuration.md#multiple-oidc-providers) for the full list of available options per provider.
```
c-thiel, line 503: I think we should add a short note on identity continuity here: existing user IDs are formatted as `oidc~<sub>` from the primary provider. Don't move that provider into `OPENID_PROVIDERS` under a different key (e.g. okta), or all previous assignments orphan.

docs/docs/configuration.md
```diff
@@ -230,11 +230,45 @@ Please check the [Authentication Guide](./authentication.md) for more details.
...
 | `LAKEKEEPER__KUBERNETES_AUTHENTICATION_AUDIENCE`                          | `https://kubernetes.default.svc`             | Audiences that are expected in Kubernetes tokens. Only has an effect if `LAKEKEEPER__ENABLE_KUBERNETES_AUTHENTICATION` is true. |
 | `LAKEKEEPER_TEST__KUBERNETES_AUTHENTICATION_ACCEPT_LEGACY_SERVICEACCOUNT` | `false`                                      | Add an authenticator that handles tokens with no audiences and the issuer set to `kubernetes/serviceaccount`. Only has an effect if `LAKEKEEPER__ENABLE_KUBERNETES_AUTHENTICATION` is true. |
 
+#### Multiple OIDC Providers
+
+For advanced scenarios requiring multiple OIDC providers (e.g., Okta for users + EKS OIDC for Kubernetes workloads), configure providers under `LAKEKEEPER__OPENID_PROVIDERS__<IDP_ID>__`. When set, this takes precedence over `LAKEKEEPER__OPENID_PROVIDER_URI`.
+
+The `<IDP_ID>` key is the identity-provider ID used in user IDs like `<idp_id>~<subject>`. It must not contain `~`; `kubernetes` is reserved for the built-in Kubernetes authenticator.
```
c-thiel: I think one sentence here on env-var → key mapping would help: figment lowercases the segment and __ is the separator, so `MY_PROVIDER` becomes key `my_provider` but `MY__PROVIDER` creates nested keys. The example shows uppercase env vars and lowercase IDs in the user-ID format, but the relationship is implicit.

crates/lakekeeper/src/config.rs
```diff
@@ -1065,7 +1149,7 @@ impl DynAppConfig {
     }
 
     pub fn authn_enabled(&self) -> bool {
-        self.openid_provider_uri.is_some()
+        self.openid_provider_uri.is_some() || !self.openid_providers.is_empty()
     }
```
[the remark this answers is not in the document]
c-thiel: How about making this clearer:

```rust
// "Is any authentication active?"  — used by /info to reject anonymous
pub fn authn_enabled(&self) -> bool {
    self.openid_provider_uri.is_some()
        || !self.openid_providers.is_empty()
        || self.enable_kubernetes_authentication
}

// "Does the UI have an SSO target?"  — used by ui.rs
pub fn ui_login_enabled(&self) -> bool {
    self.openid_provider_uri.is_some()
}

```

Then we can use `ui_login_enabled` in ui.rs in lakekeeper-bin.

c-thiel (commented):
I had a change of heart regarding deprecation - see my comments below. I would favor to keep the old configuration (not deprecate it) to prevent breaking existing deployments and keep the notion of a "main" provider for the UI.

crates/lakekeeper/src/config.rs
mithilajoshi: the type is used to represent *any* OIDC provider config (including the primary `OPENID_PROVIDER_URI` config once we build the provider list). Calling it “Additional” would be misleading . I could update the doc comment to clarify it’s the config shape used for **entries in `OPENID_PROVIDERS`**, and that the primary provider uses the same shape internally. Or go with a neutral alternative like OidcProviderSettings or SingleOidcProviderConfig if you'd prefer a rename — just want to avoid the misleading implication. What do you think?

c-thiel:
@mithilajoshi we still need the CLA to be signed. You can find more information on why we need this here:
https://docs.lakekeeper.io/docs/0.10.x/developer-guide/?h=cla#foundation-cla

You can follow the bot: https://github.com/lakekeeper/lakekeeper/pull/1760#issuecomment-4492471902

mithilajoshi:
> @mithilajoshi we still need the CLA to be signed. You can find more information on why we need this here: https://docs.lakekeeper.io/docs/0.10.x/developer-guide/?h=cla#foundation-cla
> 
> You can follow the bot: [#1760 (comment)](https://github.com/lakekeeper/lakekeeper/pull/1760#issuecomment-4492471902)

@c-thiel Do you allow/support corporate CLAs?

c-thiel:
> > @mithilajoshi we still need the CLA to be signed. You can find more information on why we need this here: https://docs.lakekeeper.io/docs/0.10.x/developer-guide/?h=cla#foundation-cla
> > You can follow the bot: [#1760 (comment)](https://github.com/lakekeeper/lakekeeper/pull/1760#issuecomment-4492471902)
> 
> @c-thiel Do you allow/support corporate CLAs?

I am not quite sure what you mean by that but I think yes - its just important that you as the author signs it wheather from a corporate context or not is irrelevant :)

mithilajoshi:
> I am not quite sure what you mean by that but I think yes - its just important that you as the author signs it wheather from a corporate context or not is irrelevant :)

What I am referring to is the part in the CLA that says `that your employer has executed a separate Corporate CLA with Vakamo.` -- How would we go about that? 
```
If your employer(s) has rights to intellectual property that you create that includes your Contributions, 
you represent that you have received permission to make Contributions on behalf of that employer,
that your employer has waived such rights for your Contributions to Vakamo, or that your employer 
has executed a separate Corporate CLA with Vakamo.
```

c-thiel:
> > I am not quite sure what you mean by that but I think yes - its just important that you as the author signs it wheather from a corporate context or not is irrelevant :)
> 
> What I am referring to is the part in the CLA that says `that your employer has executed a separate Corporate CLA with Vakamo.` -- How would we go about that?
> 
> ```
> If your employer(s) has rights to intellectual property that you create that includes your Contributions, 
> you represent that you have received permission to make Contributions on behalf of that employer,
> that your employer has waived such rights for your Contributions to Vakamo, or that your employer 
> has executed a separate Corporate CLA with Vakamo.
> ```

We never needed this so far as most users commit to OSS as themselves and not the company they represent. Would this work for you too? Or does your company own the code your committed? If your company owns the code, then I guess we need to go the extra mile. We don't have a document ready for this.

bechbd:
> We never needed this so far as most users commit to OSS as themselves and not the company they represent. Would this work for you too? Or does your company own the code your committed? If your company owns the code, then I guess we need to go the extra mile. We don't have a document ready for this.


@c-thiel Our company owns the code we develop and commit.  I am authorized to sign on behalf of StitchFix for this and I have signed the CLA.  Is that acceptable or do we need to get a Corporate CLA?

c-thiel:
> > We never needed this so far as most users commit to OSS as themselves and not the company they represent. Would this work for you too? Or does your company own the code your committed? If your company owns the code, then I guess we need to go the extra mile. We don't have a document ready for this.
> 
> 
> 
> 
> 
> @c-thiel Our company owns the code we develop and commit.  I am authorized to sign on behalf of StitchFix for this and I have signed the CLA.  Is that acceptable or do we need to get a Corporate CLA?

That is absolutely acceptable - thank you very much for your contribution!

mithilajoshi:
@c-thiel Are we good to go with this PR then? When would it be available in main?

c-thiel:
@mithilajoshi — pushed follow-ups to your branch:

- Fail-closed fix. With no primary URI and an optional provider that fails JWKS discovery, the server was booting anonymous-open with just a warn! line. Added a guard that refuses to boot when any providers are configured but all get skipped. Regression test included.
- Refactor. Extracted assemble_authenticator_chain so tests run the real production path instead of a parallel copy. Moved OidcProviderConfig from config.rs → service::authn so the bidirectional config ↔ service::authn import becomes one-way.
- Grammar. Tightened idp_id to [a-z0-9-]+ (same shape as RoleProviderId). The HashSet uniqueness check, control-char check, separator check, and case-insensitive reserved-name aliases all collapse into the grammar — half the validator is gone.
- Security docs. Spelled out chain routing semantics (first authenticator where both iss and aud match the token; no fall-through after a signature failure). Flagged audience-strongly-recommended on both single- and multi-provider tables.

And a few nits.

Please review these changes from your side. Good to merge from our side after your review.

mithilajoshi:
Great changes, especially that first fix in the list.  LGTM!

c-thiel:
Thanks for this nice PR @mithilajoshi - we are good to go then!

PR #1770, opened 2026-05-23
fix(view): preserve protection during view commits

### LINKED ISSUE #1769: Protected views lose deletion protection after view commits
**Describe the bug**
Protected views lose deletion protection after a normal view commit / `CREATE OR REPLACE VIEW`.

The current Postgres view commit path deletes the existing `tabular` row and recreates it instead of updating it in place. Because `tabular.protected` is stored on that row and defaults to `false`, a protected view becomes unprotected after a successful view commit.

This makes deletion protection non-durable for views: a user can enable protection, perform a normal metadata update, and silently lose that protection.

**To Reproduce**
Steps to reproduce the behavior, starting from a running Lakekeeper instance (for example `examples/minimal`):
1. Create a view.
2. Set view protection to `true` via `POST /management/v1/warehouse/{warehouse_id}/view/{view_id}/protection`.
3. Verify `GET /management/v1/warehouse/{warehouse_id}/view/{view_id}/protection` returns `protected: true`.
4. Update the view with `CREATE OR REPLACE VIEW ...` from a supported engine, or submit a normal REST view commit.
5. Query the protection endpoint again.

**Expected behavior**
The view should remain protected after a normal view commit. Protection is documented and exposed as deletion protection for views, so metadata evolution should not clear it.

**Actual behavior**
After the commit succeeds, the view's protection state becomes `false`.

**Screenshots**
Not applicable.

**System Information:**
 - Deployment Type: Any
 - Lakekeeper Version: current `main` (also appears to affect earlier versions after protection was introduced)
 - Query Engine: Any supported engine that performs view commits / `CREATE OR REPLACE VIEW`
 - Authentication Provider: Any
 - Authorization Provider: Any
 - OS: Any
 - Browser: N/A

**Additional context**
A few implementation details that make this look deterministic rather than environment-specific:

- `commit_view_impl` currently calls `drop_tabular(..., force = true, ...)` and then `create_view(...)`.
- `drop_tabular(..., force = true, ...)` allows deleting protected tabular rows.
- `create_tabular(...)` inserts a fresh `tabular` row without preserving the previous `protected` value.
- `tabular.protected` is `NOT NULL DEFAULT false`.
- By contrast, table commits update the existing `tabular` row in place and preserve `protected`.

History-wise, this looks like an old design mismatch:
- view commits were introduced before deletion protection existed
- deletion protection for tabulars/views was added later
- the view commit path was not updated to preserve row-level state

A good fix would be to update the existing `tabular` row in place during view commits. A smaller fix would be to at least carry forward the previous `protected` value when recreating the row.

fallintoplace:
## Summary
View commits in the Postgres catalog were implemented as drop-and-recreate operations. Once deletion protection was added at the `tabular` row level, that flow started resetting protected views back to the column default on every commit.

This PR changes view commits to update the existing tabular row in place and rebuild the view metadata children, so deletion protection survives normal view commits.

## Changes
- replace the Postgres view commit drop/recreate flow with an in-place update
- preserve tabular-level state such as `protected`
- clear and rebuild view metadata child rows during commit
- add a regression test covering protected views across commit

## Impact
Protected views now stay protected after `CREATE OR REPLACE VIEW` and normal REST view commits instead of silently becoming unprotected.

## Validation
- `cargo fmt --all`
- `cargo test -p lakekeeper test_commit_view --no-run`

I also attempted the runtime test invocation, but the local environment does not currently provide a usable Postgres test backend for `sqlx::test` (`DATABASE_URL` unset and no running Docker/Postgres service).

Closes #1769

## Summary by CodeRabbit

* **New Features**
  * Faster, batched metadata writing for views and more robust atomic view commits.
  * Ensures view properties are immediately visible after creation.

* **Bug Fixes**
  * Preservation of view protection when committing updates.
  * Concurrent-update checks to reject stale commits.
  * Detection and rejection of storage-location collisions for tabular/view objects.

* **Tests**
  * Added tests validating protection preservation, concurrent-commit rejection, location collision handling, and metadata repopulation.

[Review Change Stack](https://app.coderabbit.ai/change-stack/lakekeeper/lakekeeper/pull/1770?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


c-thiel:
@fallintoplace thank you very much for finding, reporting and creating this PR!
I just reviewed the PR and fixed my comments in [abcdd79](https://github.com/lakekeeper/lakekeeper/pull/1770/commits/abcdd79447a704f099f6e5d7b42d544202d1858b) - feel free to review them too.
I'll do another pass too - afterwards we should be good to merge!

fallintoplace:
All good from my side. @c-thiel

PR #1773, opened 2026-05-24
fix(kv2): replace stale vault health status

### LINKED ISSUE #1772: KV2/Vault health entries accumulate and can keep /health unhealthy after recovery
**Describe the bug**
When Lakekeeper uses the KV2/Vault secret backend, the Vault health state is appended on every check instead of being replaced. As a result, a transient Vault failure can keep `/health` unhealthy even after Vault recovers.

This also causes the in-memory health vector to grow for the lifetime of the process.

**To Reproduce**
1. Configure Lakekeeper to use the KV2/Vault secret backend.
2. Start Lakekeeper and allow the periodic health check loop to run.
3. Make Vault temporarily unavailable so one KV2 health check fails.
4. Restore Vault and let subsequent health checks succeed.
5. Query `/health`.

**Expected behavior**
After Vault recovers, the KV2/Vault provider should report its current health state, and `/health` should become healthy again if all providers are healthy.

**Actual behavior**
Old KV2/Vault health entries are retained. Since `/health` folds over all stored entries for a provider, one stale `Unhealthy` record can keep the overall service unhealthy after recovery.

**Screenshots**
Not applicable.

**System Information:**
 - Deployment Type: Any deployment using KV2/Vault as secret backend
 - Lakekeeper Version: current `main`
 - Query Engine: Any / N/A
 - Authentication Provider: Any
 - Authorization Provider: Any
 - OS: Any
 - Browser: N/A

**Additional context**
This looks like an implementation inconsistency rather than an intentional design choice.

Current behavior:
- `crates/lakekeeper/src/implementations/kv2.rs` appends with `push(Health::now(...))` on every update
- `crates/lakekeeper/src/service/health.rs` computes provider health by checking all stored entries
- Postgres and OpenFGA health implementations clear and replace their vectors on each update instead of accumulating history

History-wise:
- the shared health collector was introduced in `Service health checks` (#181)
- KV2 support landed shortly after in `feat(secrets): add support for kv2 secret storage` (#192)
- the append-only Vault health behavior appears to date back to that original KV2 implementation

A straightforward fix would be to make `SecretsState::update_health()` replace the current Vault health entry on each update, matching the Postgres/OpenFGA pattern.

fallintoplace:
## Summary
KV2/Vault health checks were appending a new `vault` entry on every update instead of replacing the current status.

Because `/health` folds over every stored status for a provider, one stale `Unhealthy` entry could keep the service unhealthy after Vault recovered. The in-memory health vector also grew over time.

## What changed
- replace the KV2 health entry on each update instead of appending
- add a regression test that verifies the KV2 provider keeps only the current Vault status

## Impact
Lakekeeper now reports the current Vault health state correctly when using the KV2 secret backend, rather than retaining old failures indefinitely.

## Validation
- `cargo fmt --all`
- `cargo test -p lakekeeper test_set_health_status_replaces_previous_entry`

Closes #1772

## Summary by CodeRabbit

* **Bug Fixes**
  * Health tracking now replaces prior vault health entries with the latest status, preventing accumulation of duplicate entries.

* **Tests**
  * Added a unit test verifying that updating vault health replaces the previous entry and stores only the most recent status.

[Review Change Stack](https://app.coderabbit.ai/change-stack/lakekeeper/lakekeeper/pull/1773?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/crates/lakekeeper/src/implementations/kv2.rs
+++ b/crates/lakekeeper/src/implementations/kv2.rs
@@ -217,12 +217,6 @@ impl SecretsState {
             }
         }
     }
-
-    async fn set_health_status(&self, status: HealthStatus) {
-        let mut lock = self.health.write().await;
-        lock.clear();
-        lock.extend([Health::now("vault", status)]);
-    }
 }
 
 #[async_trait]
@@ -237,14 +231,20 @@ impl HealthExt for SecretsState {
         match t {
             Ok(_) => {
                 tracing::debug!("Vault is healthy");
-                self.set_health_status(HealthStatus::Healthy).await;
+                set_vault_health(&self.health, HealthStatus::Healthy).await;
             }
             Err(err) => {
                 tracing::error!(?err, "Vault is unhealthy");
-                self.set_health_status(HealthStatus::Unhealthy).await;
+                set_vault_health(&self.health, HealthStatus::Unhealthy).await;
             }
         }
     }
+}
+
+async fn set_vault_health(health: &Arc<RwLock<Vec<Health>>>, status: HealthStatus) {
+    let mut lock = health.write().await;
+    lock.clear();
+    lock.extend([Health::now("vault", status)]);
 }
 
 impl std::fmt::Debug for SecretsState {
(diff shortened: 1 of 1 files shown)


c-thiel (approved):
Very good finding - thanks for the PR @fallintoplace!

PR #1802, opened 2026-06-02
fix: return 503 for unhealthy health endpoint

corleyma:
## Summary

This fixes /health so its HTTP status matches the aggregate health state:

- 200 OK when aggregate health is ok
- 503 Service Unavailable when aggregate health is error or unknown

The JSON response body is unchanged.

This also includes two related health-check fixes:

- bounds periodic provider health updates by the configured health-check interval, so one stuck provider update cannot block future refreshes or shutdown forever
- makes bare `lakekeeper healthcheck` check the server /health endpoint by default instead of succeeding without checking anything

## Motivation

The monitoring docs describe /health as database-dependent and suitable for Kubernetes probes. Before this change, the endpoint always returned HTTP 200 because it serialized the health body directly, even when the aggregate body reported an unhealthy state.

Kubernetes HTTP probes only evaluate the HTTP status code. That meant a probe configured against /health could continue passing while Lakekeeper reported unhealthy database-backed health in the response body.

This PR makes the documented probe behavior true.

## Compatibility

The /health JSON response shape is preserved. Clients that inspect the body can continue doing so. Clients and probes that rely on HTTP status codes now receive a non-2xx status for unhealthy or unknown aggregate health.

## Testing

- `cargo +nightly fmt --all -- --check`
- `cargo test -p lakekeeper health_endpoint`
- `cargo test -p lakekeeper update_health_loop_exits_after_a_provider_update_times_out`
- `cargo test -p lakekeeper-bin`
- `cargo clippy -p lakekeeper --all-targets --all-features -- -D warnings`
- `cargo clippy -p lakekeeper-bin --all-features -- -D warnings`

## Summary by CodeRabbit

* **New Features**
  * `/health` endpoint now maps overall status to HTTP codes: 200 when healthy, 503 when unhealthy/unknown.
  * Health update loop now bounds provider updates to avoid hanging and allows timely cancellation.

* **Bug Fixes**
  * Healthcheck CLI normalizes flag combinations so at least the server check runs when none specified; help text clarified.

* **Documentation**
  * Expanded monitoring/production guidance with Kubernetes probe examples and expected status codes.


c-thiel:
This absolutely makes sense @corleyma - thanks for the PR!
I fixed the flaky test and also made shutdown more robust in https://github.com/lakekeeper/lakekeeper/pull/1802/commits/431e2abcfa739ec31315a5420c50fbf29a60aa3e.

c-thiel (approved):

PR #1849, opened 2026-06-16
fix(roles): evict membership caches on role source-system rebind

c-thiel:
Re-pointing a role's source system (`PUT /role/{id}/source-system`) changes its `RoleIdent` (provider_id/source_id) but left the in-process caches untouched. That ident is cached per row in every assignee's USER_ASSIGNMENTS closure (`AssignedRole.role_ident`) and in the role's ROLE_MEMBERS entry. External authorizers (e.g. Cedar) key on the ident, so until those entries expired (up to the cache TTL) they evaluated the stale binding — a correctness/security gap on the rebind path.

The handler now mirrors `delete_role`: read the affected-user closure pre-commit on the rebind transaction (so a failed read rolls the rebind back), then evict the user-assignment closures and the role's member entry immediately after commit. Unlike delete, the rows are updated rather than cascade-deleted, so the affected set is identical pre- and post-commit. ROLE_CACHE is refreshed separately by the `role_updated` event the handler already emits.

Local-replica eviction only; cross-replica freshness stays bounded by TTL as before.

Adds `source_system_rebind_evicts_user_assignment_and_member_closures`, asserting both closures reflect the new ident (and return fresh `Arc`s) after a rebind.

## Summary by CodeRabbit

* **Tests**
  * Added integration test for cache invalidation when role source systems are updated.

* **Bug Fixes**
  * Fixed cache invalidation to ensure user assignments and role membership data are properly refreshed when a role's source system is updated.

### DIFF
--- a/crates/lakekeeper-integration-tests/tests/role_membership_ops.rs
+++ b/crates/lakekeeper-integration-tests/tests/role_membership_ops.rs
@@ -997,3 +997,109 @@ async fn transitive_role_member_of(pool: PgPool) {
         ["A".to_string(), "B".to_string()].into_iter().collect()
     );
 }
+
+// ==================== source-system rebind invalidates closures ====================
+
+/// Rebinding a role's source system changes its `RoleIdent`, which is cached per
+/// row in every assignee's USER_ASSIGNMENTS closure and in the role's ROLE_MEMBERS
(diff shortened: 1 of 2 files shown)


c-thiel:
@coderabbitai review

PR #1858, opened 2026-06-17
fix: clean up empty namespace folders on storage after drop

### LINKED ISSUE #1064: Deleted namespaces may leave traces on filesystems
On storage types that follow filesystem semantics, like hdfs or adls, we end up leaving empty folders behind when we drop the namespace. We create the folder after creating the first table within a namespace. Upon dropping the namespace, we do not delete the folder. While the prefix simply disappears on non-filesystem object stores like s3, we have those remnants in hdfs and adls.

1. create namespace
2. create table within namespace
3. drop table
4. drop namespace
5. go to filesystem and see that the namespace folder still exists

We cannot simply issue a delete location upon drop since we may want to support moving tables between namespaces at a later point. 

We should check that the location is empty before deleting it.

dadavidtseng:
## What does this PR do?

Adds best-effort storage cleanup for namespace folders when a namespace is dropped.
On filesystem-based storage backends (HDFS, ADLS), dropping a namespace previously
left empty folders behind. This PR wires the namespace drop path to check whether the
namespace's storage folder is empty after the database transaction commits, and removes
it if so.

## Why was this PR needed?

Issue #1064 reported that on storage backends with filesystem semantics (HDFS, ADLS),
dropping a namespace leaves remnant empty folders. While object stores like S3 handle
this implicitly (prefixes disappear), true filesystems retain the empty directory.

Investigation confirmed that the `drop_namespace` code path in `server/namespace.rs`
only handled database deletion and authorizer cleanup — it never interacted with the
storage backend. The table drop path already had storage cleanup via `TabularPurgeTask`,
but this pattern was never applied to namespace folders.

## What are the relevant issue numbers?

Closes #1064

## Implementation details

1. **Extended `NamespaceDropInfo`** with `namespace_locations: Vec<(NamespaceId, Location)>`
   to carry each dropped namespace's storage location through the drop pipeline.

2. **Updated the PostgreSQL `drop_namespace` SQL query** to fetch
   `namespace_properties->>'location'` for the target namespace and all child namespaces
   (for recursive drops).

3. **Added `try_cleanup_namespace_locations()` helper** that performs best-effort cleanup:
   - Gets storage IO via the warehouse's storage profile and secret
   - For each namespace location, checks if the folder is empty using `is_empty()`
   - If empty, removes the folder using `remove_all()`
   - Errors are logged and swallowed — cleanup must not fail the drop operation

4. **Wired cleanup into both drop paths**: non-recursive and recursive.

5. **Added 3 integration tests** covering:
   - Empty namespace folder is cleaned up after drop
   - Non-empty namespace folder is preserved
   - Recursive drop correctly removes namespace from catalog

## Does this PR meet the acceptance criteria?

- [x] Tests added for new/changed behavior
- [x] All tests passing
- [x] Follows project style guide (conventional commits, clippy clean)
- [x] No breaking changes introduced
- [x] Documentation updated (N/A — no user-facing docs needed)

## Summary by CodeRabbit

* **New Features**
  * Namespace deletion with purge now performs best-effort removal of empty namespace storage directories after catalog and authorization updates.
  * Cleanup runs for both non-recursive and recursive purge flows.
  * Cleanup is resilient: invalid location entries are skipped with a warning, and cleanup failures are logged without blocking deletion.

* **Tests**
  * Added integration tests for: empty-directory cleanup on non-recursive purge, retention when storage isn’t empty, and recursive purge where only catalog removal is asserted (storage cleanup not validated).


dadavidtseng:
Hi @c-thiel @twuebi — this is my first contribution to lakekeeper. I've implemented storage cleanup for namespace folders on drop, as described in #1064. The folder is only removed if it's empty (checked via is_empty()), and cleanup is best-effort so it never fails the drop. Would appreciate a review when you have time!

dadavidtseng:
Fixed — replaced ?-propagating collect with filter_map that logs and skips unparseable locations. Cleanup metadata parse failures no longer abort the drop operation. Updated in commit 2f439a37.

c-thiel:
Thank you for your contribution @dadavidtseng!
Please note that we recently changed our default storage layout to not have namespace folders anymore (https://github.com/lakekeeper/lakekeeper/pull/1853). This will be helpful as we are adding a "move" Operation soon.

Nontheless this is a useful feature for folks using custom `StorageLayout`s! We'll try to get a review in shortly.

dadavidtseng:
@c-thiel Thank you for the context on #1853 and the upcoming move operation! I'm glad this will still be useful for custom StorageLayouts. Looking forward to the review.

c-thiel:
Thanks for picking this up — empty namespace folders on HDFS/ADLS have been a real, long-standing gap. The shape is good: `is_empty` before delete, errors logged-and-swallowed so cleanup never fails the drop, `filter_map` skipping unparseable locations, and going through the `LakekeeperStorage` trait so it's storage-profile-agnostic.

Let's keep this best-effort and inline (a dedicated task queue would be overkill here) — but two things need fixing, and the limitations should be stated loudly so nobody expects more than it delivers.

1. Don't let it delete the warehouse base.
On the default tabular-only/flat layout, `render_namespace_path` returns vec![] (`storage_layout.rs:273`), so the persisted namespace location equals the warehouse base. That makes `is_empty(base)` → `remove_all(base)` capable of wiping the warehouse root when it happens to be empty. Please gate on the persisted location, not the current layout (the layout can be switched, and the location is a create-time snapshot — pre-0.13 namespaces even carry a `{uuid}` dir): only act when `loc != base && loc.starts_with(base)` (strictly below the base). Fails closed, layout-switch-safe.

2. Deepest-first ordering for recursive drops.
The SQL orders by namespace_id (arbitrary), but nested child folders sit under their parents, so a parent is often checked before its children → is_empty false → skipped. Order by location depth/length descending so children go first. Cheap, and it's the difference between recursive cleanup working or not on full-hierarchy.

3. Make the tests honest.
They build a separate `MemoryStorage::new()` rather than the warehouse's IO handle, so `test_drop_empty_namespace_cleans_up_folder` only asserts a catalog 404 — not that the folder was removed; the recursive test asserts no storage state. Pulling IO from `warehouse.storage_profile.file_io(...)` and asserting on the location after drop would actually cover #1/#2. (The tokio::time::sleep "give async cleanup time" calls are no-ops, since cleanup is inline — fine to drop them.)

4. Document the limitations — in a doc comment on `try_cleanup_namespace_locations` and a line in the PR description from which we generate the release notes:

It only removes a folder that is already empty at the moment of the drop. Table-data purge is asynchronous, so dropping a namespace soon after its tables (and any recursive drop that still contains live tables) will find the folder non-empty and skip it. In practice it cleans up namespaces whose data was purged earlier. It's a single attempt with no retry — a skipped folder isn't revisited. 

That way it does what it can, never does harm, and sets the right expectation. The core idea is solid — mostly about scoping the location it touches and being upfront about when it fires.

dadavidtseng:
@c-thiel Thank you for the thorough review! All three points make sense — I'll address them:

1. Guard against deleting the warehouse base by checking loc != base && loc.starts_with(base)
2. Order locations by path length descending for deepest-first cleanup
3. Fix tests to use the warehouse's actual IO handle and assert on real storage state

I'll also add doc comments documenting the limitations (single attempt, no retry, depends on prior data purge). Will push the updates shortly.

dadavidtseng:
@c-thiel Thanks for the detailed review! I've addressed all three points:

1. Base-location guard: Added a check that skips cleanup when location == base || !location.starts_with(base) — strictly below only. Also added doc comments covering the limitations (single attempt, depends on prior purge, errors swallowed).
2. Deepest-first ordering: Changed both ORDER BY clauses to length(n.namespace_properties->>'location') DESC, n.namespace_id so children are cleaned before parents in recursive drops. Updated the .sqlx offline cache accordingly.
3. Honest tests: Tests now use the shared thread-local MemoryStorage::new() (same backing store as file_io()) and assert actual storage state after drop — not just catalog 404. Removed all sleep calls.

All tests pass locally, clippy and fmt are clean.

crates/lakekeeper/src/server/namespace.rs
```diff
@@ -848,6 +864,102 @@ async fn try_recursive_drop<A: Authorizer, C: CatalogStore>(
...
+            );
+            return;
+        }
+    };
+
+    for (ns_id, location) in namespace_locations {
+        // Guard: never delete the warehouse base itself or locations outside it.
+        // On flat/default layouts the persisted location equals the base, and on
+        // layout switches the snapshot may no longer match the current layout.
+        if *location == base || !location.as_str().starts_with(base.as_str()) {
```
c-thiel: `starts_with` isn't trailing-slash safe; S3's `base_location()` has no trailing slash, so `s3://bucket/wh` matches sibling `s3://bucket/wh-other/....`. We should use the existing helper:


`if *location == base || !location.is_sublocation_of(&base) {`

crates/lakekeeper-integration-tests/tests/namespace_storage_cleanup.rs
```diff
@@ -0,0 +1,252 @@
...
+        )
+        .await
+        .unwrap();
+
+        // The file should still exist — cleanup skipped because folder was non-empty.
+        let content = storage.read(&file_path).await;
+        assert!(
+            content.is_ok(),
+            "File should still exist because the namespace folder was not empty"
+        );
```
c-thiel: the marker is deleted at :108-113, so the prefix is already empty before drop, and MemoryStorage has no empty-dir entity → the assertion passes even if cleanup never runs. Assert remove_all was called, or rename to reflect it only covers the guard path.

crates/lakekeeper/src/server/namespace.rs
```diff
@@ -578,6 +581,14 @@ impl<C: CatalogStore, A: Authorizer + Clone, S: SecretStore>
...
                 })
                 .ok();
+
+            // Best-effort cleanup of namespace storage folders
+            try_cleanup_namespace_locations(
+                &warehouse,
+                &state.v1_state.secrets,
+                &drop_info.namespace_locations,
+            )
+            .await;
```
c-thiel: We should probably not run this if `flags.purge` is explicitly set to false.

crates/lakekeeper/src/server/namespace.rs
```diff
@@ -848,6 +864,102 @@ async fn try_recursive_drop<A: Authorizer, C: CatalogStore>(
...
+        if *location == base || !location.as_str().starts_with(base.as_str()) {
+            tracing::debug!(
+                "Skipping cleanup for namespace {ns_id}: location {location} is at or outside warehouse base"
+            );
+            continue;
+        }
+
+        match crate::service::storage::is_empty(&file_io, location).await {
+            Ok(true) => {
+                if let Err(e) = super::io::remove_all(&file_io, location).await {
```
c-thiel, line 963: I don't think we should remove recursive. I also not sure if it works because it lists recursive and deletes then.

We should gate the whole functionality to Storages that are hierarchical - should probably expose that via the Trait.

c-thiel (commented):
@dadavidtseng are you sure this works currently? I believe the detection `is_empty` -> `remove_all` pretty much cancels out as `remove_all` uses the same list internally. We definitly need a test for the happy path.

dadavidtseng:
@c-thiel Thanks for the follow-up review! Addressed all four points:

1. `is_sublocation_of` for trailing-slash safety
2. Honest tests reflecting MemoryStorage limitations
3. `flags.purge` gate on both drop paths
4. `StorageProfile::is_hierarchical()` gate — S3/GCS/Memory skip cleanup entirely

The one open design question is whether to replace `remove_all` with a dedicated trait method for directory removal. I've left it as-is for now since ADLS's `remove_all` override uses native `DELETE ?recursive=true` which does remove the directory entry correctly, but happy to add a narrower `delete_directory` method if you'd prefer that.

crates/lakekeeper/src/server/namespace.rs
dadavidtseng: Fixed, now using `location.is_sublocation_of(&base)` which handles the trailing-slash edge case correctly.

crates/lakekeeper-integration-tests/tests/namespace_storage_cleanup.rs
dadavidtseng: You're right, MemoryStorage has no directory entities, so the old assertion was vacuous. With the `is_hierarchical()` gate (see comment 4), cleanup is correctly skipped for non-hierarchical backends. Renamed the tests and updated comments to honestly reflect they only cover the catalog-side contract.

crates/lakekeeper/src/server/namespace.rs
dadavidtseng: Agreed, added `if flags.purge { ... }` around both call sites (non-recursive and recursive paths).

crates/lakekeeper/src/server/namespace.rs
dadavidtseng: Added `StorageProfile::is_hierarchical()` (returns `true` for ADLS/`OneLake`, `false` for S3/GCS/Memory) and gated the entire cleanup behind it, so object stores never hit this path.
On the `remove_all` concern, I agree a recursive delete is a heavier hammer than needed for removing an empty directory. The `is_empty` guard prevents data loss, but the abstraction isn't ideal. Since adding a trait method (e.g. `delete_directory`) to `LakekeeperStorage` is a public API change, I'd appreciate your guidance on how you'd like it shaped. Happy to implement whichever direction you prefer.

dadavidtseng:
@c-thiel Friendly ping, I've addressed all four review points from your last round (is_sublocation_of, purge gate, hierarchical gate, honest tests). Would you have a chance to re-review? Also still wondering about your preference on the `delete_directory` trait method vs `remove_all` question. Happy to split that out into a follow-up PR if you'd prefer.

PR #1886, opened 2026-07-02
fix(authz): stop evaluating discarded Select on target view load

c-thiel:
Emit Select only for intermediate referenced-by views (where it gates DEFINER chain traversal); the target view's Select was always discarded by interpret_authz_results_for_load_view but still cost an authorizer evaluation and logged a misleading SelectView decision. Target is now identified by ident to stay in lockstep with the consumer.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved authorization handling for loading tables and views, especially in view chains.
  * Target views now require only metadata access, while intermediate views apply stricter checks during traversal.
  * Updated behavior for delegated execution so authorization results are more accurate and consistent.

* **Tests**
  * Refreshed test coverage to match the updated table/view authorization behavior.


c-thiel:
@coderabbitai full review

Issue #1878, opened 2026-06-29
Support OpenTelemetry trace for observability


fm100:
**Is your feature request related to a problem? Please describe.**
Lakekeeper exposes prometheus metrics endpoint but often times it is not enough. Having OpenTelemetry trace support would be great for the application observability. OpenFGA already supports OpenTelemetry trace, and adding `traceparent` to the HTTP header and/or gRPC metadata when calling openfga API will allow full trace across the services. The caller of lakekeeper API might contain `traceparent` header as well and would be great to extract and reuse it for trace connectivity.

**Describe the solution you'd like**
OpenTelemetry trace is sent when OpenTelemetry SDK standard environment variables are configured:
* `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT=http://127.0.0.1:4318/v1/traces`
* `OTEL_SERVICE_NAME=lakekeeper`

**Describe alternatives you've considered**
Enabling sentry is also an option but OpenTelemetry is the open standard, and sentry accepts open telemetry traces.

**Additional context**

hiSandog:
It would help to define whether trace context must propagate into background catalog or storage work that outlives the initiating request. Compatibility coverage should include requests with W3C `traceparent`, requests without tracing headers, and tracing disabled entirely.

PR #1889, opened 2026-07-04
feat(tracing): add OpenTelemetry trace export

fm100:
Lakekeeper previously emitted tracing spans only through JSON logs, making it difficult to follow requests across services or inspect backend latency.

- Export server traces over OTLP HTTP/protobuf when an OTLP endpoint is configured; preserve JSON-only logging for other commands.
- Continue incoming W3C trace context and mark HTTP request spans as server spans.
- Add child spans for PostgreSQL transactions and OpenFGA client operations.
- Batch exports off the request path, validate endpoint configuration, and flush pending spans during shutdown.
- Document the OpenTelemetry environment variables and add propagation and exporter regression tests.

This PR implement #1878 

## Summary by CodeRabbit

* **New Features**
  * Added optional OpenTelemetry tracing with OTLP HTTP export for server operations.
  * Added end-to-end request trace linking through W3C trace context headers.
  * Added detailed tracing for authorization and database transaction operations.
* **Bug Fixes**
  * Improved handling of invalid or missing trace headers.
  * Invalid telemetry configuration now prevents startup with clear validation.
* **Documentation**
  * Expanded logging documentation with telemetry setup and configuration details.
* **Tests**
  * Added coverage for trace context propagation, endpoint configuration, and telemetry export.


fm100:
image

Sample of collected trace.

fm100:
Hmm... I merged the main and tried `cargo clippy` but it passes for me. I have no idea why it failed from CI.

c-thiel:
Hey @fm100, thank you very much for this PR! I think this is a solid foundation. I have two remarks:
1. In its current state we cannot correlate logs with traces. I believe we should probably inject `trace_id` to logs too?
2. This is only partially solving https://github.com/lakekeeper/lakekeeper/issues/1878 as we are not yet forwarding `traceparent` to OpenFGA gRPC metadata. So I think we should have a follow-up.

Let me know your thoughts on 1. Once that is settled I would do another more detailed pass.

crates/lakekeeper-storage-postgres/src/lib.rs
```diff
@@ -111,24 +135,45 @@ impl Transaction<CatalogState> for PostgresTransaction {
...
+            db.operation.name = "commit"
+        )
+    )]
     async fn commit(self) -> Result<()> {
-        self.transaction
-            .commit()
-            .await
-            .map_err(|e| e.into_error_model("Error committing transaction".to_string()))?;
+        self.transaction.commit().await.map_err(|e| {
+            tracing::error!(error = ?e, "Failed to commit PostgreSQL transaction");
```
c-thiel, line 157: `error!` is the wrong level here. A failed commit is routine in this codebase, not a backend fault:

- We rely on unique-index / constraint violations as normal control flow — `dbutils.rs`, `role.rs`, `warehouse.rs` all map `is_unique_violation()` / `constraint()` to expected outcomes (RoleNameAlreadyExists, RoleSourceIdConflict, 409s).
- There are `DEFERRABLE INITIALLY DEFERRED` constraint triggers (migrations/20250108131004_statistics.sql), which fire at COMMIT — so a benign conflict genuinely surfaces right here.
- Serialization failures (40001) / deadlocks under load also surface at commit.

This logs before the upstream layer classifies the error, so a normal conflict gets an ERROR log line. Worse, given `with_error_events_to_status(true) + with_error_records_to_exceptions(true)` in `telemetry.rs`, it also flips this span's status to Error and records an exception — routine 409s will look like backend failures in the trace waterfall.

I think we should drop this error! (or downgrading to debug!) and letting the layer that knows whether it's a 409 vs a 500 log it once. The span still records latency + propagates the error either way.

Same for rollback.

c-thiel (commented):
(retracted)

Issue #1893, opened 2026-07-14
Add Alibaba Cloud OSS storage support


yoogoc:
**Is your feature request related to a problem? Please describe.**
Lakekeeper supports vending downscoped temporary credentials for AWS S3 and Cloudflare R2, but not for Alibaba Cloud OSS. While OSS is S3-compatible on the data plane, it cannot currently be used with credential vending because its STS does not speak AWS SigV4 — so users on Alibaba Cloud have no way to get short-lived, location-scoped credentials for their tables.

**Describe the solution you'd like**
Add a new aliyun-oss S3 credential type that vends downscoped temporary credentials for Alibaba Cloud OSS:

- Signs and issues AssumeRole requests against the Alibaba Cloud STS API using its RPC signing scheme (HMAC-SHA1 over a canonicalized query string), since the AWS SDK STS client cannot be reused.
- Derives the regional STS endpoint from the profile region, or honours an explicit sts-endpoint override.
- Downscopes credentials to the table location via a RAM policy using acs:oss resource names and OSS actions (GetObject/PutObject/AbortMultipartUpload/DeleteObject, gated by the requested StoragePermissions).
- Resolves the role to assume from sts-role-arn, falling back to assume-role-arn.
- Data-plane access reuses the existing S3 access-key path.

Unit tests cover STS response parsing, percent-encoding, timestamp formatting, and the RPC signature against the documented Alibaba Cloud example. The management OpenAPI specs are regenerated for the new credential type.

**Describe alternatives you've considered**
- Reusing the AWS SDK STS client — not possible: Alibaba Cloud STS uses the RPC signing scheme (SignatureVersion=1.0, HMAC-SHA1), not SigV4.
- A dedicated OSS storage backend — unnecessary duplication. OSS is S3-compatible on the data plane, so only the credential-vending (STS) path needs to differ; everything else extends the existing S3 profile.
- Pulling in an Alibaba Cloud SDK — avoided to keep the dependency footprint small. The RPC signing is implemented directly with hmac + sha1 + base64.

**Additional context**
New credential type is aliyun-oss, reusing the S3AccessKeyCredential shape (access key + secret). Requires either sts-role-arn or assume-role-arn (an Alibaba Cloud RAM role ARN) to be set on the storage profile.

yoogoc:
Implemented, thanks @v-kessler @brgr-s

brgr-s:
@yoogoc many thanks to you for your contribution

PR #1894, opened 2026-07-14
feat(storage): add Alibaba Cloud OSS credential support

yoogoc:
https://github.com/lakekeeper/lakekeeper/issues/1893

## Summary by CodeRabbit

* **New Features**
  * Added `aliyun-oss` support for storage credentials, enabling Alibaba Cloud OSS temporary credential vending.
  * Introduced an Alibaba STS `AssumeRole` flow for access-key/secret-key based authentication.
  * Added support for the new credential type and improved role configuration handling.
  * Generates scoped access policies with safer path handling to avoid unintended wildcard/question-mark entries.
* **Documentation**
  * Updated management API OpenAPI specs to document the new `aliyun-oss` credential option.
* **Tests**
  * Added unit coverage for Alibaba STS AssumeRole response parsing, signing/timestamp formatting, percent-encoding, and policy scoping.


yoogoc:
@c-thiel could you review the PR

crates/lakekeeper/src/service/storage/s3.rs
```diff
@@ -755,6 +762,187 @@ impl S3Profile {
...
+        Ok(aws_sdk_sts::types::Credentials::builder()
+            .access_key_id(creds.access_key_id)
+            .secret_access_key(creds.access_key_secret)
+            .session_token(creds.security_token)
+            .expiration(
+                (std::time::SystemTime::now() + std::time::Duration::from_secs(ttl_seconds)).into(),
+            )
+            .build()
+            .unwrap())
+    }
```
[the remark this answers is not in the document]
brgr-s: is there a reason why the returned `Expiration` value is not used? Looking at the test `test_deserialize_aliyun_assume_role_response` it seems that a successfull response would return it?

crates/lakekeeper/src/service/storage/s3.rs
brgr-s, line 354: there is an explicit `normalize_r2` call for `CloudflareR2` credentials. I think we might need a `normalize_oss` function that sets flavor to `s3compat` and checks if an endpoint is set. Otherwise user could set flavor to `AWS` and if `endpoint` is not provided, it would be set to the default AWS endpoint template, leading to IO errors while the `AssumeRole` call itself works. `normalize_r2` also always sets `sts_enabled=true`

brgr-s: see comment above. if `sts_enabled` is set to false, and we have no `normalize_oss` to enable it, we have `can_use_vended_credentials==false` even if can use sts

crates/lakekeeper/src/service/storage/s3.rs
```diff
@@ -755,6 +762,235 @@ impl S3Profile {
...
+        &self,
+        sts_request: &ShortTermCredentialsRequest,
+        cred: S3AccessKeyCredential,
+    ) -> Result<aws_sdk_sts::types::Credentials, CredentialsError> {
+        let role_arn = self
+            .sts_role_arn
+            .as_ref()
+            .or(self.assume_role_arn.as_ref())
+            .ok_or_else(|| {
+                CredentialsError::Misconfiguration(
```
brgr-s, line 795: this should also be enforced at `normalize_oss` time.

crates/lakekeeper/src/service/storage/s3.rs
```diff
@@ -755,6 +762,235 @@ impl S3Profile {
...
+                (std::time::SystemTime::now() + std::time::Duration::from_secs(ttl_seconds)).into(),
+            )
+            .build()
+            .unwrap())
+    }
+
+    /// Determine the Alibaba Cloud STS endpoint. Honours an explicit `sts_endpoint` override,
+    /// otherwise derives the regional endpoint from the profile `region`
+    /// (e.g. `https://sts.cn-hangzhou.aliyuncs.com`).
+    fn aliyun_sts_endpoint(&self) -> String {
```
brgr-s, line 904: Maybe add a unit test for this?

crates/lakekeeper/src/service/storage/s3.rs
```diff
@@ -755,6 +762,235 @@ impl S3Profile {
...
+                "oss:DeleteObject",
+            ],
+        };
+
+        // Bucket-scoped, prefix-independent actions. `oss:GetBucketInfo` is always needed (clients
+        // discover the bucket region with it). `oss:ListMultipartUploads` (`GET /?uploads`) lists
+        // in-progress uploads for the whole bucket, so — like `ListObjects` — it must be granted on
+        // the bucket resource, not the object prefix. It is only needed on write-capable
+        // permissions.
+        let mut bucket_actions = vec!["oss:GetBucketInfo"];
```
brgr-s: Do we really need `GetBucketInfo`? It seems to be a oss specific thing, I don't think a "normal" iceberg client using s3 sdk would try to use this.

AWS allows `GetBucketLocation` for client region discovery (see line 1239), and oss seems to have that permission as well, so if a "normal" iceberg client would try to get the region, it would probably need `GetBucketLocation`?

crates/lakekeeper/src/service/storage/s3.rs
```diff
@@ -755,6 +762,235 @@ impl S3Profile {
...
+        if matches!(
+            storage_permissions,
+            StoragePermissions::ReadWrite | StoragePermissions::ReadWriteDelete
+        ) {
+            bucket_actions.push("oss:ListMultipartUploads");
```
brgr-s: as stated in the comment, this is granted unconditionally on the whole bucket, so a client could read in flight multipart-uploads for every prefix in that bucket. AWS does not set that permission (see lines 1216 to 1248). I couldn't find an authorative list which permissions are needed, but in this "minimal" example here https://stackoverflow.com/questions/75559892/what-s3-iam-permissions-are-needed-to-write-a-parquet-file-with-spark I just see "normal" `Put` and no `ListMultipartUpload`. Is this a specific thing for a client using  oss sdk?

crates/lakekeeper/src/service/storage/s3.rs
```diff
@@ -1434,6 +1784,219 @@ pub(crate) mod test {
...
+        )
+        .unwrap_err();
+        assert!(
+            matches!(err, CredentialsError::ShortTermCredential { .. }),
+            "expected ShortTermCredential error, got: {err:?}"
+        );
+    }
+
+    #[test]
+    fn test_deserialize_aliyun_assume_role_response_uses_returned_expiration() {
```
brgr-s: this does not actually test if the return expiration is used, it just tests if a string can be deserialized using aws sdk to the expected value in seconds.

brgr-s (changes requested):
Thank you for your PR @yoogoc, we appreciate the work and would love to support using Alibaba Cloud OSS.

I left a few review comments.

One thing that is missing from the comments I left in the code: if we have support for Alibaba OSS Cloud we should also mentions this in the documentation and f. ex. explain required and optional parameters. See f. ex. the documentation for R2 in `docs/docs/storage.md` lines 560-603. It would be really great if you could a little bit of documentation there.

crates/lakekeeper/src/service/storage/s3.rs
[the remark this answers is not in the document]
brgr-s: @yoogoc question was meant for you, sorry

crates/lakekeeper/src/service/storage/s3.rs
yoogoc: Done — added `normalize_oss()` that forces `flavor = s3-compat`, requires endpoint, and sets `sts_enabled = true`.

crates/lakekeeper/src/service/storage/s3.rs
yoogoc: Added already

crates/lakekeeper/src/service/storage/s3.rs
yoogoc: You're right. Replaced `oss:GetBucketInfo` with `oss:GetBucketLocation` to mirror the AWS policy's region-discovery statement.

crates/lakekeeper/src/service/storage/s3.rs
yoogoc: Removed `oss:ListMultipartUploads` entirely. It was bucket-wide with no equivalent in the AWS policy, and an Iceberg client only needs the object-scoped `oss:ListParts`(still granted) to complete its own uploads.

crates/lakekeeper/src/service/storage/s3.rs
yoogoc: You're right, thanks. The credential vending now parses the STS-returned `Expiration` (via `DateTime::from_str`) instead of recomputing `now + DurationSeconds`, and I replaced the test with `test_aliyun_assume_role_response_carries_returned_expiration`, which deserializes a full response and asserts the returned expiration flows through to the value the vended credential carries.

yoogoc:
> Thank you for your PR @yoogoc, we appreciate the work and would love to support using Alibaba Cloud OSS.
> 
> I left a few review comments.
> 
> One thing that is missing from the comments I left in the code: if we have support for Alibaba OSS Cloud we should also mentions this in the documentation and f. ex. explain required and optional parameters. See f. ex. the documentation for R2 in `docs/docs/storage.md` lines 560-603. It would be really great if you could a little bit of documentation there.

Done — added an "Alibaba Cloud OSS" section to docs/docs/storage.md (right after Cloudflare R2), following the same structure: setup steps, a full warehouse-creation example, the auto-set params (flavor=s3-compat, sts-enabled=true), and the required/optional parameters (endpoint and sts-role-arn/assume-role-arn required; sts-endpoint and external-id optional).

brgr-s (dismissed):
Your changes LGTM! :) Thanks a lot @yoogoc !

I did some local tests agains OSS and found a few things that I implemented directly on your PR. Hope you don't mind

- OSS does not support `path-style-access`, so we check for that in normalize
- OSS does not support chunked encoding, so I had to disable the default behavior, which in turn...
- means we have to use Legacy MD5 CRC for certain operations that do require a checksum

I added a few notes in the docs specifically for clients; they need to make sure that chunked encoding ist not used, otherwise multipart uploads will always fail. They also need to use the specific s3-style endpoint.

I also added some integration tests against a bucket we provide for our integration tests.

yoogoc:
> Your changes LGTM! :) Thanks a lot @yoogoc !
> 
> I did some local tests agains OSS and found a few things that I implemented directly on your PR. Hope you don't mind
> 
> * OSS does not support `path-style-access`, so we check for that in normalize
> * OSS does not support chunked encoding, so I had to disable the default behavior, which in turn...
> * means we have to use Legacy MD5 CRC for certain operations that do require a checksum
> 
> I added a few notes in the docs specifically for clients; they need to make sure that chunked encoding ist not used, otherwise multipart uploads will always fail. They also need to use the specific s3-style endpoint.
> 
> I also added some integration tests against a bucket we provide for our integration tests.

very thanks!!!

brgr-s:
@yoogoc I pushed another small change: we decided to mark OSS support as beta. This has nothing to do with your additions, we are just not sure how well OSS plays with the actual clients (Spark, pyiceberg, Trino, ...). When we receive some feedback that everything looks good we will remove the "beta flag". I will merge your changes today. Thanks again!

PR #1910, opened 2026-07-24
feat(storage): enable S3 request signing for Generic Tables

### LINKED ISSUE #1908: Support S3 remote signing for Generic Tables
**Is your feature request related to a problem? Please describe.**

Loading a generic table returns a `config` that advertises remote signing
(`s3.remote-signing-enabled: "true"` plus table-scoped `signer.uri` /
`signer.endpoint` properties), but sign requests for URIs under the table's
own `base-location` are rejected with `NoSuchTableLocationException` — by
both the advertised table-scoped signer endpoint and the warehouse-scoped
one.

This makes generic tables unusable on S3-compatible stores that have no STS
(e.g. Scaleway Object Storage): credential vending can't work there, and remote signing refuses to sign.

**Describe the solution you'd like**
Get a signed request under the Generic Table location

**Describe alternatives you've considered**
Excluding Generic Tables from remote signing would at least mean that the load response should not return 's3.remote-signing-enabled': 'true' and the signer.uri and signer.endpoint. But this is not preferred as we would then have to create long lived S3 credentials that would undermine the catalog managing all access to the object store.

**Additional context**

## Environment

- Lakekeeper `0.13.1` (`quay.io/lakekeeper/catalog:latest-main`, digest `sha256:5a7c2fc4…`)
- Storage: MinIO (S3-compatible, `flavor: s3-compat`, `sts-enabled: false`)
- Auth disabled (allow-all); also reproduces with authn/authz enabled

## Reproduction

`docker-compose.yaml` (trimmed from `examples/access-control-simple`):

```yaml
services:
  lakekeeper:
    image: quay.io/lakekeeper/catalog:latest-main
    environment:
      - LAKEKEEPER__PG_ENCRYPTION_KEY=This-is-NOT-Secure!
      - LAKEKEEPER__PG_DATABASE_URL_READ=postgresql://postgres:postgres@db:5432/postgres
      - LAKEKEEPER__PG_DATABASE_URL_WRITE=postgresql://postgres:postgres@db:5432/postgres
    command: ["serve"]
    healthcheck:
      test: ["CMD", "/home/nonroot/lakekeeper", "healthcheck"]
      interval: 1s
      timeout: 10s
      retries: 10
    depends_on:
      migrate:
        condition: service_completed_successfully
      minio-init:
        condition: service_completed_successfully
    ports:
      - "8181:8181"

  migrate:
    image: quay.io/lakekeeper/catalog:latest-main
    environment:
      - LAKEKEEPER__PG_ENCRYPTION_KEY=This-is-NOT-Secure!
      - LAKEKEEPER__PG_DATABASE_URL_READ=postgresql://postgres:postgres@db:5432/postgres
      - LAKEKEEPER__PG_DATABASE_URL_WRITE=postgresql://postgres:postgres@db:5432/postgres
    restart: "no"
    entrypoint: ["/home/nonroot/lakekeeper", "migrate"]
    depends_on:
      db:
        condition: service_healthy

  db:
    image: postgres:17
    environment:
      - POSTGRES_PASSWORD=postgres
    healthcheck:
      test: ["CMD-SHELL", "pg_isready -U postgres -p 5432 -d postgres"]
      interval: 2s
      timeout: 10s
      retries: 5

  minio:
    image: minio/minio:latest
    command: ["server", "/data"]
    environment:
      - MINIO_ROOT_USER=minioadmin
      - MINIO_ROOT_PASSWORD=minioadmin
      - MINIO_REGION=local
    healthcheck:
      test: ["CMD", "curl", "-f", "http://localhost:9000/minio/health/live"]
      interval: 2s
      timeout: 10s
      retries: 5
    ports:
      - "9000:9000"

  minio-init:
    image: minio/mc:latest
    restart: "no"
    entrypoint:
      - /bin/sh
      - -c
      - mc alias set local http://minio:9000 minioadmin minioadmin && mc mb --ignore-existing local/lakehouse
    depends_on:
      minio:
        condition: service_healthy
```

`repro.py` (only needs `requests`):

```python
import requests

LAKEKEEPER = "http://localhost:8181"
S3_ENDPOINT = "http://minio:9000"

# bootstrap + warehouse with STS disabled (no vending, like any non-AWS S3 store)
requests.post(f"{LAKEKEEPER}/management/v1/bootstrap", json={"accept-terms-of-use": True})
requests.post(f"{LAKEKEEPER}/management/v1/warehouse", json={
    "warehouse-name": "local-lakehouse",
    "storage-profile": {
        "type": "s3", "bucket": "lakehouse", "region": "local",
        "endpoint": S3_ENDPOINT, "path-style-access": True,
        "flavor": "s3-compat", "sts-enabled": False,
    },
    "storage-credential": {
        "type": "s3", "credential-type": "access-key",
        "aws-access-key-id": "minioadmin", "aws-secret-access-key": "minioadmin",
    },
})

prefix = requests.get(f"{LAKEKEEPER}/catalog/v1/config",
                      params={"warehouse": "local-lakehouse"}).json()["defaults"]["prefix"]
requests.post(f"{LAKEKEEPER}/catalog/v1/{prefix}/namespaces", json={"namespace": ["repro"]})

# create + load a generic table
requests.post(f"{LAKEKEEPER}/lakekeeper/v1/{prefix}/namespaces/repro/generic-tables",
              json={"name": "blobs", "format": "dataset"})
t = requests.get(f"{LAKEKEEPER}/lakekeeper/v1/{prefix}/namespaces/repro/generic-tables/blobs").json()
cfg, loc = t["config"], t["table"]["base-location"]
print("advertised:", {k: cfg.get(k) for k in
                      ("s3.remote-signing-enabled", "signer.uri", "signer.endpoint")})

# sign a PUT under the table's own base-location
body = {
    "method": "PUT", "region": "local",
    "uri": loc.replace("s3://", S3_ENDPOINT + "/") + "/data/file.bin",
    "headers": {"x-amz-content-sha256": ["UNSIGNED-PAYLOAD"]},
}
r = requests.post(cfg["signer.uri"].rstrip("/") + "/" + cfg["signer.endpoint"], json=body)
print("table-scoped signer:    ", r.status_code, r.json()["error"]["type"])
r = requests.post(f"{LAKEKEEPER}/catalog/v1/{prefix}/v1/aws/s3/sign", json=body)
print("warehouse-scoped signer:", r.status_code, r.json()["error"]["type"])
```

```
docker compose up -d --wait && python repro.py
```

## Actual output

```
advertised: {'s3.remote-signing-enabled': 'true',
             'signer.uri': 'http://localhost:8181/catalog/',
             'signer.endpoint': 'v1/signer/bed7d7de-…/tabular-id/019f8f77-…/v1/aws/s3/sign'}
table-scoped signer:     400 NoSuchTableLocationException
warehouse-scoped signer: 400 NoSuchTableLocationException
```

Full error message:

```
Table does not exist or user does not have permission to view it at location
`s3://lakehouse/019f8f77-…/data/file.bin` in warehouse `bed7d7de-…`
```

## Root cause (from reading main) (investigated using Claude)

The signer resolves the tabular at the requested location, but drops
everything that isn't an Iceberg table:

- `crates/lakekeeper/src/server/s3_signer/sign.rs` — the table-id path uses
  `C::get_table_info(...)`, which doesn't resolve generic-table ids, so it
  falls back to location-based lookup via `get_table_info_by_location`.
- The location lookup (`crates/lakekeeper-storage-postgres/src/tabular/load_by_location.rs`)
  *does* find the generic table (its SQL handles `typ = 'generic-table'`),
  but `sign.rs::get_table_info_by_location` then calls
  `GenericTabularInfo::into_table_info()`, which maps
  `GenericTable(_) => None`
  (`crates/lakekeeper/src/service/catalog_store/tabular.rs`), and `None` is
  treated as "no table at this location" → `NoSuchTableLocationException`.
  (The adjacent `tracing::warn!` only mentions views.)

So the pieces are nearly all there: the location index already resolves
generic tables and the load response already emits the signer properties —
only the conversion in the sign path discards them. Presumably signing for
generic tables would also want its own authz action (analogous to
`get_data_access` on Iceberg tables).

## Use case / motivation

We ingest opaque files (NOAA GRIB2 weather data) into generic tables from
Airflow. Our object store is Scaleway (S3-compatible, **no STS**), so vended
credentials are not available; `sts-enabled` is false on the warehouse. Remote signing is the
mechanism that covers this exact situation and works well for our Iceberg
tables; extending it to generic tables would make them usable on any
S3-compatible store without static credentials on the client.

Happy to contribute a PR if the approach above (map generic tables through in
the signer's location lookup + appropriate authz action) sounds right

N-Clerkx:
Fixes lakekeeper/lakekeeper#1908

Problem

Generic tables advertise remote signing in their load response (s3.remote-signing-enabled, signer.uri, signer.endpoint), but the S3 sign endpoint rejected every request for a generic table's own data with 400 NoSuchTableLocationException. On STS-disabled S3-compatible storage remote signing is the only credential mechanism, so generic tables were effectively unusable there.

Cause

The signer resolves the tabular by request location, then discarded anything that wasn't an Iceberg table — so generic tables (and their locations) were dropped before authorization. This affected both the warehouse-scoped (/v1/aws/s3/sign) and table-scoped (/signer/{prefix}/tabular-id/{id}/…) entrypoints.

Change

sign() now keeps generic tables during location lookup and branches on the resolved tabular type:

* Iceberg tables keep the existing authorization path unchanged.
* Generic tables authorize via require_generic_table_action using the existing CanReadData/CanWriteData relations, audited under a generic-table event context.

Both branches rejoin the shared signing tail. Views remain non-signable, and requests for locations outside any tabular still return NoSuchTableLocationException.

No authorization-model change: the generic-table read/write relations already exist.

Testing

New integration test server_generic_table_signing (STS-disabled S3-compat warehouse) covering warehouse-scoped signind → location fallback), and rejection of foreign locations.

Disclaimer: I used AI to draft this PR, rust is not my strong suit. So happy to spend some more time on it if another implementation is wanted.

## Summary by CodeRabbit

* **New Features**
  * Added S3 remote signing support for generic tables.
  * Added warehouse- and table-scoped signing for generic table locations.
  * Generic table signing now applies the appropriate authorization checks.
* **Bug Fixes**
  * Invalid or unrelated S3 locations are rejected with a clear error.
  * Improved handling of generic table locations during request signing.
* **Tests**
  * Added integration coverage for successful signing and invalid-location scenarios.

## Summary by CodeRabbit

* **New Features**
  * Added remote S3 signing support for generic tables when using S3-compatible storage without STS.
  * Enforced permissions based on read, write, and bulk-delete operations.
  * Restricted signing to requests within a table’s base location; views remain unsupported.
* **Documentation**
  * Documented configuration, permissions, client behavior, and storage requirements for generic-table remote signing.
  * Clarified remote-signing support in the storage documentation.
* **Bug Fixes**
  * Improved tabular resource resolution and authorization across generic and Iceberg tables.

## Summary by CodeRabbit

- **New Features**
  - Added remote signing support for generic tables on S3-compatible storage without STS.
  - Generic-table requests now receive table-specific signed access configuration.
  - Improved support for resolving tabular resources by identifier or storage location.

- **Bug Fixes**
  - Corrected authorization for generic-table read, write, and delete operations.
  - Improved handling of mismatched locations and unsupported or non-signable resources.

- **Documentation**
  - Added guidance for remote signing configuration, permissions, supported requests, and storage limitations.


crates/lakekeeper-integration-tests/tests/server_generic_table_signing.rs
```diff
@@ -0,0 +1,246 @@
...
+    // `remote_signing_enabled` defaults to true in the builder.
+    profile.normalize(Some(&cred)).unwrap();
+    (profile, cred)
+}
+
+async fn setup(pool: PgPool) -> (Ctx, NamespaceIdent, WarehouseId) {
+    lakekeeper_storage_postgres::migrations::migrate_core_only(&pool)
+        .await
+        .unwrap();
+    let ctx = get_api_context(&pool, AllowAllAuthorizer::default()).await;
```
fresh-borzoni: All three tests use AllowAllAuthorizer, so the new generic-table authz branch is only ever tested in the "allow" direction. 
Could we add one test with a restrictive authorizer that asserts a caller without the generic-table read/write relation is rejected?
Also what about 'delete' test?

crates/lakekeeper/src/server/s3_signer/sign.rs
fresh-borzoni: The signer URL is deliberately type-agnostic (tabular-id/{uuid}), but resolution here assumes type = 'table', so generic-table ids always miss and only work by falling through the pyiceberg location fallback. 

The comment above says is for removal. Since this is the only route that takes a bare tabular id (everywhere else the type is in the path), would it be worth resolving the id by its actual type here so generic tables are handled deliberately and don't break when that fallback goes?

fresh-borzoni (commented):
@N-Clerkx Thanks for the PR, LGTM overall, couple of comments

Also shall we change docs to provide guidance how to use signing for generic tables?

crates/lakekeeper-integration-tests/tests/server_generic_table_signing.rs
brgr-s: @N-Clerkx I agree with the comment from @fresh-borzoni. 

A `DELETE` test next to the `PUT` and `GET` test is a good idea.

We also have a `HidingAuthorizer`, which could be used to have test cases where we could block `WriteData` and check if `PUT`/`DELETE` is rejected, and block `ReadData` to check if `GET` is rejected.

crates/lakekeeper/src/server/s3_signer/sign.rs
brgr-s: @fresh-borzoni @N-Clerkx I agree that this would be a clean solution, but it is a bit more involved. A possible solution could be a default method on `CatalogStore`: `get_tabular_info_by_uuid` that "probes" the existing `get_tabular_infos_by_id` with `Table`, `View` and `GenericTable`. As a result (on success, if id exists) we should get a `ViewOrTableInfo` that also tells us the type and gives us the matching `table_info` (`TableInfo`, `ViewInfo`, `GenericTableInfo`). We can fail early if it is a `View`

With this, we could also restructure the code a bit. Currently the `match` on `table_info` tries to preserve the old structure of the code. Assume we had a small helper `resolve_signable_by_id` that takes the `Uuid` from the request, uses above mentioned `get_tabular_info_by_uuid` and returns a `Result<Option<SignableTabular>, ...` we could do something like this:

```
enum ResolvedSignable {
    Generic(GenericTabularInfo),
    Iceberg(Result<Option<TableInfo>, RequireTableActionError>),
}
```

```
let resolved = match table_info {
    Ok(Some(SignableTabular::GenericTable(gt))) => ResolvedSignable::Generic(gt),
    Ok(Some(SignableTabular::Table(t))) => ResolvedSignable::Iceberg(Ok(Some(t))),
    Ok(None) => ResolvedSignable::Iceberg(Ok(None)), // the pyiceberg <= 0.9.1 case
    Err(e) => ResolvedSignable::Iceberg(Err(e)),
};

let (location, table_id) = match resolved {
    ResolvedSignable::Generic(gt_info) => { /* generic authz + event_ctx */ }
    ResolvedSignable::Iceberg(info)    => { /* existing iceberg authz + event_ctx, including pyiceberg <= 0.9.1 case */ }
};
```

A small `location` accessor on `ViewOrTableInfo` might be helpfull in the case where we need the `Location`.

brgr-s:
@N-Clerkx Thank you for your PR and taking the time to solve this even though Rust is not your "primary language". Sorry for taking a bit long to have a look at it.

I agree with @fresh-borzoni comments.

I posted a small idea on how we could clean everything up and make it a bit more "ergonomic". If you are interested in taking a swing at that, we are happy to help! :)

An addition to the docs regarding signing for generic tables would also be highly appreciated.

crates/lakekeeper-integration-tests/tests/server_generic_table_signing.rs
N-Clerkx: Added a `DELETE` and `PUT` test and also added tests to check the authz rejection path

crates/lakekeeper/src/server/s3_signer/sign.rs
N-Clerkx: Implemented as sketched — `resolve_signable_by_id` returning `Result<Option<SignableTabular>, _>`, plus `ResolvedSignable` with your four arms. `ViewOrTableInfo::location()` already existed.

I didn't fail early on `View`, though: a view maps to `None` and falls through to the location lookup, ending in the existing `NoSuchTableLocationException`. Under the PyIceberg ≤ 0.9.1 bug the path id can be a stale id from an unrelated tabular — including a view — while the URI legitimately belongs to a table the caller may access, so failing on the id's type would break the clients the fallback exists for. Views are still rejected; `test_sign_view_is_not_signable` pins it. Happy to switch to a hard failure if you prefer.

N-Clerkx:
Thanks for the review!
I tried to address your comments, let me know what you think :)

brgr-s:
@N-Clerkx LGTM, I just have 2 small nits. I tried pushing to your repo because I thought it would be the easiest way, but I do not have push permissions to your branch. I attached a git patch that you can checkout.

[0001-A-few-small-nits.patch](https://github.com/user-attachments/files/30503219/0001-A-few-small-nits.patch)

brgr-s:
@N-Clerkx merged. Many thanks for your contribution and thank you @fresh-borzoni for the review and help.

PR #1924, opened 2026-08-05
fix(authz): emit authorization events before applying assignment writes

### LINKED ISSUE #1609: Audit: emit authorization events before OpenFGA writes in update_* assignment endpoints
Context

    Follow-up from PR #1607 (see linked comment below).
    Requester: @c-thiel.

Problem

    The update_* assignment endpoints in crates/authz-openfga/src/api.rs call the combined checked_write (authorization + OpenFGA writes) before emitting audit/authz events via event_ctx.emit_authz.
    If the process crashes between the write and the emit, the authorization attempt can be lost in audit logs.
    This ordering conflicts with our audit guideline to log the attempted action before committing side-effects.

Scope
Refactor the following endpoints in crates/authz-openfga/src/api.rs to perform an authorization decision first, emit the audit/authz event, and only then apply OpenFGA writes:

    update_server_assignments
    update_project_assignments
    update_project_assignments_by_id
    update_warehouse_assignments_by_id
    update_namespace_assignments_by_id
    update_table_assignments_by_id
    update_view_assignments_by_id
    update_role_assignments_by_id

Acceptance criteria

    For each listed endpoint, authorization events (authorization_succeeded / authorization_failed) are emitted before any OpenFGA write side-effects.
    If an OpenFGA write fails after the emit, the HTTP response still returns the appropriate write error, without double-logging the authorization outcome.
    Tests cover the ordering (e.g., simulate a failing write and still observe the audit attempt).
    No externally visible API contract changes.
    Documentation updates, if necessary, to reflect the ordering guarantee.

Backlinks

    PR: https://github.com/lakekeeper/lakekeeper/pull/1607
    Comment: https://github.com/lakekeeper/lakekeeper/pull/1607#discussion_r2807589185

Next steps

    If you’d like, I can retry creating it from the bot once issue-creation permissions are confirmed.
    Do you want labels like area:authz and type:refactor, and keep you as the assignee?

AndreaBozzo:
Closes #1609.

## What changed

The permission-assignment endpoints committed their OpenFGA writes *before* the authorization event reached the audit log. A crash in that window could apply a permission change that no audit record ever mentions — the case where an audit trail matters most.

These endpoints now decide authorization, emit the audit event, and only then apply the grant. If the grant fails after that point the caller still gets the write error, and the authorization outcome is logged exactly once — never twice.

Affected endpoints, all under `POST /management/v1/permissions/…/assignments`: server, project (default-project and by-id forms), warehouse, namespace, table, view, generic-table, role, and tag.

The issue lists eight endpoints; the generic-table and tag assignment updates were added after it was filed and carried the same ordering, so they are included here — leaving them behind would have reintroduced the bug on the newest endpoints.

No API contract changes. The two `set_*_managed_access` endpoints in the same router already had the correct ordering and are untouched.

## Audit-log semantics

An authorization event records the **attempt**. An `"allowed"` decision means the caller was permitted to perform the action, not that the action succeeded — if the operation fails afterwards, the request returns an error while the authorization event remains in the log. Consumers reconciling audit logs against actual state should correlate authorization events with the corresponding operational event. Documented in `docs/docs/logging.md`.

## Tests

Two regression tests in `crates/authz-openfga/src/api.rs`:

- **`test_endpoint_audits_authz_before_failing_write`** drives `update_warehouse_assignments_by_id` through its real extractors, with a capturing `EventListener` appended to the live dispatcher after setup. It calls the endpoint twice with an identical request: the second call's authorization still passes while OpenFGA rejects the duplicate tuple, so the write fails strictly *after* the decision. It asserts exactly two `authorization_succeeded` events — the attempt is audited despite the failed write, and is not logged twice.
- **`test_check_assignment_writes_has_no_side_effects`** pins the contract that the authorization phase performs no writes.

Both were verified to fail against the previous ordering, reverting each half of the change separately:

| | check re-fused with apply | endpoint writes before emit |
|---|---|---|
| `test_endpoint_audits_authz_before_failing_write` | fails | fails |
| `test_check_assignment_writes_has_no_side_effects` | fails | **passes** |

The contract test does not discriminate endpoint ordering — the endpoint test is what covers that. All 22 tests in the OpenFGA integration module pass, including the 20 pre-existing ones.

## Release notes

Authorization events for permission-assignment changes are now recorded in the audit log before the change is applied, so a crash mid-request can no longer apply a permission change that leaves no trace in the audit trail. An `allowed` decision in the audit log now denotes a permitted attempt rather than a completed change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Authorization audit events are now recorded before permitted assignment changes are applied.
  * Prevented duplicate authorization audit events when an assignment update fails, including duplicate-write errors.
  * Authorization checks no longer modify assignment state before the update is applied.
  * Failed assignment updates continue to return errors while preserving the authorization audit record.

* **Documentation**
  * Clarified that an allowed authorization event confirms permission for an attempted action, not successful completion.

### DIFF
--- a/crates/authz-openfga/src/api.rs
+++ b/crates/authz-openfga/src/api.rs
@@ -1972,15 +1972,19 @@ async fn update_server_assignments<C: CatalogStore, S: SecretStore>(
         request.clone(),
         lakekeeper::service::authz::Authorizer::server_id(&authorizer),
     );
-    let authz_result = checked_write(
-        authorizer,
+    let authz_result = check_assignment_writes(
+        &authorizer,
         event_ctx.request_metadata().actor(),
-        request.writes,
-        request.deletes,
+        &request.writes,
+        &request.deletes,
         &server_id,
     )
     .await;
     let _ = event_ctx.emit_authz(authz_result)?;
+
+    apply_assignment_writes(authorizer, request.writes, request.deletes, &server_id)
+        .await
+        .map_err(authz_to_error_no_audit)?;
 
     Ok(StatusCode::NO_CONTENT)
 }
@@ -2012,15 +2016,20 @@ async fn update_project_assignments<C: CatalogStore, S: SecretStore>(
         project_id,
         Arc::new(request.clone()),
     );
-    let authz_result = checked_write(
-        authorizer,
+    let object = event_ctx.user_provided_entity().to_openfga();
+    let authz_result = check_assignment_writes(
+        &authorizer,
         event_ctx.request_metadata().actor(),
-        request.writes,
-        request.deletes,
-        &event_ctx.user_provided_entity().to_openfga(),
+        &request.writes,
+        &request.deletes,
+        &object,
     )
     .await;
     let _ = event_ctx.emit_authz(authz_result)?;
+
+    apply_assignment_writes(authorizer, request.writes, request.deletes, &object)
+        .await
+        .map_err(authz_to_error_no_audit)?;
 
     Ok(StatusCode::NO_CONTENT)
 }
@@ -2052,15 +2061,20 @@ async fn update_project_assignments_by_id<C: CatalogStore, S: SecretStore>(
         project_id,
         request.clone(),
     );
-    let authz_result = checked_write(
-        authorizer,
+    let object = event_ctx.user_provided_entity().to_openfga();
+    let authz_result = check_assignment_writes(
+        &authorizer,
         event_ctx.request_metadata().actor(),
-        request.writes,
-        request.deletes,
-        &event_ctx.user_provided_entity().to_openfga(),
+        &request.writes,
+        &request.deletes,
+        &object,
     )
     .await;
     let _ = event_ctx.emit_authz(authz_result)?;
+
+    apply_assignment_writes(authorizer, request.writes, request.deletes, &object)
+        .await
+        .map_err(authz_to_error_no_audit)?;
 
     Ok(StatusCode::NO_CONTENT)
 }
@@ -2092,15 +2106,20 @@ async fn update_warehouse_assignments_by_id<C: CatalogStore, S: SecretStore>(
         warehouse_id,
         request.clone(),
     );
-    let authz_result = checked_write(
-        authorizer,
+    let object = event_ctx.user_provided_entity().to_openfga();
+    let authz_result = check_assignment_writes(
+        &authorizer,
         event_ctx.request_metadata().actor(),
-        request.writes,
-        request.deletes,
-        &event_ctx.user_provided_entity().to_openfga(),
+        &request.writes,
+        &request.deletes,
+        &object,
     )
     .await;
     let _ = event_ctx.emit_authz(authz_result)?;
+
+    apply_assignment_writes(authorizer, request.writes, request.deletes, &object)
+        .await
+        .map_err(authz_to_error_no_audit)?;
 
     Ok(StatusCode::NO_CONTENT)
 }
@@ -2132,15 +2151,20 @@ async fn update_namespace_assignments_by_id<C: CatalogStore, S: SecretStore>(
         namespace_id,
         request.clone(),
     );
-    let authz_result = checked_write(
-        authorizer,
+    let object = namespace_id.to_openfga();
+    let authz_result = check_assignment_writes(
+        &authorizer,
         event_ctx.request_metadata().actor(),
-        request.writes,
-        request.deletes,
-        &namespace_id.to_openfga(),
+        &request.writes,
+        &request.deletes,
+        &object,
     )
     .await;
     let _ = event_ctx.emit_authz(authz_result)?;
+
+    apply_assignment_writes(authorizer, request.writes, request.deletes, &object)
+        .await
+        .map_err(authz_to_error_no_audit)?;
 
     Ok(StatusCode::NO_CONTENT)
 }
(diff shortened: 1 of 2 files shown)


docs/docs/logging.md
```diff
@@ -70,6 +70,8 @@ Emitted for every authz check. Always contain `action`/`actions`, `entity`/`enti
 
 **Note:** Empty arrays and objects are omitted from the output. For example, if `stack` is empty, the field will not appear in the log.
 
+**Ordering:** An authorization event records the *attempt*, and is emitted before the authorized operation commits its side-effects. An `"allowed"` decision therefore means the caller was permitted to perform the action, not that the action succeeded — if the operation fails afterwards the request returns an error while the authorization event remains in the log. This ordering is deliberate: a crash between the decision and the side-effect leaves the attempt audited rather than silently lost. Consumers reconciling audit logs against actual state should treat authorization events as attempts and correlate them with the corresponding operational event (see [Operational Audit Events](#operational-audit-events)) to confirm completion.
```
c-thiel: Small accuracy nit on this great addition: `emit_authz` dispatches the event via `tokio::spawn` (fire-and-forget), so the event is initiated before the write rather than durably persisted before it — a crash right after the spawn could still drop it. Could we soften "a crash between the decision and the side-effect leaves the attempt audited rather than silently lost" to something like "the attempt is dispatched to the audit log before the write is issued"? Also, the "correlate with the corresponding operational event" pointer doesn't quite hold for assignment endpoints — they don't emit an operational event, so there's nothing to reconcile against here (unlike role/tag ops). Maybe drop that clause for this case.

crates/authz-openfga/src/api.rs
```diff
@@ -2680,11 +2730,16 @@ async fn get_allowed_actions<A: ReducedRelation + IntoEnumIterator>(
     Ok(actions)
 }
 
-async fn checked_write<RA: Assignment>(
-    authorizer: OpenFGAAuthorizer,
+/// Authorize an assignment update without applying it.
+///
+/// Split out from the write so that callers can emit the authorization audit
+/// event *before* any OpenFGA side-effects are committed. Apply the writes with
```
c-thiel: Same nuance as the docs: since emit_authz spawns the dispatch, this is "before the write is issued" rather than "before any side-effects are committed." Tiny wording tweak to keep the doc comment honest.

crates/authz-openfga/src/api.rs
c-thiel: Now that the fn takes `&OpenFGAAuthorizer` and `check` is `&self`, the `.clone()` here is no longer needed I think and `authorizer.check(key)` should work.

crates/authz-openfga/src/api.rs
```diff
@@ -2734,7 +2789,20 @@ async fn checked_write<RA: Assignment>(
     }))
     .await?;
 
-    // ---------------------- APPLY WRITE OPERATIONS -----------------------
+    Ok(())
+}
+
+/// Apply an assignment update that [`check_assignment_writes`] has authorized.
+///
+/// Callers must have emitted the authorization event before calling this, and
```
c-thiel: Nit: "callers must map failures with authz_to_error_no_audit" is true for the endpoint callers, but the in-crate test callers don't (and can't, since they return OpenFGAResult). "endpoint callers" would be exact.

c-thiel (dismissed):
Thanks for this nice PR @AndreaBozzo! This cleanly closes the ordering gap from #1609/#1607.

Left a few small wording/cleanup nits inline; none are blockers.

Separately — while reviewing I noticed `role.rs` and `tag.rs` have the same "emit after the mutation commits" shape: their `authorize_*` helpers do the DB + OpenFGA write and commit before `emit_authz` runs (e.g. `create_role` → `authorize_create_role` at role.rs:371-373, `set_warehouse_tag` at tag.rs:765-781), so they carry the same crash-window + write-failure-mislabel behavior this PR fixes for assignments. It's ~19 handlers and more involved (the check and write are interleaved inside one helper), so definitely out of scope here. Would you be willing to take that on as a follow-up PR, mirroring the `delete_project` shape (pure `require_*_action` → `emit_authz` → business logic)? Happy to file an issue with the full handler list if that helps.

AndreaBozzo:
> Thanks for this nice PR @AndreaBozzo! This cleanly closes the ordering gap from #1609/#1607.
> 
> Left a few small wording/cleanup nits inline; none are blockers.
> 
> Separately — while reviewing I noticed `role.rs` and `tag.rs` have the same "emit after the mutation commits" shape: their `authorize_*` helpers do the DB + OpenFGA write and commit before `emit_authz` runs (e.g. `create_role` → `authorize_create_role` at role.rs:371-373, `set_warehouse_tag` at tag.rs:765-781), so they carry the same crash-window + write-failure-mislabel behavior this PR fixes for assignments. It's ~19 handlers and more involved (the check and write are interleaved inside one helper), so definitely out of scope here. Would you be willing to take that on as a follow-up PR, mirroring the `delete_project` shape (pure `require_*_action` → `emit_authz` → business logic)? Happy to file an issue with the full handler list if that helps.

Hi @c-thiel , i addressed the nits in the latest commit

I would appreciate alot if you could open the ticket about the follow-up and assign it to me so i can work on the changes in the next few days.

Take care

c-thiel:
Thanks @AndreaBozzo!
Follow-up issue is here: https://github.com/lakekeeper/lakekeeper/issues/1933
I can assing you after you comment into the issue :)

PR #1926, opened 2026-08-05
feat(postgres): place Lakekeeper tables in a configurable schema

vyruss:
Lakekeeper's tables land in whichever schema Postgres resolves through `search_path`, and the only supported way to move them was to set the default on the role with `ALTER ROLE`. That is a manual step performed directly on the database, outside the environment variables the rest of the deployment is configured with, and it has to be repeated for every role and every new database.

This PR makes the schema a setting. With `LAKEKEEPER__PG_SCHEMA` set, an `after_connect` hook applies `set_config('search_path', $1, false)` to every connection, added in `to_pool_opts()` so all pools inherit it rather than each call site remembering. The name is validated as a bare identifier and travels as a bind parameter, so no configured value reaches SQL text. 

`migrate` creates the schema under the advisory lock it already holds, then aborts unless `current_schema()` matches, so a `search_path` that never arrived cannot quietly populate `public`. It probes `pg_namespace` before issuing `CREATE SCHEMA`, because `CREATE SCHEMA IF NOT EXISTS` checks `CREATE` on the database before its `IF NOT EXISTS` short-circuit and will fail for a low-privilege role even if the schema is already there. `check_migration_status` reports a missing schema as `NoMigrationsTable` but errors on a mismatch, so an existing install in `public` cannot read as `Complete` and start the server against the wrong data.

`public` stays on the `search_path` so unqualified extension objects still resolve, including `uuid_generate_v1mc()` in column defaults, `pgp_sym_encrypt` and `pgp_sym_decrypt` in the secret backend, and the `gin_trgm_ops` operator class. The configured schema comes first, so nothing in public can shadow a Lakekeeper object.

Unset, no hook is installed and nothing changes. Migration SQL, the .sqlx cache and every public signature are untouched, so callers outside the postgres crate are unaffected and no migration checksum moves.

`ALTER ROLE` remains supported and is still the answer behind a pooler in transaction mode, which does not keep a session-level `SET`. Running several Lakekeepers in one database still requires creating the Postgres extensions in `public` beforehand, otherwise the first migration installs them inside its own schema where the others cannot reach them; that is equally true of the `ALTER ROLE` approach today.

14 tests (against a live Postgres):

1. Name validation and configuration parsing (4)
2. Hook applying per connection, and not applying when unset (2)
3. Migrate: schema creation, low-privilege role with a pre-created schema and no `ALTER ROLE`, idempotency, abort when `search_path` was stripped (4)
4. Two schemas in one database: coexistence and concurrent migration (2)
5. Migration status: missing schema and mismatched schema (2)

## Summary by CodeRabbit

* **New Features**
  * Added support for configuring a dedicated PostgreSQL schema.
  * Connections automatically use the configured schema.
  * Migrations can create, validate, and track migration data within the selected schema.
  * Added validation to prevent invalid or unsafe schema names.

* **Bug Fixes**
  * Startup now stops with a clear error when the configured schema cannot be resolved correctly.

* **Documentation**
  * Added guidance for custom schemas, permissions, connection pooling, extensions, migrations, and development environment settings.


crates/lakekeeper-storage-postgres/src/migrations/mod.rs
c-thiel, line 1: I think we should also modify `crates/lakekeeper-bin/src/wait_for_db.rs:67 — the unready => arm`:

Since this match binds the whole `Result`, a schema-mismatch `Err` from `verify_current_schema` (pooler dropped the session SET, or the role is missing USAGE) lands here and gets retried for the full budget, then bails with the generic "run migrate" message. That's a permanent misconfig retrying can't fix — same shape as `Ahead` two arms up, which we fail fast on. I think we should match the `Err` (or add a `SchemaMismatch` state) and bail immediately, surfacing the anyhow message at error. The good guidance (grant USAGE / session pooling / ALTER ROLE) is already in that error but currently only shows up as an info-level.

docs/docs/configuration.md
```diff
@@ -89,14 +90,23 @@ CREATE EXTENSION IF NOT EXISTS "btree_gist";
...
+Set `LAKEKEEPER__PG_SCHEMA=lakekeeper` to keep Lakekeeper's tables in a dedicated schema instead of `public`. Lakekeeper sets `search_path` to `"lakekeeper", public` on every connection, and `lakekeeper migrate` creates the schema when the role has `CREATE` on the database. Otherwise create it first: `CREATE SCHEMA lakekeeper AUTHORIZATION lakekeeper;`
+
+`public` stays on the path so extension functions and operator classes installed there keep resolving. The schema name is quoted, so it is case-sensitive, and both the read and the write role need `USAGE` on it.
+
+If you run more than one Lakekeeper in the same database, create the [required extensions](#required-postgres-extensions) in `public` first. Otherwise the first migration installs them inside its own schema, where the other deployments cannot reach them.
+
+Pointing an existing deployment at a new schema shows an empty catalog: the data stays where it is, so move it yourself with `ALTER TABLE ... SET SCHEMA` or a dump and restore.
```
c-thiel, line 99: On a fresh privileged install the CREATE EXTENSION statements have no WITH SCHEMA public, so with search_path = "lakekeeper", public the relocatable extensions actually land in lakekeeper, not public. That's fine while PG_SCHEMA is set — but if someone later unsets it (or moves tables back to public and drops the schema), pgp_sym_encrypt/pgp_sym_decrypt in the secrets backend would stop resolving. This paragraph covers moving tables — should we also mention the stranded extensions (recommend pre-creating them in public even for single deployments, or note ALTER EXTENSION … SET SCHEMA public as the rollback step)?

c-thiel (commented):
Thanks for this PR!
Is there a reason from your side this is still in draft?

vyruss:
> Thanks for this PR! Is there a reason from your side this is still in draft?

Hi @c-thiel, it was just my first PR here, and didn't want to clutter the review pipeline in case I had got something really obvious wrong. I've switched it to "ready for review".

crates/lakekeeper-storage-postgres/src/migrations/mod.rs
vyruss: OK, let's go for a `MigrationState::SchemaMismatch` instead of using the general `Err:`, because those should be legitimately retried. Putting it as an arm next to `Ahead` in `wait_for_db` that makes it fail hard. Changing the enum will affect anything downstream matching it exhaustively though (don't know that there is anything).

docs/docs/configuration.md
vyruss: Sure, let's do both: creating in `public` before the 1st migration even for single deployments, and suggesting `ALTER EXTENSION ... SET SCHEMA public` as the rollback step.

PR #1928, opened 2026-08-06
feat: support AWS partitions in S3 policies

### LINKED ISSUE #1927: Vended-Credentials AssumeRole alternate AWS Partition does not work
**Describe the bug**
Lakekeeper is advertised to work with AWS S3 with vended-credentials. Unfortunately, it will only work right now with AWS Commercial cloud. If someone uses a different AWS partition, the vended credentials attaches a STS inline policy for AWS commercial partition hardcoded which when gives permission denied when attempt to access the S3 bucket for a different partition. I think the AWS partition should be an optional parameter which defaults to AWS commercial but gives the ability for people to override where needed.

See AWS IAM Partition Docs: https://docs.aws.amazon.com/IAM/latest/UserGuide/reference-arns.html#arns-syntax

Two example partitions they provide…
arn:aws-cn:s3:::{}
arn:aws-us-gov:s3:::{}

https://github.com/lakekeeper/lakekeeper/blob/main/crates/lakekeeper/src/service/storage/s3.rs#L1241

`rust
let bucket_arn = format!(
      "arn:aws:s3:::{}",
    table_location.bucket_name().trim_end_matches('/')
        );`

**To Reproduce**
Hard to reproduce unless you have access to one of these other AWS regions. However, I have access to one of these other regions and tried to setup vended credentials. I got access denied by AWS S3 service and dug through the cloud trail logs. The vended credentials assume role inline policy mismatch on the partition was the issue.

**Expected behavior**
Should be able to do vended-credentials with assume role with alternative AWS partitions.

**Screenshots**
N/A

**System Information:**
 - Deployment Type: Helm chart
 - Lakekeeper Version: main

**Additional context**
N/A

123digits:
Resolves #1927 

## Summary by CodeRabbit

* **New Features**
  * S3 AWS partition selection is now determined automatically from the region, endpoint, or role ARN.
  * Added support for commercial, China, GovCloud, ISO, and European Sovereign Cloud regions.
  * Non-AWS storage flavors consistently use the standard AWS partition.

* **Bug Fixes**
  * Improved STS policy generation and role-assumption handling across AWS partitions.
  * Added restrictions and clearer errors for unsupported storage namespace moves.

* **Documentation**
  * Updated S3 configuration guidance and examples to reflect automatic partition selection.


123digits:
@c-thiel Do you have a chance to review this and the other two connecting MRs across REST API, Console, and Components?

crates/lakekeeper/src/service/storage/s3.rs
c-thiel, line 392: partition isn't handled here, so a full-replace update that omits it deserializes to the "aws" default and overwrites a configured aws-us-gov/aws-cn — silently breaking vended-credential ARNs on an unrelated update. Since partition is a function of the (immutable) region, carry the old value forward when region is unchanged, e.g. after the region check add: `if self.region == other.region { other.partition = self.partition; }`

c-thiel, line 340: Every other free-text field goes through a `validate_*/normalize_*` step, but partition flows unvalidated into `format!("arn:{}:s3:::{}", ...)` at L1245. An empty or malformed value yields a broken ARN that only fails opaquely at STS time. Add a `validate_partition` (allowlist aws/aws-cn/aws-us-gov/aws-iso* or `^[a-z0-9-]+$`, reject empty) and ideally check consistency with region.

c-thiel (commented):
Thanks for the PR @123digits. I left a two comments. Additionally, we need the following:
- Run `just update-management-openapi` and commit. Otherwise CI will fail. This PR updates a public API.
- Missing user docs — docs/docs/storage.md: The S3 config table has no partition row and no GovCloud/China mention. Prior field additions (e.g. sts-endpoint in #1653) added a doc row; please do the same here.
-  Test the serde-default path — L3469: The new test covers aws-us-gov; the "aws" default is only exercised incidentally. test_s3_profile_de_from_v1 in mod.rs already deserializes a pre-partition profile and asserts "aws", so this is minor — a one-line dedicated assertion would make the default explicit.

c-thiel:
@123digits did you see my comments from last week? Do you find some time to fix them?

123digits:
@c-thiel I just tried to update them. Let me know if I missed anything.

c-thiel:
Hi @123digits — thanks a lot for this, and sorry for the slow turnaround.

Your diagnosis was exactly right: the hardcoded `arn:aws:` in the STS session policy is the bug, and `get_sts_policy_string` is precisely the place to fix it. I hope you don't mind that I pushed a rework directly to your branch rather than sending you round another review cycle — the fix is the same, but I changed the configuration surface, so I'd like your eyes on it.

**What changed:** instead of an optional `partition` field, Lakekeeper now derives the partition.

- A profile with no `endpoint` is one where the AWS SDK resolves the endpoint from the region, so there a `us-gov-*`, `cn-*`, `us-iso*`, `eu-isoe-*` or `eusc-de-*` region determines the partition. The prefixes come from AWS's published `partitions.json`.
- Every other profile takes the partition from `sts-role-arn` / `assume-role-arn`, which `normalize` already requires for `flavor: aws` with STS enabled. So there is always something to go on, and a partition AWS adds after a given release still works.
- Only ARNs that name an AWS partition count, and a non-`aws` flavor always stays commercial — otherwise a MinIO deployment using `arn:minio:iam:::role/<id>` would end up with `arn:minio:s3:::bucket` in its policy, which MinIO matches against nothing.

**Why derive instead of configure.** Storage profiles are persisted and part of the public API, so a field is effectively permanent — and it would have to be plumbed through the console, the Terraform provider and the client SDKs, with every GovCloud and China user needing to learn the concept to get a working warehouse. Your second commit is the clearest argument for it: because storage-profile updates are a full replace and the field had a serde default, an update that omitted `partition` silently reset a GovCloud profile back to `aws`, so you had to add a carry-forward to compensate. Deriving the value removes that whole class of problem.

The asymmetry also matters: adding an optional override later is additive and painless, whereas removing one isn't. So if a case turns up where the derivation is wrong, we can add it then.

**Would you take a look?** Especially: does this cover your setup as you'd configure it, and is anything about the derivation surprising from an operator's point of view? And if you have access to a GovCloud environment, a real end-to-end check would be very welcome — our CI only reaches the commercial partition, so that part is unit-tested only.

Thanks again for finding and diagnosing this properly;

crates/lakekeeper/src/service/storage/s3.rs
```diff
@@ -1724,6 +1765,64 @@ fn validate_region(region: &str) -> Result<(), InvalidProfileError> {
...
+/// If AWS adds a partition, add its prefix here.
+fn partition_from_region(region: &str) -> Option<&'static str> {
+    const PARTITION_PREFIXES: &[(&str, &str)] = &[
+        ("cn-", "aws-cn"),
+        ("us-gov-", "aws-us-gov"),
+        ("us-iso-", "aws-iso"),
+        ("us-isob-", "aws-iso-b"),
+        ("us-isof-", "aws-iso-f"),
+        ("eu-isoe-", "aws-iso-e"),
+        ("eusc-de-", "aws-eusc"),
```
123digits, line 1835: @c-thiel Interesting alternative to derive it. This refactor looks fine to me. I see the other needed partitions in the list. Only thing I will point out is it would be nice if there was an option in the AWS Rust SDK to convert region-to-partition. They have the mapping but doesn't appear to be exposed... https://github.com/awslabs/aws-sdk-rust/blob/main/aws-models/sdk-partitions.json ... How soon do you think it might be for a release to try this?

Your welcome to close my other merges or I can close them if we aren't adding an option to be added by UI anymore.

c-thiel: I copied that list exactly from the rust sdk. Quite unfortunate that this isn't exposed from the crate :/
Feel free to close your other PRs then.

As this is now a change that does not modify state - its basically a fix - we could even backport it to 0.13 if you need it urgently. Otherwise I would release it as part of 0.14 probably sometime end of September.

123digits: @c-thiel Back port to 0.13 would be great to give it a try.

The merge button wasn’t available because I had to click the update button and now pipelines need a maintainer approval to run again to check for merging.

crates/lakekeeper/src/service/storage/s3.rs
123digits: I think you resolved this.

123digits: I think you resolved this.

crates/lakekeeper/src/service/storage/s3.rs
123digits: @c-thiel Did you see this yet?

c-thiel: Sorry, I was busy with other topics-  back to this now.
I'll merge and mark it for 0.13 backport too.

c-thiel (approved):

PR #1930, opened 2026-08-06
feat(config): make the default project id configurable via LAKEKEEPER__DEFAULT_PROJECT_ID

bolma-lila:
## Problem

Requests without an `x-project-id` header fall back to `DEFAULT_PROJECT_ID` (see `request_metadata.rs`: `self.project_id.clone().or(DEFAULT_PROJECT_ID.clone())`). But that static is **hardcoded to the NIL uuid**, gated only by `enable_default_project`:

```rust
pub static DEFAULT_PROJECT_ID: LazyLock<Option<ArcProjectId>> = LazyLock::new(|| {
    CONFIG.enable_default_project.then_some(Arc::new(uuid::Uuid::nil().into()))
});
```

So there is no way to make a **non-NIL** project the header-less default. This matters for single-tenant deployments that keep everything in one non-NIL project: every client (Trino, Spark, PyIceberg) is forced to send `x-project-id`, and query engines addressing a warehouse by bare name (`warehouse=<name>`) can't resolve it.

Notably, the docs **already reference `LAKEKEEPER__DEFAULT_PROJECT_ID`** (in the `LAKEKEEPER__OPENID_ROLES_CLAIM` entry: *"Requires a project ID to be set via the `x-project-id` header or `LAKEKEEPER__DEFAULT_PROJECT_ID`"*), but no such config field exists — the env var is silently ignored today.

## Change

Add a `default_project_id: Option<ProjectId>` config field. When `enable_default_project` is true, the effective default is `default_project_id` if set, otherwise the NIL uuid — **unchanged behaviour when the new field is unset**. When `enable_default_project` is false, the default is `None` regardless (a configured id is ignored).

The resolution is factored into `resolve_default_project_id(&DynAppConfig)` so it's unit-testable.

## Tests

- `resolve_default_project_id`: unset → NIL uuid; configured → that id; disabled → `None`.
- figment env-parse test: `LAKEKEEPER__DEFAULT_PROJECT_ID` populates `config.default_project_id`.
- `default_project_id` defaults to `None`.
- Added a `configuration.md` table entry and clarified the `ENABLE_DEFAULT_PROJECT` row.

## Notes

- Backward compatible: unset field ⇒ existing NIL-default behaviour.
- `ProjectId` already validates its input (`try_new`) and implements `Deserialize`, so an invalid value fails config parse loudly.
- I wasn't able to run `cargo test` in my environment; the change is small and type-checked by hand against `ArcProjectId = Arc<ProjectId>` and the `From<uuid::Uuid> for ProjectId` impl. Happy to adjust to match maintainer preferences (e.g. accepting a bare uuid vs. the `ProjectId` string form).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added support for configuring a custom default project.
  * Default project selection uses the configured project when available.
  * Falls back to the NIL Project ID when no custom project is configured.
  * Default project resolution can be disabled.

* **Documentation**
  * Documented the `LAKEKEEPER__DEFAULT_PROJECT_ID` configuration option and clarified default-project behavior.


bolma-lila:
Thanks @coderabbitai — addressed in 3fa865d:

**Docs (`configuration.md`)** — good catch on the ambiguity. The `None` in the default column read as a settable literal. Changed it to `_unset_ (uses NIL Project ID)`, added a real UUID example, and stated the variable must be omitted to keep the NIL default.

  (Minor: the flagged failure mechanism isn't quite right — `ProjectId::try_new` *allows* underscores (`c == '_'`), so `=None` wouldn't fail parsing on that; it would parse as a valid-but-nonexistent project id. Either way, documenting it as a settable value was misleading, so the clarification stands.)

**`typed-builder` for `DynAppConfig`** — leaving this out. It's self-labeled Trivial/Low-value, and the test fixtures use direct `DynAppConfig { ..Default::default() }` literals to match the existing style in this file. Deriving `TypedBuilder` on the core config struct is a broader convention change that seems better decided by maintainers than bundled into this small feature PR — happy to follow up separately if you'd like it.

c-thiel:
Small correction on the motivation: engines can already address a non-NIL project without the x-project-id header by using the `warehouse=<project-id>/<warehouse-name>` form — `parse_warehouse_arg` resolves the leading id segment as the project id. So `DEFAULT_PROJECT_ID` isn't closing a capability gap here; its value is ergonomics — letting operators serve a single non-NIL project so clients can pass the bare warehouse name (and skip repeating the project id on every connection). 

I think its still OK to add this.

c-thiel:
Out of curiosity: What's you use-case of using a different default?

Also @bolma-lila please note we need the CLA to be signed.

c-thiel:
Also a second hint: Lakekeeper requires a reverse proxy for TLS Termination. That can also easily fill-in a non-default project id.

c-thiel (approved):

bolma-lila:
Thanks for both hints, and for the correction on the motivation. Agreed, this is ergonomics rather than a capability gap.

**About the use case**: We bootstrapped our catalog by hand: a single warehouse in the NIL project, created through the API. That was fine for a proof of concept and stopped being fine once several teams started writing to it. We are now moving the whole topology into IaC, and the Terraform provider ([`baptistegh/lakekeeper`](https://github.com/baptistegh/terraform-provider-lakekeeper)). With that in place, the project became our unit of deployment: one project per environment, with its warehouses and namespaces created and versioned alongside it. It is also where we want to end up for multi-tenancy, since the project is the natural isolation boundary for a tenant, and we would rather not build on the assumption that everything lives in NIL.

So we moved off the default project, and the cost landed on internal users. Where they used to pass a warehouse name, they now have to know a project UUID and pass `<project-id>/<warehouse>` in every engine config, notebook, and connection string. `DEFAULT_PROJECT_ID` lets an operator serve one project per deployment (which is what our IaC declares anyway) and give people the bare name back, with the UUID in one Helm value instead of N client configs.

 **On the reverse proxy:** good hint. It does not fit our deployment though. The TLS-terminating proxy is a shared ingress controller serving the whole org. Its options are either controller-global, or per-route but gated behind a cluster-wide flag that allows arbitrary proxy config from any namespace
 
 CLA signed

c-thiel (dismissed):

bolma-lila:
I had to push commit to fix the failing CI

PR #1931, opened 2026-08-06
fix(kv2): treat Vault standby nodes as healthy

### LINKED ISSUE #1929: KV2 secret backend reports Vault unhealthy when reached through load balancer
**Describe the bug**
When Lakekeeper is configured to use an HA HashiCorp Vault cluster behind a load balancer, /health intermittently returns 503 even though Vault is healthy. This causes Kubernetes liveness probes to fail and can put the catalog into a restart loop If you are unlucky enough.
The root cause is that the KV2 secret backend polls Vault's /sys/health endpoint and treats any non-2xx response as unhealthy. Vault returns 429 from standby nodes and 473 from performance standbys [by design](https://developer.hashicorp.com/vault/api-docs/system/health). So we are effectively flagging a healthy node as unhealthy due to the response.

**To Reproduce**
1. Run Vault in HA mode with multiple nodes behind a load balancer that distributes requests across all of them.
2. Configure Lakekeeper with:
    * LAKEKEEPER__SECRET_BACKEND=kv2
    * LAKEKEEPER__KV2__URL=<vault address>
3. Verify the Vault node is healthy but in standby:
```
$ curl -i https://<vault>/v1/sys/health
HTTP/1.1 429 Too Many Requests

{"initialized":true,"sealed":false,"standby":true,"performance_standby":false,
 "ha_connection_healthy":true, ...}
```
4. Verify the same node returns 200 when standbys are explicitly accepted:
```
$ curl -o /dev/null -w '%{http_code}\n' \
'https://<vault>/v1/sys/health?standbyok=true'
200
```
5. Query Lakekeeper's /health endpoint. It returns 503, and the logs report Vault as unhealthy due to the 429 response.
6. With the default Kubernetes liveness probe (periodSeconds: 5, failureThreshold: 5), the pod is restarted after approximately 25 seconds if probes continue to hit standby nodes.

**Expected behavior**
Standby and performance standby Vault nodes should be considered healthy as they successfully serve every request Lakekeeper makes by forwarding requests to the active leader.

**System Information**
* Deployment Type: Helm chart 0.11.0 on GKE
* Lakekeeper Version: 0.13.1
* Query Engine: Not relevant
* Authentication Provider: Not relevant
* Authorization Provider: Not relevant
* OS: Linux (official container image)
* Browser: Not relevant

**Additional context**
The secret store is HashiCorp Vault 2.0.3 running in HA mode behind a GCP load balancer.
There is currently no configuration workaround:
* There is no configuration option to add query parameters to the health request.
* LAKEKEEPER__KV2__URL is used as the base for path concatenation, so including a query string in the URL produces an invalid request.

Will open a PR with a potential fix

samyrhn:
Fixes #1929 

Fix the KV2 Vault health check to treat HA standby nodes as healthy. Currently we make a request to vaults /health endpoint with no parameters. 
In this case, Vault returns HTTP `429` for standby nodes and `473` for performance standbys, and since they are not 2xx responses, we wrongly interpret it as unhealthy.

This PR proposes to pass relevant parameters ([standbyok and perfstandbyok](https://developer.hashicorp.com/vault/api-docs/system/health#parameters) to properly flag these responses as healthy instead.

This way, Lakekeeper pods doesn't fail their liveness probe and doesn't restart unnecessarily.

## Summary by CodeRabbit

* **Bug Fixes**
  * Vault health checks now recognize standby and performance-standby nodes as healthy.
  * Improved health-check compatibility for Vault deployments using standby forwarding.

### DIFF
--- a/crates/lakekeeper-secrets-kv2/src/lib.rs
+++ b/crates/lakekeeper-secrets-kv2/src/lib.rs
@@ -7,6 +7,7 @@ use serde::{Serialize, de::DeserializeOwned};
 use tokio::{sync::RwLock, time::Sleep};
 use uuid::Uuid;
 use vaultrs::{
+    api::{exec_with_no_result, sys::requests::ReadHealthRequest},
     client::{Client, VaultClient},
     error::ClientError,
 };
@@ -230,7 +231,16 @@ impl HealthExt for SecretsState {
 
     async fn update_health(&self) {
         let handle = self.vault_client.read().await;
-        let t = vaultrs::sys::health(&*handle).await;
+        // Vault returns 429 for standby nodes and 473 for performance standbys.
+        // These nodes transparently forward requests to the active leader, so
+        // treat them as healthy by requesting standbyok/perfstandbyok.
+        // https://developer.hashicorp.com/vault/api-docs/system/health
+        let endpoint = ReadHealthRequest::builder()
+            .standbyok(true)
+            .perfstandbyok(true)
+            .build()
+            .expect("ReadHealthRequest has no required fields");
+        let t = exec_with_no_result(&*handle, endpoint).await;
         match t {
             Ok(_) => {
                 tracing::debug!("Vault is healthy");


brgr-s:
@samyrhn Thank you for your PR and taking the time trying to resolve this.

The fix would indeed solve the problem that Lakekeeper restarts when K2V is in (perf)standy, but it does introduce a "leaderless-standby false-positive". This is also a problem a restart of Lakekeeper can't fix, and it probably is a much narrower gap (Vault outage).

However, I would suggest adding a config option, where the new behaviour is the default, and we have a "`LEGACY`" behaviour for the K2V Healh Probe if this is something someone struggles with, for some reason.

Apart from that, we'd need you to sign the [CLA](https://cla-assistant.io/lakekeeper/lakekeeper?pullRequest=1931) and sign your commits, please.

crates/lakekeeper-secrets-kv2/src/lib.rs
```diff
@@ -230,7 +231,16 @@ impl HealthExt for SecretsState {
 
     async fn update_health(&self) {
         let handle = self.vault_client.read().await;
-        let t = vaultrs::sys::health(&*handle).await;
+        // Vault returns 429 for standby nodes and 473 for performance standbys.
+        // These nodes transparently forward requests to the active leader, so
+        // treat them as healthy by requesting standbyok/perfstandbyok.
+        // https://developer.hashicorp.com/vault/api-docs/system/health
```
brgr-s, line 237: The comment isn't quite right: a plain standby forwards *or* 307-redirects to the leader's `api_addr` depending on cluster state, and a performance standby does the opposite — it serves reads locally from its replicated view. Could you reword, and note that sealed (503), uninitialized (501), DR secondary (472) and removed (530) are unaffected by these flags?

crates/lakekeeper-secrets-kv2/src/lib.rs
```diff
@@ -230,7 +231,16 @@ impl HealthExt for SecretsState {
...
+        let endpoint = ReadHealthRequest::builder()
+            .standbyok(true)
+            .perfstandbyok(true)
+            .build()
+            .expect("ReadHealthRequest has no required fields");
+        let t = exec_with_no_result(&*handle, endpoint).await;
```
brgr-s, line 243: `exec_with_no_result` actually returns a parsed `ReadHealthResponse` (`exec_with_empty_result` is the discarding one). Could you use it to log `standby` / `performance_standby` / `ha_connection_healthy` in the debug line?

brgr-s (changes requested):

PR #1950, opened 2026-08-12
feat: Add move/rename namespace Management API endpoint

brgr-s:
Adds a management endpoint for moving a namespace to a new location in its warehouse's hierarchy — re-parenting it, renaming it, or both.

```
POST /management/v1/warehouse/{warehouse_id}/namespace/{namespace_id}/move
{ "destination": ["analytics", "reporting"], "force": false }
→ 200 { "namespace": [...], "namespace-id": "...", "parent-namespace-id": "..." }
```

`destination` is the full new path, mirroring the IRC rename-table request. An empty path moves the namespace to the warehouse root. Keeping the parent and changing the last element renames in place. A destination equal to the current path is a no-op returning 200.

Moving requires grant authority at both ends: `manage_grants` on the namespace being moved, and on the destination parent/warehouse.

Reason: 

- Every actor holding `select`/`modify`/`describe` at the `destination` immediately gains it on the moved namespace and all its tables and views, through inheritance and with no assignment recorded. `manage_grants` makes the move grant nothing the actor could not already have granted directly.
- The `destination` needs the same treatment for a similar reson: an inbound move arrives carrying contents and existing direct grants, so gating on `create` would let a namespace be populated and granted under a permissive parent and then moved into a managed_access subtree.

Consequences:

- An owener cannot move a namespace out of a `managed_access` subtree
- `manage_grants` is independant of `modify`, so `can_move` neither implies nor is implied by `can_delete` => A `security_admin` can restructure the hierarchy with `move` without write access and an actor granted only `modify` cannot move a namespace

Other details:

- A namespace with child namespaces cannot be moved (`409 NamespaceHasChildren`), but `destination` being a full path means recursive moves can be added later.
- No cross-warehouse move
- Protected namespaces need force
- Storage layouts that derive locations from the hierarchy or from names reject the move (`400 StorageLayoutForbidsNamespaceMove`). This also applies to full-hierarchy with a {uuid}-only template on a re-parent.
- Case-only renames work.
- Input validation reuses zvalidate_namespace_ident_creationz and the reserved-namespace check, so a move cannot produce a namespace that `createNamespace` would have rejected.

## Summary by CodeRabbit

* **New Features**
  * Added an API for moving or renaming namespaces within a warehouse, including protected-namespace overrides and idempotent retries.
  * Added validation for hierarchy, naming conflicts, child namespaces, permissions, and storage-layout restrictions.
  * Added authorization controls for moving and accepting moved namespaces.
  * Expanded authorization coverage for projects, warehouses, namespaces, tables, views, roles, tags, and users.
  * Updated caches, events, and authorization relationships after successful moves.
* **Documentation**
  * Documented API behavior, authorization requirements, and storage-layout limitations.
* **Tests**
  * Added coverage for moves, retries, permissions, hierarchy updates, and cache invalidation.


authz/openfga/v4.9/components/server.fga
```diff
@@ -0,0 +1,34 @@
...
+    define can_list_all_projects: admin or operator
+    define can_list_users: admin or operator
+    define can_provision_users: admin or operator
+    define can_update_users: admin or operator
+    define can_delete_users: admin or operator
+    define can_read_assignments: admin or operator
+
+    # GRANT Permissions
+    define can_grant_admin: admin or operator
+    define can_grant_operator: admin or operator
```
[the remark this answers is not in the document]
brgr-s: 1) Out of scope for this PR. This is pre-existing, it only appears as new because the openfga model version changed and everything is re-added
2) This is intentional, see `store.fga.yaml`

crates/authz-openfga/src/authorizer.rs
```diff
@@ -927,6 +927,65 @@ impl Authorizer for OpenFGAAuthorizer {
...
+                OpenFGAError::CannotDeleteTupleNotFound(_) => {
+                    tracing::debug!(
+                        "Hierarchy tuples for namespace {namespace_id} under its old parent are already gone; treating as applied"
+                    );
+                    Ok(())
+                }
+                e => Err(e),
+            })
+            .map_err(authz_to_error_no_audit)
+            .map_err(Into::into)
```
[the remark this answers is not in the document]
brgr-s: The finding is correct, but this is also pre-existing/similar, see f. ex `delete_namespace`, `delete_table`.
A repair-mechanism should cover all sites and would be out of scope for this PR. Covering only this finding does not make much sense.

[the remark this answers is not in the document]
brgr-s: Error is fixed locally by changing and splitting commits. Re-check

docs/docs/authorization-openfga.md
```diff
@@ -115,6 +115,15 @@ Managed access combines elements of Role-Based Access Control (RBAC) and Discret
...
+
+Two consequences are worth knowing:
+
+* Managed access removes `manage_grants` from *owners*, so ownership alone does not let someone move a namespace out of a managed subtree — which is the point, since moving it out would restore their ability to grant on it. A principal **directly assigned** `manage_grants` on that namespace still holds it, and can move it: the direct assignment is unaffected by managed access, exactly as it is for every other operation `manage_grants` gates.
```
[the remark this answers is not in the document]
brgr-s: This is directly stated in the paragraph above and should be clear enough from context

authz/openfga/v4.9/components/namespace.fga
```diff
@@ -27,7 +27,9 @@ type namespace
     define can_create_view: create
     define can_create_generic_table: create
     define can_create_namespace: create
+    define can_accept_moved_namespace: manage_grants
```
c-thiel: `and create`?

authz/openfga/v4.9/components/namespace.fga
```diff
@@ -27,7 +27,9 @@ type namespace
...
     define can_delete: modify
+    define can_move: manage_grants
```
c-thiel: `and modify`?

authz/openfga/v4.9/components/warehouse.fga
```diff
@@ -24,6 +24,7 @@ type warehouse
 
     # ------------------ Actions ------------------
     define can_create_namespace: create
+    define can_accept_moved_namespace: manage_grants
```
c-thiel: `and create`?

authz/openfga/README.md
```diff
@@ -4,14 +4,27 @@
...
-## `v4.8`
+## `v4.9`
 
 ```text
 MODIFIES_TUPLES: FALSE
 ADDS_TUPLES:     FALSE
 ```
 
-Adds governance tags. Backwards-compatible: existing tuples authorize the same actions, no tuple rewrites.
+Adds namespace moves and governance tags. Backwards-compatible: existing tuples authorize the same actions, no tuple rewrites. Supersedes `v4.8`, which was never released — the governance-tag changes below were folded into this version.
```
c-thiel:
```suggestion
Adds namespace moves and governance tags. Backwards-compatible: existing tuples authorize the same actions, no tuple rewrites.
```

crates/lakekeeper-storage-postgres/src/namespace.rs
```diff
@@ -647,6 +680,224 @@ pub(crate) async fn create_namespace(
...
+        if NamespaceId::from(parent_id) == namespace_id {
+            return Err(NamespaceCannotMoveIntoSelf::new(warehouse_id, namespace_id).into());
+        }
+        Some(parent_id)
+    } else {
+        None
+    };
+    let has_destination_parent = destination_parent_id.is_some();
+
+    // Collisions are caught by the `unique_namespace_per_warehouse` constraint rather
```
c-thiel, line 912: A case-only rename never bumps version/updated_at. namespace_name is text[] COLLATE "case_insensitive"

crates/lakekeeper-storage-postgres/src/namespace.rs
```diff
@@ -647,6 +680,224 @@ pub(crate) async fn create_namespace(
...
+
+    // If it doesn't fit in an i32 it is way too large; validation would have rejected it
+    // in the catalog layer already. Matches `list_namespaces`.
+    let source_depth: i32 = source
+        .namespace_name
+        .len()
+        .try_into()
+        .unwrap_or(MAX_NAMESPACE_DEPTH + 1);
+
+    // Moving a namespace that has descendants would require rewriting the stored path of
```
c-thiel: Hm, this is a serious limitation. Why wouldn't a re-write be possible?

crates/lakekeeper/src/api/management/v1/namespace.rs
```diff
@@ -83,6 +331,171 @@ where
...
+            return Err(err);
+        }
+
+        let response = MoveNamespaceResponse {
+            namespace: moved.namespace.canonical_ident().clone(),
+            namespace_id: moved.namespace.namespace_id(),
+            parent_namespace_id: moved.namespace.parent_namespaces_id(),
+        };
+
+        // Post-commit: publish the new edge, now that the catalog has accepted the move.
```
c-thiel, line 520: new_parent here comes from authorize_namespace_move, where it was resolved from destination.parent() back at line 103 — before the write transaction opens, and without a lock. But the storage layer resolves the destination parent again inside the transaction under FOR SHARE and hands it back as moved.namespace.parent_namespaces_id(). You're already using that value for the response on line 477, just not here.

So if the destination path gets re-pointed in between — say P is renamed to Q and a fresh namespace is created at P — the row commits under the new P, but the authorizer gets told to attach to the old one. The namespace then inherits from a subtree the catalog says it isn't in, which is a surplus edge: exactly the state the table in the doc comment says this ordering rules out, and the one that needs --mode add-and-delete-drift rather than a plain additive reconcile.

Narrow window, and you need a concurrent re-parent to hit it, so not scary in practice. But the fix looks free — swap new_parent for moved.namespace.parent_namespaces_id() and both systems agree by construction, no matter what happened in between. (Would also mean authorize_namespace_move no longer needs to return the NamespaceParent at all, if you want to simplify its signature.)

crates/lakekeeper/src/api/management/v1/namespace.rs
```diff
@@ -83,6 +331,171 @@ where
...
+        // Its failure cannot be reported — the move happened — so it is logged and left to
+        // reconciliation, per the contract on `attach_namespace_parent`.
+        if reparented {
+            authorizer
+                .attach_namespace_parent(event_ctx.request_metadata(), namespace_id, new_parent)
```
c-thiel, line 525: Attaches new_parent, resolved pre-transaction at :103 with no lock. The committed parent is moved.namespace.parent_namespaces_id() — already computed eight lines up at :477. If the destination path is re-pointed in between, OpenFGA gets an edge to a subtree the catalog says the namespace isn't in

crates/lakekeeper-storage-postgres/src/namespace.rs
```diff
@@ -647,6 +680,224 @@ pub(crate) async fn create_namespace(
...
+    )
+    .fetch_one(&mut **transaction)
+    .await
+    .map_err(DBErrorHandler::into_catalog_backend_error)?;
+
+    if has_children {
+        return Err(NamespaceHasChildren::new(warehouse_id, namespace_id).into());
+    }
+
+    // Resolve the destination's parent and hold a share lock on it, so it cannot be
```
c-thiel: "hold a share lock on it, so it cannot be dropped between here and the commit" is false. drop_namespace evaluates its emptiness guard in an unlocked SELECT (:913) and deletes at :1005; the DELETE only waits on the share lock, then proceeds without re-checking.

crates/lakekeeper-storage-postgres/src/namespace.rs
```diff
@@ -647,6 +680,224 @@ pub(crate) async fn create_namespace(
...
+        .len()
+        .try_into()
+        .unwrap_or(MAX_NAMESPACE_DEPTH + 1);
+
+    // Moving a namespace that has descendants would require rewriting the stored path of
+    // every one of them. Deliberately unsupported for now — reject rather than corrupt.
+    // Any descendant counts, not just direct children.
+    let has_children = sqlx::query_scalar!(
+        r#"
+        SELECT EXISTS (
```
c-thiel: `namespace_name[1:$2]` isn't indexable, so this is a seq scan; the planner ignores the depth index O(warehouse size). Using the existing UNIQUE (warehouse_id, namespace_name) btree — descendants of P sort immediately after P — one row settles it:

```sql
SELECT coalesce(namespace_name[1:$3] = $2, false)
FROM namespace WHERE warehouse_id = $1 AND namespace_name > $2
ORDER BY namespace_name LIMIT 1
```

crates/lakekeeper-storage-postgres/src/namespace.rs
```diff
@@ -552,6 +553,38 @@ pub(crate) async fn create_namespace(
...
+    // otherwise leave the inserted child stranded under a path that no longer resolves.
+    // Taken in a separate statement because Postgres does not allow a locking clause in
+    // a CTE that is combined with a data-modifying CTE.
+    let locked_parent_id = if let Some(ref parent) = parent {
+        sqlx::query_scalar!(
+            r#"
+            SELECT namespace_id
+            FROM namespace
+            WHERE warehouse_id = $1 AND namespace_name = $2
+            FOR SHARE
```
c-thiel: This must be `FOR KEY SHARE` right?

crates/lakekeeper/src/api/management/v1/namespace.rs
```diff
@@ -17,6 +24,247 @@ impl<C: CatalogStore, A: Authorizer + Clone, S: SecretStore> NamespaceManagement
...
+}
+
+/// Request to move a namespace to a new location in the hierarchy.
+#[derive(Debug, Clone, Deserialize)]
+#[serde(rename_all = "kebab-case", deny_unknown_fields)]
+#[cfg_attr(feature = "open-api", derive(utoipa::ToSchema))]
+pub struct MoveNamespaceRequest {
+    /// Full new path of the namespace, including its new name as the last element.
+    ///
+    /// The preceding elements identify the new parent; an empty list moves the namespace to
```
c-thiel: "an empty list/path moves the namespace to the warehouse root" — validate_move_destination:174 returns 400 NamespaceEmpty for []. Root is a single-element destination. This ships into the OAPI docs

crates/lakekeeper/src/api/management/mod.rs
```diff
@@ -3756,6 +3756,61 @@ pub mod v1 {
         .await
     }
 
+    /// Move a Namespace
+    ///
+    /// Moves a namespace to a new location in the warehouse's namespace hierarchy,
+    /// re-parenting it and/or renaming it. The request body carries the full destination
+    /// path; an empty path moves the namespace to the warehouse root.
```
c-thiel: Same as https://github.com/lakekeeper/lakekeeper/pull/1950/changes#r3781378514

crates/lakekeeper/src/api/management/v1/namespace.rs
c-thiel, line 1: ## Idea: index-backed `has_children` + a possible path to nested moves

The description mentions recursive moves as a later step — I spent some time sketching how that could work and stumbled over something that I think is also useful for this PR as-is. Sharing it here for discussion; please double-check my reasoning, I may well be missing something. Numbers below are from a synthetic replica (PG 18, real ICU `case_insensitive` collation, 400k namespaces in one warehouse), so treat them as indicative, not gospel.

### Observation

`namespace_name[1:$d] = $prefix` can't use an index, so the `has_children` guard plans as a seq scan — ~146 ms at 400k namespaces on my box, paid on every *successful* move and growing with warehouse size.

The predicate itself seems index-computable though, because descendants appear to be **contiguous** in the `UNIQUE (warehouse_id, namespace_name)` btree: for any `q > p` that is not a descendant, the first element where `q` differs from `p` has `q[i] > p[i]`, which also puts `q` after every descendant `p || rest`. So all descendants of `p` sit in one contiguous range right after `p`. Worth verifying independently — this is the load-bearing assumption for everything below.

### Possible quick win for this PR

If the contiguity holds, "does `p` have descendants" is answered by the *first* row `> p`:

```sql
SELECT coalesce((
    SELECT namespace_name[1:$2] = $3
    FROM namespace
    WHERE warehouse_id = $1 AND namespace_name > $3
    ORDER BY namespace_name
    LIMIT 1
), false) AS "has_children!"
```

Index Only Scan, ~0.2 ms in my test. It also behaved correctly under the case-insensitive collation in my checks (case-variant descendants caught, `child_7x`-style prefix-similar siblings excluded) — but I'd feel better if someone re-verified that with adversarial names.

One trap I hit while experimenting, leaving it here so nobody "simplifies" this later: the tempting range form `namespace_name > $p AND namespace_name < $p || ARRAY[E'\U0010FFFF']` is also fast but **unsound** — elements starting with `U+10FFFF` or even `U+FFFF` collate above the sentinel under our ICU collation and silently escape the range (I could reproduce a missed descendant). The LIMIT-1 probe doesn't need a sentinel and doesn't have this hole.

### Sketch for nested moves (~3 statements, all index-backed)

Allowing descendants actually *deletes* the guard — the subtree becomes the work set. Rough shape:

**1. Discover + lock everything in one statement** — recursive LATERAL walk over the unique index (each step is one bounded probe, O(k·log N)), outer query takes `FOR UPDATE` on source + descendants, destination parent gets `FOR KEY SHARE` via a CTE in the same statement (locking CTEs aren't inlined, so they execute — I verified with two sessions that both locks genuinely bind):

```sql
WITH RECURSIVE walk AS (
    (SELECT namespace_id, namespace_name FROM namespace
     WHERE warehouse_id = $1 AND namespace_name > $src_path
     ORDER BY namespace_name LIMIT 1)
    UNION ALL
    SELECT nxt.namespace_id, nxt.namespace_name
    FROM walk w
    CROSS JOIN LATERAL (
        SELECT n.namespace_id, n.namespace_name FROM namespace n
        WHERE n.warehouse_id = $1 AND n.namespace_name > w.namespace_name
        ORDER BY n.namespace_name LIMIT 1
    ) nxt
    WHERE w.namespace_name[1:$d] = $src_path
),
dest_parent AS (
    SELECT namespace_id FROM namespace
    WHERE warehouse_id = $1 AND namespace_name = $dest_parent_path
    FOR KEY SHARE
)
SELECT n.*, (SELECT namespace_id FROM dest_parent) AS dest_parent_id
FROM namespace n
WHERE n.namespace_id = $2
   OR n.namespace_id IN (SELECT namespace_id FROM walk
                         WHERE namespace_name[1:$d] = $src_path)
FOR UPDATE OF n
```

(The LATERAL form matters — the plain-join spelling is correct but the planner can't push the worktable bound into the index condition and it regressed to a warehouse scan per step for me.)

Rust then checks: not-found, no-op (early return — 1 statement total), protected, and the cycle check, which becomes trivially exact: reject iff `dest_parent_id ∈ subtree_ids` — pure id membership against rows we hold locked, no path/collation comparison.

**2. Verify with a fresh snapshot** — re-run the walk read-only and compare id sets. This is *not* an optimization, it's the correctness mechanism: if a concurrent `create_namespace` had its parent `KEY SHARE` granted before statement 1 and was still in flight, statement 1 blocks and then resumes with its *original* snapshot — the new child is invisible to the walk. A fresh statement sees it. If the sets are equal we're safe (everything visible is now locked, so no further create can commit under any of it); if a new id appeared, lock it and verify again (bounded by depth ≤ 5, in practice never loops). Notably the *current* PR is immune to this exact race only because `has_children` runs as a separate later statement — that separation is quietly load-bearing, worth a comment either way.

**3. One prefix-swap UPDATE over the id set:**

```sql
UPDATE namespace
SET namespace_name = $dest_path || namespace_name[$d + 1:]
WHERE warehouse_id = $1 AND namespace_id = ANY($subtree_ids)
RETURNING namespace_id, namespace_name, version
```

PK-indexed, collisions stay constraint-backed, and the tabular `ON UPDATE CASCADE` repairs every contained table/view automatically — that machinery already scales to nested. OpenFGA also needs nothing new if I read it right: descendant parent edges are id-based and unchanged, only the root's edge re-points, so the existing detach/attach hooks should already be correct for nested.

So versus today: 4 statements + an O(warehouse) scan → 3 statements, all index-backed, and nested moves come along for the ride. (I convinced myself 1 statement is impossible in READ COMMITTED — a statement's snapshot is fixed before its locks are acquired, so discovery can never be ordered after locking within one statement — and a 2-statement variant leaves a residual double-race window that doesn't seem worth the saved index probe.)

### What it would cost

- `MoveNamespaceEvent` would need to carry every descendant's old ident (computable via prefix swap) so the cache listener can evict each old path.
- The storage-layout guard on the root should already cover descendants (re-parenting the root changes every descendant's chain), so `move_desyncs_location` stays as-is — please double-check that reading.
- Open design question: should `protected` on a *descendant* block without `force` (parity with drop's `ChildNamespaceProtected`)? Statement 1 returns the subtree, so the check would be free either way.

Happy to split this into (a) the `has_children` probe for this PR and (b) nested moves as the follow-up — and equally happy to be told the contiguity assumption doesn't hold somewhere I didn't test. 🙂

c-thiel: Cache invalidation for descendants may need **no event changes at all**: `build_hierarchy_from_cache` validates the parent-ident link on every read, and a subtree move leaves descendants' parent *ids* untouched while the root's cache entry gets refreshed by the existing event. So every stale descendant's chain breaks at the root link on first access → invalidate → refetch → healed, and the by-ident miss path already retires the old-path key. If that reading holds (please double-check — the version check at `namespace_cache.rs:405` guards the other direction, so the `is_parent_ident` check is what's load-bearing; worth pinning with a test: move a subtree, then resolve a depth-2 descendant by old path and by id on the same replica), the event stays root-only as in this PR. If eager eviction is preferred, carrying the descendant **ids** (already in hand from the discovery walk) is enough — `namespace_cache_invalidate(id)` cascades to the ident key, and recursive drop's `NamespaceDropInfo.child_namespaces` is precedent for an id-list payload. Old *idents* never need to be computed or carried.

authz/openfga/README.md
brgr-s: @c-thiel Removing that sentence leaves a gap from 4.7 to 4.9. 4.7 mentions it contains all changes since 4.0.

How about something like this

"Adds namesapces moves and includes governance tags, which where carried forward from 4.8 (unreleased)"

crates/lakekeeper-storage-postgres/src/namespace.rs
brgr-s: Not impossible, I just thought out of scope. Fixes for other finding will come first, then think about recursive move in this PR :)

crates/lakekeeper/src/api/management/v1/namespace.rs
brgr-s: Agree, that is misleading. Maybe "an empty prefix" or something better

crates/lakekeeper-storage-postgres/src/namespace.rs
brgr-s: See line 870 and 871:

```sql
let row = sqlx::query_as!(
        NamespaceWithParentVersionRow,
        r#"
        WITH updated_ns AS (
            UPDATE namespace
            SET namespace_name = $3,
                version = version + 1,
                updated_at = now()
```

crates/lakekeeper/src/api/management/v1/namespace.rs
brgr-s: (a) is done, see `lakekeeper-storage-postgress/src/namespace.rs`:

```sql
SELECT coalesce((
            SELECT namespace_name[1:$2] = $3
            FROM namespace
            WHERE warehouse_id = $1
                AND namespace_name > $3
            ORDER BY namespace_name
            LIMIT 1
        ), false) AS "has_children!"
```

(b) suggest follow-up PR with extensive subtree move testing

crates/lakekeeper-storage-postgres/src/namespace.rs
```diff
@@ -647,6 +683,258 @@ pub(crate) async fn create_namespace(
...
+    // compares equal and the trigger never fires. The trigger *assigns*
+    // `NEW.version = OLD.version + 1` rather than incrementing `NEW`, so setting both here
+    // cannot double-bump when it does fire. The byte-exact no-op returns above, so a true
+    // no-op still bumps nothing.
+    let row = sqlx::query_as!(
+        NamespaceWithParentVersionRow,
+        r#"
+        WITH updated_ns AS (
+            UPDATE namespace
+            SET namespace_name = $3,
```
c-thiel, line 929: Think this might be a free fix while you're in here: the UPDATE writes `namespace_name = $3` exactly as the caller spelled it, but the parent gets resolved case-insensitively a few lines up (`namespace_name = $2` at :830, ICU collation). So `move ns → ["PARENT", "child"]` happily succeeds when the parent row is stored as `["parent"]`.

The catalog stays correct — `parent_ns` resolves to the right id — but it quietly breaks an invariant the cache leans on. `is_parent_ident` compares byte-wise:

```rust
// Both idents come from NAMESPACE_CACHE which only stores canonical case
// ... byte-identical by construction for any valid hierarchy.
let expected_parent = &child[..child.len().saturating_sub(1)];
expected_parent == parent
```
`["PARENT"] != ["parent"]`, so `build_hierarchy_from_cache` invalidates and returns None on every subsequent by-id lookup of that namespace. Permanent miss plus removal churn, and "byte-identical by construction" stops being true.

`create_namespace` can already produce this on main (same verbatim insert, same case-insensitive parent match), so not something you introduced — but move turns it into a single request against an existing namespace, which makes it much easier to trip over.

Cheap fix here since you already hold the parent row under FOR KEY SHARE: select its stored namespace_name too and build the written array as parent_name || destination.last(). Then what lands in the table is canonical by construction. Side benefit: the case-sensitive reparented check in the handler (:203) stops treating a case-only difference in an ancestor segment as re-parenting.

Can you create an issue for `create_namespace` for this too and add it to the 0.14 Milestone?

brgr-s:
@coderabbitai review

brgr-s:
@coderabbitai review

PR #1989, opened 2026-08-24
fix(catalog): match a conditional load against the bare ETag

c-thiel:
`match_not_modified` compared the raw value, so a tag echoed back in process — weak marker and quotes included, having never passed through the header parsing that strips them — matched nothing. Any in-process caller's conditional load silently never succeeded;

Add `ETag::validator()` for the bare tag and `ETag::strip_wire_syntax()` as the single definition of that transform, then route every consumer through it: header parsing, the comparison, and the two test helpers that had each hand-rolled their own copy. `as_str()` still returns the wire form, which is what belongs in the response header.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved ETag handling for table requests and conditional responses.
  - ETag comparisons now consistently support weak, strong, quoted, unquoted, and whitespace-padded values.
  - Fixed table reload and replay scenarios involving service-generated ETags.
  - Preserved original ETag formatting while improving validation and comparison.
  - Only a bare `*` is treated as the wildcard; quoted or weakly quoted asterisks are treated as regular entity tags.


c-thiel:
@coderabbitai review

c-thiel:
https://github.com/coderabbitai review

c-thiel:
@coderabbitai review

PR #2013, opened 2026-09-07
fix(authz): re-point a tabular's OpenFGA parent on cross-namespace rename

c-thiel:
Renaming a table, view or generic table into another namespace updated the catalog but left its OpenFGA `parent` edge on the old namespace, so a principal granted only there kept access after the move.

Adds `detach_tabular_parent`/`attach_tabular_parent` hooks (no-op by default), driven by a shared `commit_rename_with_reparent`: detach pre-commit, commit, attach post-commit. Both namespace ids are now authoritative — `rename_tabular` matches the row on the caller's source namespace and name, and the destination is checked against what the statement wrote.

## Release notes

Fixed a permissions leak: a table, view or generic table renamed into another namespace kept inheriting grants from the namespace it left.

Deployments that performed such renames should run `lakekeeper openfga reconcile --mode add-and-delete-drift` once after upgrading. The default `add-missing` cannot repair it — the stale edge is a surplus tuple.

## Summary by CodeRabbit

* **New Features**
  * Cross-namespace renames of tables, views, and generic tables now transfer inherited permissions to the destination namespace.
  * Rename operations coordinate catalog and authorization updates, including retries and partial completion recovery.
  * Rename idempotency handling is now transactional.

* **Bug Fixes**
  * Prevented stale namespace permissions after cross-namespace renames.
  * Improved handling of concurrent renames, namespace changes, and destination mismatches.
  * Added guidance for repairing legacy rename drift through reconciliation.


crates/lakekeeper/src/server/generic_tables/rename.rs
brgr-s: Suggestion:
`C::get_namespace_cache_aware(warehouse_id, &destination.namespace, CachePolicy::Skip, state.clone())` (only for destination)

In combination with suggestion to use `destination_namespace_id` in `rename_tabular`, this avoids failing with "destination namespace could not be found" if we get a stale destination id from `C::get_namespace` (Request is to move to "Z", but "Z" was concurrently renamed to "A", and "B" was renamed to "Z")

Same suggestion for `tables/rename_tables.rs` and `views/rename.rs`.

I think the read is fine cost-wise.

crates/lakekeeper-storage-postgres/src/tabular/mod.rs
brgr-s: If we wire through `destination_namespace_id` which we already obtained in the caller we avoid the problem of stale destination id (Request is to move to "Z", but "Z" was concurrently renamed to "A", and "B" was renamed to "Z"). There is a similar pattern in `create_tabular`. Only problem is that the error message is off: "Either the destination namespace, or the source tabular under the given namespace and name, could not be found" is really "destination namespace cannot be found or the destinations namespace was concurrently changed to be a different namespace".... well, or something like that...

brgr-s: If we wire through `destination_namespace_id` (see suggestion below) which we already have in the caller, we can compare Ids and do not face the problem of moving `a.tb` to `A.tb2`.

brgr-s (commented):
Suggestions

brgr-s (approved):

PR #2018, opened 2026-09-11
fix: give AssertViewUUID an explicit type: object

raffael-dzikowski:
## Summary

Codegen tools that check discriminator mappings choke on `ViewRequirement` in `rest-catalog-open-api.yaml`. Its discriminator maps `assert-view-uuid` to `AssertViewUUID`, but that schema only lists `properties` and `required` — no `type: object` — so a tool has no way to confirm it's actually an object schema rather than, say, a bare `allOf`. `openapi-typescript` treats this as invalid and skips resolving it, so any generated client loses the enum narrowing on the `type` field for that variant.

Since this file is a vendored copy of the upstream Apache Iceberg spec rather than something generated from our own code, I patched it the way the other known quirks here already are: a new `fix-view-requirement-discriminator-rest-openapi` recipe, wired into `update-rest-openapi`, and applied once to the currently vendored `rest-catalog-open-api.yaml` so the fix doesn't wait for the next full re-sync.

## Summary by CodeRabbit

* **Documentation**
  * Updated the REST API specification to explicitly identify the `AssertViewUUID` schema as an object.
  * Clarified the schema structure alongside its required fields and properties.
  * Improved compatibility with tools that validate and generate code from the API specification.
  * Ensured generated API documentation reflects the corrected schema definition consistently.

### DIFF
--- a/docs/docs/api/rest-catalog-open-api.yaml
+++ b/docs/docs/api/rest-catalog-open-api.yaml
@@ -2933,6 +2933,7 @@ components:
           const: "assert-view-uuid"
         uuid:
           type: string
+      type: object
     StorageCredential:
       type: object
       required:
--- a/justfile
+++ b/justfile
@@ -53,6 +53,7 @@ update-rest-openapi:
     just add-return-protection-status-to-rest-openapi
     just add-namespace-delete-extension-to-rest-openapi
     just clarify-table-purge-default-rest-openapi
+    just fix-view-requirement-discriminator-rest-openapi
     # Remove multiple empty lines due to https://github.com/mikefarah/yq/issues/2074
     if [[ "$(uname)" == "Darwin" ]]; then \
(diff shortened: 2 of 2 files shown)


c-thiel:
Thanks for the PR @raffael-dzikowski!
I created a PR for the proper upstream fix here: https://github.com/apache/iceberg/pull/18099

raffael-dzikowski:
Hey @c-thiel, 

Thank you for that. I was under the impression that this part is vendored (copied over from the source) but fixing it upstream is the better approach as this way, this repo will get the change more naturally.

Feel free to close this one as it's not urgent anyway. 👍

Best wishes, 

Raffael

raffael-dzikowski:
Closing this to wait for the upstream merge. We can wait for it and we don't necessarily need to bridge the gap by fixing it early here.

PR #2025, opened 2026-09-16
fix: add Principal to S3-compatible STS policies

### LINKED ISSUE #1960: STS AssumeRole session policy lacks Principal, breaks vended-credentials on Yandex Object Storage (S3-compatible)
**Describe the bug**

With `sts-enabled` on an S3-compatible profile pointed at Yandex Object Storage, both warehouse creation (the storage-profile write-test) and any subsequent vended-credentials read/write fail with:

```
MalformedPolicy: required field missed: principal should be specified
```

I traced this to `get_sts_policy_string` (`crates/lakekeeper/src/service/storage/s3.rs`, `main`, currently around line 1230): every statement in the generated inline STS session policy omits `Principal`. That's correct/required for AWS proper, but Yandex Object Storage's STS (`sts.yandexcloud.net`, AssumeRole-compatible) rejects any request whose attached policy statements don't carry an explicit `Principal` — for *both* read and write operations on the resulting temporary credentials, not just at the `AssumeRole` call itself.

**To Reproduce**

1. S3-compatible storage profile, `sts-enabled: true`, `sts-endpoint: https://sts.yandexcloud.net`, valid static credentials for a service account, `sts-role-arn` set (Yandex requires this too, ≥20 chars — separate quirk, but at least it's a real config field).
2. Create/update the warehouse → the built-in write-test fails with the error above.
3. Isolated the mechanism outside Lakekeeper entirely: hand-signed `AssumeRole` (SigV4) against `sts.yandexcloud.net` with a policy shaped like Lakekeeper's (no `Principal`) → `AssumeRole` itself succeeds (200, real temp credentials returned), but a subsequent signed `GetObject` (and `PutObject`) using those credentials fails with the identical `MalformedPolicy: ... principal should be specified`.
4. Repeated the same `AssumeRole` call with `"Principal": "*"` added to each statement in the policy → `AssumeRole` still succeeds, and the follow-up `GetObject` now succeeds too (200, correct object body returned).

So the fix, empirically, is just: add `Principal` to the statements in the generated session policy. Whether that should be unconditional or gated behind something like `flavor: s3-compat` I'll leave to you — I don't have a way to confirm whether AWS STS tolerates a `Principal` element in this same inline session policy without side effects (docs are silent, we only tested against Yandex).

**Expected behavior**

`sts-enabled` should work against Yandex Object Storage the same way it does against AWS S3, without a code change on our side (there's no user-facing config in the storage-profile schema to influence policy contents).

**System Information**
- Deployment: docker compose, self-hosted
- Lakekeeper Version: 0.13.1
- Storage backend: Yandex Object Storage (S3-compatible), `flavor: s3-compat`, `sts-enabled: true`
- Trino also reads the same warehouse via `iceberg-rest` connector, static credentials, `sts-enabled: false` on that path — unaffected, only the STS/vended-credentials path hits this.

**Additional context**

Found while investigating why ClickHouse's `DataLakeCatalog` engine can't read our warehouse (separate, unrelated CH-side bug: `X-Iceberg-Access-Delegation: vended-credentials` is always sent regardless of `vended_credentials` setting — https://github.com/ClickHouse/ClickHouse/issues/105166). Enabling `sts-enabled` here was our attempt at a workaround for that, which is how we hit this.

mironovich:
This is supposed to fix https://github.com/lakekeeper/lakekeeper/issues/1960

upd:
yep, working for me with duckdb and trino

## Summary by CodeRabbit

- **New Features**
  - Added support for S3-compatible storage vendors that require an explicit `Principal` in session policies when using a separate STS endpoint.
  - Automatically includes `"Principal": "*"` in applicable downscoped policies.

- **Documentation**
  - Documented the behavior and compatibility requirements for S3-compatible storage profiles using separate STS endpoints.


mironovich:
@c-thiel Can you please look at this?

